PR #1, opened 2016-08-17
Reorganize Stuff

tpolecat:
This splits off db stuff into another project, renames the config types to `BlahConfig`, and re-organizes things more usefully. Work in progress,


cquiroz:
Should we review the code in this repo?

tpolecat:
Not much point yet, still getting organized. But in the future, definitely.

build.sbt
```diff
@@ -1,26 +1,34 @@
...
 
 lazy val core = project
   .in(file("modules/core"))
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "org.tpolecat" %% "doobie-core"               % "0.3.0-M1", // uh, lame
-      "org.tpolecat" %% "doobie-contrib-postgresql" % "0.3.0-M1"
+      "org.scalaz"  %% "scalaz-core" % "7.2.4",
+      "com.chuusai" %% "shapeless"   % "2.3.1"
```
cquiroz: Are you going to use shapeless or is it transitive to doobie?

modules/core/src/main/scala/gem/Step.scala
```diff
@@ -142,13 +144,13 @@ object GcalStep {
 // Science
 // ----------------------------------------------------------------------------
 
-final case class ScienceStep[I](instrument: I, telescope: Telescope) extends Step[I] {
+final case class ScienceStep[I](instrument: I, telescope: TelescopeConfig) extends Step[I] {
   def stepType = Step.Science
 }
 
 object ScienceStep {
```
swalker2m: This comes from the sequence model work I shared with everyone last year.  It seems like we'll want to be working with just one sequence model in the end, but it also feels like the program model work is still a bit preliminary to be relied upon in the seqexec.  I'll want to have the freedom to discover what works best and change things around for a while and I imagine the seqexec team will too.  So I'm not sure whether it makes sense to try to develop them together at this point.  I'm hoping we can refactor as necessary to put them together if necessary.

build.sbt
tpolecat: We're experimenting with self-describing data types so we can build a generic sequence editor, and I think shapeless will help eliminate some boilerplate.

tpolecat: It *is* transitive to doobie but I moved the db stuff out into its own project so this "core" project doesn't know about it.

modules/core/src/main/scala/gem/Step.scala
cquiroz: Is it clear that this code can/should be used on the front end with scala.js?

tpolecat: We haven't gotten that far yet but it would be worth checking.

build.sbt
cquiroz: The learning curve for shapeless seems steep, at least the introductions I've read are very confusing. If we need to use it, I think we should add training time for the rest of us.

PR #2, opened 2016-08-25
Functors

tpolecat:
This turns the flat model into a graph parameterized by child type, which allows the same representation to be used for different fetch depths. Experimental. Also adds deep fetching for programs. All experimental.


project/plugins.sbt
```diff
@@ -3,10 +3,10 @@ resolvers += Resolver.url(
...
 
-addSbtPlugin("org.tpolecat"      % "tut-plugin"   % "0.4.3")
+addSbtPlugin("org.tpolecat"      % "tut-plugin"   % "0.4.3") // triggers
 addSbtPlugin("com.eed3si9n"      % "sbt-unidoc"   % "0.3.1")
-addSbtPlugin("com.jsuereth"      % "sbt-pgp"      % "1.0.0")
-addSbtPlugin("com.github.gseitz" % "sbt-release"  % "1.0.0")
+addSbtPlugin("com.jsuereth"      % "sbt-pgp"      % "1.0.0") // triggers
+addSbtPlugin("com.github.gseitz" % "sbt-release"  % "1.0.0") // triggers
 addSbtPlugin("org.xerial.sbt"    % "sbt-sonatype" % "0.5.0")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-osgi"     % "0.8.0")
```
cquiroz: Why do we still have osgi?

modules/core/src/main/scala/gem/Program.scala
```diff
@@ -1,8 +1,18 @@
 package gem
 
-case class Program(id: Program.Id, title: String)
+import scalaz._, Scalaz._
+
+case class Program[A](id: Program.Id, title: String, observations: List[A])
 
 object Program {
+
   type Id = edu.gemini.spModel.core.ProgramId
```
cquiroz: Are these coming from the current ocs core module?

tpolecat: For now, yes. But I think for most things it makes sense to pull them over as needed (and some things will need to be rewritten in Scala so they work in scala-js).

cquiroz: Right, that was my concern, those clases are mostly Java so they'd break in scala.js. In ocs3 we have simple scala classes for those classes and we copy back and forth to the ocs core classes, but it would be good to avoid that

project/plugins.sbt
tpolecat: Copy/paste from another project. I'll remove it. Thanks!

tpolecat: We really don't need any of these right now.

modules/core/src/main/scala/gem/Observation.scala
```diff
@@ -26,4 +31,12 @@ object Observation {
       fromString(s).getOrElse(sys.error("Malformed Observation.Id: " + s))
 
   }
-}
\ No newline at end of file
+
+  implicit def ObservationTraverse[T]: Traverse[Observation[?]] =
```
cquiroz: I guess the `?` comes from the `kind-projector` plugin? Do you know if IDEA accepts that syntax for lambdas?

tpolecat: Yeah newer versions of IntelliJ understand kind-projector.

PR #3, opened 2016-09-01
Json

tpolecat:
This adds initial JSON encoding support and temporarily removes `Describe` implementations (they're in another project that isn't compiled right now).


build.sbt
```diff
@@ -43,3 +55,17 @@ lazy val importer = project
...
   .dependsOn(core, db)
   .settings(commonSettings)
+
+lazy val json = project
+  .in(file("modules/json"))
+  .dependsOn(core)
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "io.argonaut"                %% "argonaut"               % argonautVersion,
```
cquiroz: I'm trying to get a sense of picking argonaut over circe or viceversa. Is there a practical difference?

tpolecat: Argonaut uses scalaz, circe uses cats. That's the main difference. We may end up switching at some point.

cquiroz: Ok, good to know. I've been thinking to support binary and json on the web front end. http4s leans to circe but we have mostly used argonaut in the past

PR #8, opened 2016-09-26
initial design doc

tpolecat:


DESIGN.md
```diff
@@ -0,0 +1,67 @@
...
+
+This is a rough accounting of the design problems addressed so far.
+
+### Axis Mundi
+
+The Postgres database is the center of the world and the application design flows outward.
+
+- We want the database to maintain as much consistency as possible, such that any program interacting with the database (not just the new ODB) will see an accurate view of the world and will only be able to make modifications that are "legal" under some definition of well-formedness.
+- We want the database to be queryable easily, such that a read-replica can be made available to science staff for their own use. To this end some *checked* denormalization (such as program ids being broken into site, year, etc.) is ok.
+- Similarly astronomical units need to be in a science-friendly format, and we may need server-side functions for doing computations. We need to be cautious about pushing too much *business* logic into the database but support for domain-specific data types is probably reasonable.
```
cquiroz: What do you mean by 'business logic' in the database? Are we talking about stored procedures? I'd really have only data, relations and constraints but would be very worried about using procedures in the DB.

DESIGN.md
```diff
@@ -0,0 +1,67 @@
...
+- Enumerated types like filters, dispersers, etc., are important for reporting and consistency checking and need to be defined in the database. We use code-generation to read these in at compile time and make them available as proper enumerated types in our Scala code.
```
cquiroz: Can you expand on the code-generation part. Would it produce just e.g. case classes for the tables?

DESIGN.md
```diff
@@ -0,0 +1,67 @@
...
+- This keeps the model representation simple and flexible for dealing with the varying levels of detail needed for the UI, and means we don't have to load and ship any more information than necessary.
+
+### Data Import
+
+A separate project module (dependent on OCS) reads old XML files and then inserts programs into the new database. This works for the portion of the model described above, with the exception of users and roles which haven't been added yet.
+
+### CRDT-based Clients (i.e., no checkin/checkout or sync)
+
+The OCS program sync problem is hugely complicated and we are determined to avoid it. The new system will have no notion of "offline mode" and edits will be propagated immediately. We intend to use conflict-free replicated data types (CRDTs) for client applications to allow simultaneous editing without need for explicit sync (think Google Docs). Shane has implemented such a type that is in principle capable of handling sequence editing.
```
cquiroz: great!!! 👍

DESIGN.md
tpolecat: Yes, I agree.

DESIGN.md
tpolecat: No, it's just for enumerated types. There are a lot of them, but they relate 1:1 with lookup tables in the database and have a uniform treatment that's amenable to code generation. I don't think this is the case for other model types.

cquiroz: Sounds great, Is it done in sbt?

tpolecat: Yes, it's done as part of the build.

DESIGN.md
```diff
@@ -0,0 +1,67 @@
...
+
+- We want the database to maintain as much consistency as possible, such that any program interacting with the database (not just the new ODB) will see an accurate view of the world and will only be able to make modifications that are "legal" under some definition of well-formedness.
+- We want the database to be queryable easily, such that a read-replica can be made available to science staff for their own use. To this end some *checked* denormalization (such as program ids being broken into site, year, etc.) is ok.
+- Similarly astronomical units need to be in a science-friendly format, and we may need server-side functions for doing computations. We need to be cautious about pushing too much *business* logic into the database but support for domain-specific data types is probably reasonable.
+- Enumerated types like filters, dispersers, etc., are important for reporting and consistency checking and need to be defined in the database. We use code-generation to read these in at compile time and make them available as proper enumerated types in our Scala code.
+- Events should be logged to the database and echoed to traditional file logs in case database connectivity fails.
```
cquiroz: Traditionally logs don't go to the db due to the high cost of writing many entries to the db. How do you see that? I'm thinking we may want to log user interactions, like user X added node Y, which could be a fair amount of traffic

DESIGN.md
```diff
@@ -0,0 +1,67 @@
...
+  - F2 config (Disper, Filter, FPU, Lyot Wheel)
+
+The data model more or less reflects the above, and some read/write database access is provided but it's not totally built out. Basically just what's needed for import and simple queries.
+
+### Data Model and Level of Detail
+
+We need a straightforward way to deal with *portions* of the data model (for instance a list of programs without any observation information, for an "Open" dialog) without a proliferation of specialized model classes.
+
+- Model classes are parameterized on their children; i.e., any type that is represented by a foreign key in the schema. A `Program[Observation[...]]` is a program with a list of observations, a bit like the current model; a`Program[Observation.Id]` is a program with a list of observation *ids*; a `Program[Nothing]` has no observation information at all.
+- This keeps the model representation simple and flexible for dealing with the varying levels of detail needed for the UI, and means we don't have to load and ship any more information than necessary.
```
cquiroz: On the seqexec side we are discussing if we transfer the whole model every time there is a change or only events that change. It sounds that `gem` is leaning to only send the minimal amount of data and send edits compatible with CRDT, is that right?

DESIGN.md
tpolecat: My primary intent is to log each service call so we can see what the system is spending its time on, and have a way to do simple activity logging (who's working on what program and so on). My hope is that it will be a low constant factor but if it ends up being expensive we could reconsider. It does run on its own thread on a dedicated connection, and messages buffer up in memory.

DESIGN.md
tpolecat: Yes, after the initial load the only thing we send around is edits so they should be very small.

jdnavarro: Do you expect the UI clients to handle the synchronization of its own state with the server state with every edit?

In the long term, I think it's a good idea if those small edits is the only thing it gets sent over the wire, but wouldn't it be nice if the clients could use an API where they could expect that the state they have to deal would act as if it were immutable?

There are UI libraries that act as cache proxy between the server and the client so that UI code doesn't need to worry about state synchronization.

tpolecat: The expectation is that UI clients *publish* local edits as they happen, and integrate concurrent edits from other users as they arrive. There is never explicit synchronization except in error cases where the server might tell the client its world has become inconsistent (in which case click reload more or less). We're still trying to figure out how to do this.

PR #10, opened 2016-09-30
Adding WindowCover parameter

swalker2m:
This PR attempts to add the F2 `WindowCover` field to the model.  Besides adding the window cover itself, which is a secondary goal at best, the purpose is to explore what is required to add a field and learn more about how the model works.  To that end I plan to review my own PR with questions.

### DIFF
--- a/create.sql
+++ b/create.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.5.3
--- Dumped by pg_dump version 9.5.3
+-- Dumped from database version 9.5.4
+-- Dumped by pg_dump version 9.5.4
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
@@ -202,6 +202,20 @@ CREATE TABLE e_f2_lyot_wheel (
 
 
 ALTER TABLE e_f2_lyot_wheel OWNER TO postgres;
+
+--
+-- Name: e_f2_window_cover; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE e_f2_window_cover (
+    id identifier NOT NULL,
+    short_name character varying(20) NOT NULL,
+    long_name character varying(20) NOT NULL,
+    tcc_value character varying(20) NOT NULL
+);
+
+
+ALTER TABLE e_f2_window_cover OWNER TO postgres;
 
 --
 -- Name: e_gcal_filter; Type: TABLE; Schema: public; Owner: postgres
@@ -409,7 +423,7 @@ CREATE TABLE gem_user (
     first character varying(20) NOT NULL,
     last character varying(20) NOT NULL,
     md5 character(32) NOT NULL,
-    email character varying(40) NOT NULL,
+    email character varying(40),
     staff boolean DEFAULT false NOT NULL
 );
 
@@ -557,7 +571,8 @@ CREATE TABLE step_f2 (
     exposure_time integer NOT NULL,
     filter identifier NOT NULL,
     lyot_wheel identifier NOT NULL,
-    disperser identifier NOT NULL
+    disperser identifier NOT NULL,
+    window_cover identifier NOT NULL
 );
 
 
@@ -661,6 +676,16 @@ GemsUnder	GeMS Under	0.78400000000000003	0.0899999999999999967	f	f/33 (GeMS unde
 GemsOver	GeMS Over	0.78400000000000003	0.0899999999999999967	f	f/33 (GeMS over-sized)	GEMS_OVER
 HartmannA	Hartmann A (H1)	0	0	f	Hartmann A (H1)	H1
 HartmannB	Hartmann B (H2)	0	0	f	Hartmann B (H2)	H2
+\.
+
+
+--
+-- Data for Name: e_f2_window_cover; Type: TABLE DATA; Schema: public; Owner: postgres
+--
+
+COPY e_f2_window_cover (id, short_name, long_name, tcc_value) FROM stdin;
+Open	Open	Open	OPEN
+Close	Close	Close	CLOSE
 \.
 
 
@@ -772,7 +797,7 @@ COPY e_template (id, short_name, long_name, tcc_value, obsolete) FROM stdin;
 --
 
 COPY gem_user (id, first, last, md5, email, staff) FROM stdin;
-root			                                	 	t
+tpolecat	Rob	Norris	72b302bf297a228a75730123efef7c41	rnorris@gemini.edu	f
 \.
 
 
@@ -848,7 +873,7 @@ COPY step_dark (index, observation_id) FROM stdin;
 -- Data for Name: step_f2; Type: TABLE DATA; Schema: public; Owner: postgres
 --
 
-COPY step_f2 (observation_id, index, fpu, mos_preimaging, exposure_time, filter, lyot_wheel, disperser) FROM stdin;
+COPY step_f2 (observation_id, index, fpu, mos_preimaging, exposure_time, filter, lyot_wheel, disperser, window_cover) FROM stdin;
 \.
 
 
@@ -874,6 +899,14 @@ COPY step_science (index, observation_id, offset_p, offset_q) FROM stdin;
 
 ALTER TABLE ONLY e_f2_fpunit
     ADD CONSTRAINT e_f2_fpunit_pkey PRIMARY KEY (id);
+
+
+--
+-- Name: e_f2_window_cover_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY e_f2_window_cover
+    ADD CONSTRAINT e_f2_window_cover_pkey PRIMARY KEY (id);
 
 
 --
@@ -1215,6 +1248,14 @@ ALTER TABLE ONLY step_f2
 
 
 --
+-- Name: step_f2_window_cover_fkey; Type: FK CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY step_f2
+    ADD CONSTRAINT step_f2_window_cover_fkey FOREIGN KEY (window_cover) REFERENCES e_f2_window_cover(id);
+
+
+--
 -- Name: step_gcal_gcal_lamp_fkey; Type: FK CONSTRAINT; Schema: public; Owner: postgres
 --
 
@@ -1261,7 +1302,7 @@ ALTER TABLE ONLY step_science
 REVOKE ALL ON SCHEMA public FROM PUBLIC;
 REVOKE ALL ON SCHEMA public FROM postgres;
 GRANT ALL ON SCHEMA public TO postgres;
-GRANT ALL ON SCHEMA public TO rnorris;
+GRANT ALL ON SCHEMA public TO swalker;
 GRANT ALL ON SCHEMA public TO PUBLIC;
 
 
--- a/create.sql
+++ b/create.sql
@@ -423,7 +423,7 @@ CREATE TABLE gem_user (
     first character varying(20) NOT NULL,
     last character varying(20) NOT NULL,
     md5 character(32) NOT NULL,
-    email character varying(40),
+    email character varying(40) NOT NULL,
     staff boolean DEFAULT false NOT NULL
 );
 
@@ -797,7 +797,7 @@ COPY e_template (id, short_name, long_name, tcc_value, obsolete) FROM stdin;
 --
 
 COPY gem_user (id, first, last, md5, email, staff) FROM stdin;
-tpolecat	Rob	Norris	72b302bf297a228a75730123efef7c41	rnorris@gemini.edu	f
+root			                                	 	t
 \.
 
 
--- a/modules/core/src/main/scala/gem/TCC.scala
+++ b/modules/core/src/main/scala/gem/TCC.scala
@@ -61,10 +61,11 @@ object Tcc {
     // we repeat them below, namespaced per instrument for type safety.
 
     object F2 {
-      val Fpu       = Key[F2FpUnit   ]("fpu"      )(_.tccValue)
-      val Filter    = Key[F2Filter   ]("filter"   )(_.tccValue)
-      val LyotWheel = Key[F2LyotWheel]("lyotWheel")(_.tccValue)
-      val Disperser = Key[F2Disperser]("disperser")(_.tccValue)
+      val Fpu         = Key[F2FpUnit     ]("fpu"        )(_.tccValue)
+      val Filter      = Key[F2Filter     ]("filter"     )(_.tccValue)
+      val LyotWheel   = Key[F2LyotWheel  ]("lyotWheel"  )(_.tccValue)
+      val Disperser   = Key[F2Disperser  ]("disperser"  )(_.tccValue)
+      val WindowCover = Key[F2WindowCover]("windowCover")(_.tccValue)
     }
 
   }
--- a/modules/importer/src/main/scala/ConfigReader.scala
+++ b/modules/importer/src/main/scala/ConfigReader.scala
@@ -39,8 +39,7 @@ object ConfigReader3 {
 
     def seqOpt[S <: SequenceableSpType, A: Enumerated: TypeTag](f: A => String, default: S): Read[A] =
       cast[edu.gemini.shared.util.immutable.Option[S]].map { opt =>
-        val s = if (opt.isEmpty) default else opt.getValue
-        ufindp(f)(s.sequenceValue)
+        ufindp(f)(opt.getOrElse(default).sequenceValue)
       }
 
     def seq[S <: SequenceableSpType, A: Enumerated: TypeTag](f: A => String): Read[A] =
--- a/modules/importer/src/main/scala/ConfigReader.scala
+++ b/modules/importer/src/main/scala/ConfigReader.scala
@@ -40,23 +40,24 @@ object ConfigReader3 {
     def seq[S <: SequenceableSpType, A: Enumerated: TypeTag](f: A => String): Read[A] =
       cast[S].map(s => ufindp(f)(s.sequenceValue))
 
-    val unit:        Read[Unit       ] = _ => ()
-    val string:      Read[String     ] = cast[String]
-    val double:      Read[Double     ] = cast[Double]
-    val int:         Read[Int        ] = cast[Int]
-    val long:        Read[Long       ] = cast[Long]
-    val offsetAngle: Read[Angle      ] = string.map(s => Angle.fromArcsecs(s.toDouble))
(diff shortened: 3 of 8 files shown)


swalker2m:
First, something seems to be missing because it doesn't actually work.  Trying to import a single program with one F2 observation with 3 steps (`WindowCover` `Open`, `Closed`, `Open` each with a distinct `Filter`) fails:

````
Exception in thread "main" doobie.util.invariant$UnexpectedEnd$: ResultSet exhausted; more rows expected.
	at doobie.util.invariant$UnexpectedEnd$.<clinit>(invariant.scala)
	at doobie.hi.resultset$$anonfun$getUnique$1.apply(resultset.scala:190)
	at doobie.hi.resultset$$anonfun$getUnique$1.apply(resultset.scala:187)
	at scala.Function2$$anonfun$curried$1$$anonfun$apply$1.apply(Function2.scala:39)
	at scalaz.Free$$anonfun$map$1.apply(Free.scala:91)
	at scalaz.Free$$anonfun$map$1.apply(Free.scala:91)
	at scalaz.Free$$anonfun$foldMap$2.apply(Free.scala:213)
	at scalaz.Kleisli$$anonfun$flatMap$1$$anonfun$apply$1.apply(Kleisli.scala:47)
	at scalaz.effect.IO$$anonfun$flatMap$1$$anonfun$apply$9.apply(IO.scala:62)
	at scalaz.effect.IO$$anonfun$flatMap$1$$anonfun$apply$9.apply(IO.scala:61)
	at scalaz.Free$$anonfun$resume$2.apply(Free.scala:111)
	at scalaz.Free$$anonfun$resume$2.apply(Free.scala:111)
	at scalaz.std.FunctionInstances$$anon$1$$anonfun$map$1.apply(Function.scala:77)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free.go2$1(Free.scala:156)
	at scalaz.Free.go(Free.scala:159)
	at scalaz.Free.run(Free.scala:251)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.std.FunctionInstances$$anon$1$$anonfun$map$1.apply(Function.scala:77)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free.go2$1(Free.scala:156)
	at scalaz.Free.go(Free.scala:159)
	at scalaz.Free.run(Free.scala:251)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.std.FunctionInstances$$anon$1$$anonfun$map$1.apply(Function.scala:77)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free.go2$1(Free.scala:156)
	at scalaz.Free.go(Free.scala:159)
	at scalaz.Free.run(Free.scala:251)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.std.FunctionInstances$$anon$1$$anonfun$map$1.apply(Function.scala:77)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free.go2$1(Free.scala:156)
	at scalaz.Free.go(Free.scala:159)
	at scalaz.Free.run(Free.scala:251)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anonfun$except$1.apply(IO.scala:71)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.effect.IO$$anon$6$$anonfun$apply$26.apply(IO.scala:222)
	at scalaz.std.FunctionInstances$$anon$1$$anonfun$map$1.apply(Function.scala:77)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free$$anonfun$run$1.apply(Free.scala:251)
	at scalaz.Free.go2$1(Free.scala:156)
	at scalaz.Free.go(Free.scala:159)
	at scalaz.Free.run(Free.scala:251)
	at scalaz.effect.IO.unsafePerformIO(IO.scala:22)
	at scalaz.effect.SafeApp$class.main(SafeApp.scala:20)
	at gem.Importer$.main(Importer.scala:27)
	at gem.Importer.main(Importer.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
````

File: create.sql [hunk 1]
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.5.3
--- Dumped by pg_dump version 9.5.3
+-- Dumped from database version 9.5.4
+-- Dumped by pg_dump version 9.5.4
```
swalker2m: Note to non-Rob people viewing this code: this file is created via `pg_dump` and not directly edited.  I edited the database with Toad to make the changes and then exported.

create.sql
```diff
@@ -772,7 +797,7 @@ COPY e_template (id, short_name, long_name, tcc_value, obsolete) FROM stdin;
 --
 
 COPY gem_user (id, first, last, md5, email, staff) FROM stdin;
-root			                                	 	t
+tpolecat	Rob	Norris	72b302bf297a228a75730123efef7c41	rnorris@gemini.edu	f
```
swalker2m: I'm not sure how this got added, or where my own user would come in.

create.sql
```diff
@@ -557,7 +571,8 @@ CREATE TABLE step_f2 (
     exposure_time integer NOT NULL,
     filter identifier NOT NULL,
     lyot_wheel identifier NOT NULL,
-    disperser identifier NOT NULL
+    disperser identifier NOT NULL,
+    window_cover identifier NOT NULL
```
swalker2m: The name `step_f2` was a little confusing at first.  There is no F2 step but an F2 configuration is part of every kind of step: science, gcal, bias, etc.  That is what we have here, and actually I can't think of a better name.

create.sql
```diff
@@ -409,7 +423,7 @@ CREATE TABLE gem_user (
     first character varying(20) NOT NULL,
     last character varying(20) NOT NULL,
     md5 character(32) NOT NULL,
-    email character varying(40) NOT NULL,
+    email character varying(40),
```
swalker2m: I'm not sure how or why the email constraint disappeared.  I assume it must have to do with the two of us editing the database concurrently and doing `pg_dump`.  If so, it will continue to be an issue I guess.

create.sql
```diff
@@ -1261,7 +1302,7 @@ ALTER TABLE ONLY step_science
 REVOKE ALL ON SCHEMA public FROM PUBLIC;
 REVOKE ALL ON SCHEMA public FROM postgres;
 GRANT ALL ON SCHEMA public TO postgres;
-GRANT ALL ON SCHEMA public TO rnorris;
+GRANT ALL ON SCHEMA public TO swalker;
```
swalker2m: Is there a way to solve this issue so it doesn't change every time someone commits an update to the schema?

modules/core/src/main/scala/gem/TCC.scala
```diff
@@ -61,10 +61,11 @@ object Tcc {
...
     object F2 {
-      val Fpu       = Key[F2FpUnit   ]("fpu"      )(_.tccValue)
-      val Filter    = Key[F2Filter   ]("filter"   )(_.tccValue)
-      val LyotWheel = Key[F2LyotWheel]("lyotWheel")(_.tccValue)
-      val Disperser = Key[F2Disperser]("disperser")(_.tccValue)
+      val Fpu         = Key[F2FpUnit     ]("fpu"        )(_.tccValue)
+      val Filter      = Key[F2Filter     ]("filter"     )(_.tccValue)
+      val LyotWheel   = Key[F2LyotWheel  ]("lyotWheel"  )(_.tccValue)
+      val Disperser   = Key[F2Disperser  ]("disperser"  )(_.tccValue)
+      val WindowCover = Key[F2WindowCover]("windowCover")(_.tccValue)
```
swalker2m: The `tccValue` seems to be an integral part of the definition of all instruments fields.  I found that surprising.  Do we send `lyotWheel` to the TCC today for example?  I followed the pattern and added a `tccValue` for the window cover because it is a fundamental part of importing data from XML.

modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -75,10 +76,11 @@ object ConfigReader3 {
...
       object F2 {
-        val Fpu       = KeyRead(Tcc.Instrument.F2.Fpu,       Read.f2fpu)
-        val Filter    = KeyRead(Tcc.Instrument.F2.Filter,    Read.f2filter)
-        val LyotWheel = KeyRead(Tcc.Instrument.F2.LyotWheel, Read.f2lyotwheel)
-        val Disperser = KeyRead(Tcc.Instrument.F2.Disperser, Read.f2disperser)
+        val Fpu         = KeyRead(Tcc.Instrument.F2.Fpu,         Read.f2fpu)
+        val Filter      = KeyRead(Tcc.Instrument.F2.Filter,      Read.f2filter)
+        val LyotWheel   = KeyRead(Tcc.Instrument.F2.LyotWheel,   Read.f2lyotwheel)
+        val Disperser   = KeyRead(Tcc.Instrument.F2.Disperser,   Read.f2disperser)
+        val WindowCover = KeyRead(Tcc.Instrument.F2.WindowCover, Read.f2windowcover)
```
swalker2m: Here is where `Tcc.Instrument` is involved in importing.  Why is it TCC instrument?

project/gen2.scala
```diff
@@ -120,6 +120,12 @@ object gen2 {
         io.transact(xa).unsafePerformIO
       },
 
+      enum("F2WindowCover") {
+        type F2WindowCoverRec = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'tccValue -> String`.T
+        val io = sql"select id, id tag, short_name, long_name, tcc_value from e_f2_window_cover".query[(String, F2WindowCoverRec)].list
+        io.transact(xa).unsafePerformIO
+      },
```
swalker2m: Note, the `WindowCover` parameter is actually generated source, like all the other instrument fields.  This is actually part of the SBT build.  I would be interested in an explanation of what is going on here with `Record` and `.T` etc.  I just followed the pattern like a monkey and it worked.

Unfortunately whenever you touch this file you have to `reload` and recompile to create the enumeration.  That causes Idea to forget that you marked `target/src_managed` as generated source.  Maybe we can change where `src_managed` is written and make this a little easier.

modules/importer/src/main/scala/Importer.scala
```diff
@@ -179,7 +179,8 @@ object Importer extends SafeApp {
         val filter        = config.uget(Legacy.Instrument.F2.Filter)
         val lyoutWheel    = config.uget(Legacy.Instrument.F2.LyotWheel)
         val disperser     = config.uget(Legacy.Instrument.F2.Disperser)
-        F2Config(fpu, mosPreimaging, exposureTime, filter, lyoutWheel, disperser)
+        val windowCover   = config.uget(Legacy.Instrument.F2.WindowCover)
+        F2Config(fpu, mosPreimaging, exposureTime, filter, lyoutWheel, disperser, windowCover)
```
swalker2m: `F2Config` is the Scala case class that defines the F2 configuration at a single step.  Exactly how this gets tied into the various step types though isn't clear yet.

create.sql
tpolecat: Yeah the `root` user is required by the importer, which is why you got the error above. Unclear why it went away and mine showed up. I think maybe you need to rebase.

create.sql
tpolecat: It's not clear to me why this happens since the create is run as `postgres` and we never give ourselves grants.

modules/core/src/main/scala/gem/TCC.scala
tpolecat: The `tccValue` is only important if the value gets sent to the TCC; if we're reading it back out of the config from OCS when we import then apparently it does get sent.

modules/importer/src/main/scala/ConfigReader.scala
tpolecat: `TCC.Instrument.Filter` is a symbolic representation of the `instrument:filter` key in the sequence config.

project/gen2.scala
tpolecat: Shapeless uses this trick to allow you to express types that you can construct but can't write down in Scala. For instance:

```scala
scala> val a = 1 // Int
a: Int = 1

scala> final val a = 1 // singleton type 1
a: Int(1) = 1

scala> val a: 1 = 1 // but you can't do this (yet)
<console>:1: error: identifier expected but integer literal found.
val a: 1 = 1
       ^

scala> val a: Witness.`1`.T = 1 // so shapeless lets you do this
a: Int(1) = 1
```

create.sql
jdnavarro: I still trying to understand this project, but if this SQL table was automatically created with `pg_dump` where does the original `step_f2` come from?

tpolecat: The sequence steps are sliced across multiple tables and put back together via joining. The portion of a step that contains F2-specific configuration is stored in `f2_step`.

project/gen2.scala
jdnavarro: Is a `shapeless` an *official* dependency in `ocs3`? Can we assume we could use `shapeless` indiscriminately anywhere in `ocs3`?

I was reluctant to introduce it for the Seqexec because, for now, I could get away well without it, but if it's already included as dependency I can take advantage of it.

tpolecat: This project is independent of your work on OCS3 at the moment. We should discuss your need for Shapeless on HipChat I guess. It's a very useful library for certain kinds of generic programming but it is often overkill.

create.sql
swalker2m: Hmm, it seems I'm already on the tip.  At any rate, adding root did solve that issue.

modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -37,26 +37,33 @@ object ConfigReader3 {
     def enum[A: Enumerated: TypeTag](f: A => String): Read[A] =
       cast[String].map(ufindp(f))
 
+    def seqOpt[S <: SequenceableSpType, A: Enumerated: TypeTag](f: A => String, default: S): Read[A] =
+      cast[edu.gemini.shared.util.immutable.Option[S]].map { opt =>
+        val s = if (opt.isEmpty) default else opt.getValue
+        ufindp(f)(s.sequenceValue)
+      }
+
```
swalker2m: Turns out that "staff-only" fields are added as instances of Gemini `Option`.  This is some rather gross code to extract a value from the sequence when we know it is one of these types.

modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -37,26 +37,33 @@ object ConfigReader3 {
...
+    val instrument:    Read[Instrument   ] = enum(_.tccValue)
+    val yesNo:         Read[Boolean      ] = cast[YesNoType].map(_.toBoolean)
+    val durSecs:       Read[Duration     ] = double.map(_.toInt).map(Duration.ofSeconds(_))
+    val gcalLamp:      Read[GCalLamp     ] = cast[JSet[Object]].map(_.iterator.next.toString).map(ufindp[GCalLamp, String](_.tccValue))
+    val gcalShutter:   Read[GCalShutter  ] = seq[OldGCal.Shutter, GCalShutter](_.tccValue)
+    val f2fpu:         Read[F2FpUnit     ] = seq[OldF2.FPUnit,      F2FpUnit     ](_.tccValue)
+    val f2filter:      Read[F2Filter     ] = seq[OldF2.Filter,      F2Filter     ](_.tccValue)
+    val f2lyotwheel:   Read[F2LyotWheel  ] = seq[OldF2.LyotWheel,   F2LyotWheel  ](_.tccValue)
+    val f2disperser:   Read[F2Disperser  ] = seq[OldF2.Disperser,   F2Disperser  ](_.tccValue)
+    val f2windowcover: Read[F2WindowCover] = seqOpt[OldF2.WindowCover, F2WindowCover](_.tccValue, OldF2.WindowCover.OPEN)
```
swalker2m: Here I use `seqOpt` to convert the window cover.  It's vaguely coming back to me though that this may not be a good idea.  I think in the seqexec there is a difference between defining a value for an expert field like `WindowCover` and simply leaving it be ...

create.sql
cquiroz: Toad?

project/gen2.scala
cquiroz: `shapeless` is not in `ocs3`

create.sql
swalker2m: http://www.toadworld.com/m/freeware/628

swalker2m: I'm not sure it offers all that much beyond `psql` but I probably am not using it correctly.

project/gen2.scala
tpolecat: Incidentally the witness hack is no longer required with the Typelevel Scala compiler. You can say `val x: List[42] = ...` for a list that can only contain the value `42`. With some luck this will get merged into mainline Scala in the 2.12 timeframe.

swalker2m:
I've removed inadvertent changes to the `create.sql` file.  I suppose I'll go ahead and merge this?

The issue with `WindowCover` and other expert parameters may need to be addressed at some point.

tpolecat:
Yep 👍

PR #11, opened 2016-10-12
Initial Flyway Support

swalker2m:
The idea of this PR is to solicit feedback about how and whether to use [Flyway](https://flywaydb.org) for database migrations.  In this PR I've simply added the plugin and made it possible to use from SBT via a subproject called "sql":

````
> sql/flywayMigrate
[info] Database: jdbc:postgresql:gem (PostgreSQL 9.5)
[info] Successfully validated 1 migration (execution time 00:00.029s)
[info] Creating Metadata table: "public"."schema_version"
[info] Current version of schema "public": << Empty Schema >>
[info] Migrating schema "public" to version 001 - Initial Setup
[info] Successfully applied 1 migration to schema "public" (execution time 00:00.246s).
[success] Total time: 0 s, completed Oct 11, 2016 5:33:40 PM

> sql/flywayInfo
[info] Database: jdbc:postgresql:gem (PostgreSQL 9.5)
[info] +---------+---------------+---------------------+---------+
[info] | Version | Description   | Installed on        | State   |
[info] +---------+---------------+---------------------+---------+
[info] | 001     | Initial Setup | 2016-10-11 17:33:40 | Success |
[info] +---------+---------------+---------------------+---------+
[success] Total time: 0 s, completed Oct 11, 2016 5:33:45 PM

````

Some questions to consider:

* Should we start using it now?  I've already screwed up once by updating the source but failing to update the database at the same time.  On the other hand, it might be a bit more onerous to write the migrations in separate files instead of just doing a `pgdump`.  (Though personally I think I prefer just writing SQL snippets instead of using, say, Toad and doing `pgdump`.)

* I placed the content of the `create.sql` script in the initial migration file in the `sql` directory.  This seemed like as good of a place as any to avoid polluting the base directory.  But do we plan on running flyway from an application?  If so, we'd want the migrations to be included in a JAR file in the distribution of the application I assume.

* There is enough there to simply run flyway in SBT but that isn't giving us much over just using flyway from the shell.  Do we want to clean the database when we clean from SBT for example, and then run migrate before each build?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,12 @@ lazy val commonSettings = Seq(
   libraryDependencies += "org.scala-lang" %  "scala-reflect" % scalaVersion.value
 )
 
+lazy val flywaySettings = Seq(
+  flywayUrl             := "jdbc:postgresql:gem",
+  flywayUser            := "postgres",
+  flywayLocations       := Seq("filesystem:sql")
+)
+
 // N.B. `describe` is not a project yet. It doesn't quite compile. Will need some rejiggering.
 
 lazy val gem = project
@@ -81,3 +87,8 @@ lazy val service = project
   .in(file("modules/service"))
   .dependsOn(core, db)
   .settings(commonSettings)
+
+lazy val sql = project
+  .in(file("sql"))
+  .settings(flywaySettings)
+
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -3,6 +3,9 @@ resolvers += Resolver.url(
     url("http://dl.bintray.com/content/tpolecat/sbt-plugin-releases"))(
         Resolver.ivyStylePatterns)
 
+resolvers  += "Flyway" at "https://flywaydb.org/repo"
+
+addSbtPlugin("org.flywaydb"      % "flyway-sbt"   % "4.0.3")
 addSbtPlugin("org.tpolecat"      % "tut-plugin"   % "0.4.3") // triggers
 addSbtPlugin("com.eed3si9n"      % "sbt-unidoc"   % "0.3.1")
 addSbtPlugin("com.jsuereth"      % "sbt-pgp"      % "1.0.0") // triggers
--- a/sql/V001__Initial_Setup.sql
+++ b/sql/V001__Initial_Setup.sql
@@ -1,0 +1,1312 @@
+--
+-- PostgreSQL database dump
+--
+
+-- Dumped from database version 9.5.4
+-- Dumped by pg_dump version 9.5.4
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET client_encoding = 'UTF8';
+SET standard_conforming_strings = on;
+SET check_function_bodies = false;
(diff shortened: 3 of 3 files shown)


tpolecat:
This looks good. Let's go ahead and start using it and see how it goes. It probably makes sense to run the migration as part of the build at some point but doing it manually is fine for now.

I haven't thought a whole lot about deployment but it would be nice if the app were able to do migrations on startup. I guess I was hoping we could clone the VM with its existing database and then startup a new server, migrate, and if all is well change the DNS pointer and shut down the old vm. Something like that. It might be useful to use Docker or something. Dunno.

How do you develop and test migrations? Like, if you run one and it's wrong how do you clean up and reset flyway and try again?

cquiroz:
I don't know the details about flywheel but I'm all for using  a tool for migrations 👍 
I guess you can do reverse migrations too?
Can we use them in tests too? I guess some amount of integration testing with a real db would be good. It could give you an initial test db that is in a sane state for tests

When you have db changes, do you put them in a new `sql` file? Or do you update `create.sql`?

I'd like to think about the automigration on startup. The scenario that worries me is some developer accidentally starting the production app and migrating the production db before due time.

Just a small thing, I'd somehow expect the sql files to live in `<project>/src/main/sql`

swalker2m:
> How do you develop and test migrations? Like, if you run one and it's wrong how do you clean up and reset flyway and try again?

Each migration is run as a [transaction](https://flywaydb.org/documentation/faq#rollback), if I understand it correctly.  If it blows up in the middle then you can just adjust the migration script and try again.  If it works but doesn't do what you wanted I think you're left with doing a [clean](https://flywaydb.org/documentation/command/clean) (`sql/flywayClean`) to wipe out everything followed by a [migrate](https://flywaydb.org/documentation/command/migrate) (`sql/flywayMigrate`) with an updated script.  I suppose we should probably write a `reset` command that combines these two.

swalker2m:
> I guess you can do reverse migrations too?

Nope.   They [claim](https://flywaydb.org/documentation/faq#downgrade) these don't work in practice. 

> Can we use them in tests too? I guess some amount of integration testing with a real db would be good. It could give you an initial test db that is in a sane state for tests.

Good question.  There is a [Java API](https://flywaydb.org/documentation/api/) that we could use, presumably.

> When you have db changes, do you put them in a new sql file? Or do you update create.sql?

The `create.sql` file would go away in favor of the initial migration script `V001__Initial_Setup.sql`.  Each update then goes in its own versioned file.  For example, `V002__Add_FooBarTable.sql`.  One nice thing is that you can query the version table to see which migrations have been applied (or use the provided `sql/flywayInfo` command).

> Just a small thing, I'd somehow expect the sql files to live in `<project>/src/main/sql`.

That's totally configurable.  Especially if they are used from test or application startup, that might be best?  If not, it seems more convenient to just drop them in an `sql` directory off of the project root.

tpolecat:
Let's talk (next week) about deployment options and how we're going to manage dev/test/production, migrations, and migration testing.

I'm fine merging this.

PR #12, opened 2016-10-14
Event Log

swalker2m:
This PR is meant to accompany [my notes](https://github.com/swalker2m/gem/wiki/EventLog.md) on the same subject and provide a place for feedback and discussion.  I don't think this is actually ready to be merged.

(Also, `create.sql` should actually go away at some point if we're ready to embrace Flyway.  It is inadvertently and trivially modified here.)


tpolecat:
Can we put the notes themselves in a markdown doc as part of the PR instead of using the wiki?

swalker2m:
Added the event log wiki page as part of the PR, as requested.  Since this won't be merged in its current form I just dumped it into the project root.

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+and the table itself is
+
+````sql
+CREATE TABLE log_observe_ctrl
+(
+   id            SERIAL,
+   "timestamp"   timestamp (5) WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+   observe_ctrl  evt_observe_ctrl NOT NULL,
+   sequence_id   text NOT NULL,
+   why           text
```
tpolecat: Can we characterize `why` any better than this, or is it really an arbitrary string?

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+      SELECT timestamp,
+             lifecycle,
+             sequence_id,
+             step
+        FROM log_observe_int
+       WHERE timestamp BETWEEN ${ts(start)} AND ${ts(end)}
+    ORDER BY timestamp
+    """.query[(Timestamp, Lifecycle, String, Int)].map { case (t,l,i,s) =>
+      val timestamp = t.toInstant
+      val sid       = Sequence.Id.unsafeFromString(i)
```
tpolecat: You can add mappings to/from `Instant` and `Sequence.Id` to avoid the extra encoding/decoding here. I have been putting these in the `dao` package object.

```scala

implicit val SequencedMeta: Meta[Sequence.Id] =
  Meta[String].nxmap(Sequence.Id.unsafeFromString, _.toString)
```

This will allow you to use these types directly as interpolated values in sql literals and as mapped output types in your results.

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+    for {
+      c <- selectAllObserveControl(start, end)
+      i <- selectAllObserveIntegration(start, end)
+      s <- selectAllObserveStage(start, end)
+    } yield (c ++ i ++ s).sortBy(_.timestamp)
+
+}
+````
+
+I'm just combining the data collected from the three tables in Scala code.  Maybe there's a more efficient way to do this in a query?
```
tpolecat: You can use a `UNION` query which will pull in the step and why as `Option`s and then pattern match using `None` for cases where they can't apply. You'll end up with a final "this doesn't make sense" case that can just throw an exception. But if you're doing this it might be worth smashing into a single table as described above.

tpolecat:
(In general I hate wikis … I think of doc as source code and would prefer to manage it via PR like anything else.)

EventLog.md
swalker2m: I was thinking the "why" would be typed in by a human to explain why she is pausing or aborting the observation.  That's often a weather related reason like "humidity rising" or "cloud passing" but I don't know if we can enumerate all the reasons.

Maybe it doesn't belong here though.  A nightlog app would obviously want to draw upon these same events, combined with faults and other random comments from observers.  It would need its own model though and that might be a better place for talking about "why" they are doing things.

tpolecat: Ok. Yeah as soon as you let a human start typing shit in they'll start trying to encode structured data. So we might want to push on this a little more.

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+
+where the `PROGID` has its own internal fields and the `DATASET#` has a minimum of 3 digits with leading zeros as necessary.  For example, Gemini North 2017A queue program 1, observation 2, dataset/step 3 is represented with
+
+````
+GN-2017A-Q-1-2-003
+````
+
+Presumably countless python-based (i.e., maintenance-challenged) scripts have been written assuming that format.  If we introduce some kind of sequence identification between the observation number and the dataset number, they will all need to be updated.
+
+Now add to this complexity the fact that new instruments will generate multiple (maybe even an *unknown* number of) datasets per sequence step.  Either the familiar dataset label has to change or else we have to give up on the idea of mapping datasets to individual sequence steps via the dataset label itself.  I'll assume the later and so a query will have to be performed if we want to map a label to a particular step in a particular sequence of an observation.  (Or perhaps I'm wrong?  Maybe all he python scripts work with the filename format which doesn't reference the dataset label at all?)
```
cquiroz: Are you referring to GHOST in particular, producing multiple datasets?

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+    def unapply(arg: Id): Option[(Observation.Id, String)] =
+      Some((arg.oid, arg.name))
+  }
+
+}
+````
+
+## Observation Execution Events
+
+In today's OCS, the tcc and seqexec send observation execution events to an intermediary process called the "wdba" as they slew to targets and execute sequences.  Perhaps in the future these tools will write directly to the database?  At any rate, all the events share a timestamp and an observation id but some types of events need additional information.  In the multi-sequence future, the observation id should be replaced with a sequence id for all (or maybe just most) events.
```
cquiroz: If the tcc/seqexec would write updates directly to the db, how would clients be notified. E.g. you may want to display on the OT that an observation has completed. Perhaps they would instead notify the OCS and it will update the db and notify cilents

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+  case class EventStopObserve(    timestamp: Instant, sid: Sequence.Id, why: Option[String]) extends Event
+
+  case class EventStartIntegration(timestamp: Instant, sid: Sequence.Id, step: Int) extends Event
+  case class EventEndIntegration(  timestamp: Instant, sid: Sequence.Id, step: Int) extends Event
+}
+````
+
+Here I replaced "dataset" start/end events with "integration" start/end keeping in mind that multiple datasets may be produced for each step.
+
+As indicated by vertical whitespace in the code above there are three categories of events.  Correspondingly, I assume there will need to be three tables to store them (?).
```
cquiroz: 👍 for a single table

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+ALTER TABLE log_observe_stage OWNER TO postgres;
+
+ALTER TABLE ONLY log_observe_stage
+    ADD CONSTRAINT log_observe_stage_pkey PRIMARY KEY (id);
+
+CREATE INDEX ix_log_observe_stage_timestamp ON log_observe_stage USING btree ("timestamp" DESC);
+````
+
+There are a couple of things I'm not sure about here.  First, `timestamp`, even in combination with sequence id, doesn't seem like a valid primary key.  Do we need a primary key at all?  We'll never actually join on the primary key, I think. The reason I could find for it was avoiding duplicates when something goes wrong reading from a backup.
+
```
cquiroz: I'd think is always better that the primary key has no real meaning, it is just an id

EventLog.md
```diff
@@ -0,0 +1,397 @@
...
+### Determining the Current State
+
+From the information in the event logs, we should be able to determine:
+
+1. which steps are executed and in which order across the sequences
+2. today's time accounting information
+3. which step is currently executing, if any
+
+I don't think there is a need to move executed steps into a separate log of any kind? Though we're relying on code to avoid rewriting history I suppose.
```
cquiroz: Should we also include step's in pause and/or marked as skipped

EventLog.md
jdnavarro: Not that I'm against a single table for this case because I don't understand the requirements, but wouldn't this be a case of premature optimization? It all depends of how the data will end up being accessed, if every time you access it you have to do the union query it makes sense to collapse it into one table, but if we need the flexibility to retrieve or insert data in unpredictable ways, I think it would reasonable to have separate tables. If performance is an issue we can always create an index for that query.

I lean towards doing normalization by default and only denormalize when there is a clear advantage.

EventLog.md
swalker2m: Since the seqexec is managing the sequence execution and is the source of the events, it should probably be the one to update the database.  Otherwise we have another moving part that has to be working to keep the model consistent with reality. That doesn't stop the seqexec from also notifying clients that there's been an update or clients could poll to find out about updates.

swalker2m:
Updated to use a single combined table for the event log, as suggested.

modules/db/src/main/scala/gem/dao/EventLogDao.scala
```diff
@@ -0,0 +1,93 @@
...
+             step
+        FROM log_observe_event
+       WHERE timestamp BETWEEN ${ts(start)} AND ${ts(end)}
+    ORDER BY timestamp
+    """.query[(Instant, EventType, Sequence.Id, Option[Int])].map { case (t,e,s,i) =>
+
+      (e match {
+        case Abort            => EventAbortObserve(t, s)
+        case Continue         => EventContinueObserve(t, s)
+        case EndIntegration   => EventEndIntegration(t, s, i.get)
```
swalker2m: The `get` is guaranteed to work by the constraint on the `step` column, but this still feels sketchy.  Advice?

tpolecat: I would write this as 

```
query[...].map {
  case (t, Abort, s, None)             => EventAbortObserve(t, s)
  ...
  case (t, EndIntegration, s, Some(i)) => EventEndIntegration(t, s, i.get)
  ...
  case _ => sys.error("got back a bogus row, someone screwed up")
}
```

or similar, so we check the `None` cases as well; getting a value back in that column would be an error and is something we should check I think.

swalker2m:
Cleaned up a bit and updated to address feedback.

tpolecat:
Looks good. 👍

PR #13, opened 2016-10-14
Change sourceManaged directory?

swalker2m:
I'm wondering if we might consider changing `sourceManaged` so that it isn't created in the `target` directory.  Whenever you touch the build file and sync in Idea, it forgets that the generated sources directory should be treated as source files.  That happens even if you move it out of the target directory, but at least there's a bit less fiddling with it because you don't have to cancel the exclusion on the `target` directory first.

Where they are written (other than in a `target` subdirectory) doesn't matter much to me.  I chose `src_managed` because there was already an entry in `.gitignore` for it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,6 +33,7 @@ lazy val core = project
       "org.scalaz"  %% "scalaz-core" % scalazVersion,
       "com.chuusai" %% "shapeless"   % shapelessVersion
     ),
+    sourceManaged <<= baseDirectory { _ / "src_managed" },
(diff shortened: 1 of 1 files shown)


cquiroz:
Generated files works fine in ocs3. They go under `target/scala-2.11/src_managed/main` and IDEA picks them. On reload it works fine too. 

What could be different in `gem`?

swalker2m:
I can't build in Idea this project in Idea anymore and this wasn't that great of an improvement anyway.

PR #14, opened 2016-10-18
Import fix

tpolecat:
This gets the import working again, with the flyway-managed database. Kind of two concerns I guess but that's how it goes sometimes.

- The migrate script failed for me because it mentions user `swalker`. I removed that. This shouldn't be a problem again. You'll have to drop and re-create the database or flyway will complain.
- I updated the readme and got the import working with `F2WindowCover` and missing exposure times in steps (defaults to zero).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,20 @@ You need Postgres 9.5 or better. I **highly** recommend using [Postgres.app](htt
 export PATH=$PATH:/Applications/Postgres.app/Contents/Versions/latest/bin
 ```
 
-Next you can run the following to set up the database.
+Next you can run the following to create the database and user.
 
 ```
 psql -c 'create user postgres createdb'
 psql -c 'create database gem' -U postgres
-psql -c '\i create.sql' -d gem -U postgres
 ```
+
+Now initialize the database by running the migration script.
+
+```
+sbt sql/flywayMigrate
+```
+
+You can also just do `sql/flywayMigrate` from the sbt prompt if you already have sbt running. You will repeat this step each time the schema changes. It's not part of the build yet but probably will be soon.
 
 If you ever want to wipe out the database and start over, you can do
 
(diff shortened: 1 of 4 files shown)


modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -62,7 +58,17 @@ object ConfigReader3 {
     val f2filter:      Read[F2Filter     ] = seq[OldF2.Filter,      F2Filter     ](_.tccValue)
     val f2lyotwheel:   Read[F2LyotWheel  ] = seq[OldF2.LyotWheel,   F2LyotWheel  ](_.tccValue)
     val f2disperser:   Read[F2Disperser  ] = seq[OldF2.Disperser,   F2Disperser  ](_.tccValue)
-    val f2windowcover: Read[F2WindowCover] = seqOpt[OldF2.WindowCover, F2WindowCover](_.tccValue, OldF2.WindowCover.OPEN)
+
+    // It appears that the window cover is sometimes "bare" and someimes wrapped in `Some` ... it is
+    // not `None` in my test data so there's not a mapping for it.
+    val f2windowcover: Read[F2WindowCover] = { a =>
+      ufindp[F2WindowCover, String](_.tccValue) {
```
cquiroz: What's `ufindp`?

README.md
```diff
@@ -11,14 +11,21 @@ You need Postgres 9.5 or better. I **highly** recommend using [Postgres.app](htt
...
 
+Now initialize the database by running the migration script.
+
+```
+sbt sql/flywayMigrate
+```
+
+You can also just do `sql/flywayMigrate` from the sbt prompt if you already have sbt running. You will repeat this step each time the schema changes. It's not part of the build yet but probably will be soon.
+
 If you ever want to wipe out the database and start over, you can do
```
swalker2m: I suppose we could mention `sbt sql/flywayClean` for wiping out the database and starting over as well.

tpolecat: Ah cool, I didn't know that one.

PR #15, opened 2016-10-27
Initial multi-sequence ideas

swalker2m:
This PR contains a straightforward extension of the model to allow an observation to have multiple sequences.  A few notes:

1) I changed `sql/V001__Initial_Setup.sql` directly instead of creating a new migration.  I suppose that's not the proper way to do things but maybe until we have actual production data to migrate it would be less confusing to just clean/migrate each time?

2) I don't have much experience with relational databases so bear with me.  I'm not sure for example whether it makes sense to place two indices on the step table (one for `observation_id` and one for `sequence_id`).

3) It would probably be best to wait for more information from science before merging this PR regardless.  Maybe it doesn't make sense.

### DIFF
--- a/modules/core/src/main/scala/gem/Sequence.scala
+++ b/modules/core/src/main/scala/gem/Sequence.scala
@@ -3,6 +3,10 @@ package gem
 import scalaz._
 import Scalaz._
 
+case class Sequence[S](
+  id: Sequence.Id,
+  steps: List[S])
+
 object Sequence {
 
   trait Id {
@@ -10,22 +14,34 @@ object Sequence {
     def name: String
   }
 
-  object Id extends ((Observation.Id, String) => Option[Sequence.Id]) {
-    def apply(o: Observation.Id, n: String): Option[Sequence.Id] =
+  object Id {
+    def fromName(o: Observation.Id, n: String): Option[Sequence.Id] =
       !(n.contains('-') || "" === n.trim) option new Sequence.Id {
         override val oid: Observation.Id = o
         override val name: String        = n.trim
 
+        override def equals(o: Any): Boolean =
+          o match {
+            case id: Id => id.oid === oid && id.name === name
+            case _      => false
+          }
+
+        override def hashCode: Int =
+          41 * (41 + oid.hashCode) +  name.hashCode
+
         override def toString: String =
           s"${oid.toString}-$n"
       }
+
+    def unsafeFromName(o: Observation.Id, n: String): Sequence.Id =
+      fromName(o, n) | sys.error(s"Illegal sequence name '$n'")
 
     def fromString(s: String): Option[Sequence.Id] =
       s.lastIndexOf('-') match {
         case -1 => None
         case  n =>
           val (a, b) = s.splitAt(n)
-          Observation.Id.fromString(a).flatMap { Id(_, b.drop(1)) }
+          Observation.Id.fromString(a).flatMap { Id.fromName(_, b.drop(1)) }
       }
 
     def unsafeFromString(s: String): Sequence.Id =
@@ -35,4 +51,11 @@ object Sequence {
       Some((arg.oid, arg.name))
   }
 
+  implicit def SequenceTraverse[T]: Traverse[Sequence[?]] =
+    new Traverse[Sequence[?]] {
+      def traverseImpl[G[_]: Applicative, A, B](fa: Sequence[A])(f: A => G[B]): G[Sequence[B]] =
+        fa.steps.traverse(f).map(ss => fa.copy(steps = ss))
+    }
+
+  implicit val EqualSequenceId: Equal[Sequence.Id] = Equal.equalA
 }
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/gem/Sequence.scala
```diff
@@ -3,29 +3,45 @@ package gem
...
 
   trait Id {
     def oid: Observation.Id
     def name: String
   }
 
-  object Id extends ((Observation.Id, String) => Option[Sequence.Id]) {
-    def apply(o: Observation.Id, n: String): Option[Sequence.Id] =
+  object Id {
+    def fromName(o: Observation.Id, n: String): Option[Sequence.Id] =
```
swalker2m: `Id` isn't a `case class` because not just any `name` `String` is valid.  I think there's some trick with `sealed abstract case class` that lets you control access to the constructor?  So there might be a better way to do this.

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -89,21 +90,51 @@ object StepDao {
...
+       FROM step s
+              LEFT OUTER JOIN step_gcal sg
+                 ON sg.sequence_id = s.sequence_id AND sg.index = s.index
+              LEFT OUTER JOIN step_science sc
+                 ON sc.sequence_id = s.sequence_id AND sc.index = s.index
+       WHERE s.sequence_id = $sid
+    ORDER BY s.index
+    """.query[StepKernel].list.map { (lst: List[StepKernel]) =>
+      lst.headOption.map { sk => Sequence(sk.sid, lst.map(_.toStep)) }
+    }
```
swalker2m: These two queries are almost the same.  They differ in the `WHERE` and `ORDER BY` clauses.  Is there a way to share the rest?

modules/db/src/main/scala/gem/dao/ProgramDao.scala
```diff
@@ -114,20 +114,20 @@ object ProgramDao {
       case x => sys.error("Unexpected Option[Step] inputs: " + x)
     }, _ => sys.error("decode only"))
 
-  private implicit val OptionObservationOptionStep: Composite[Option[(Observation[Nothing], Option[Step[_]])]] =
-    Composite[(Option[Observation.Id], Option[String], Option[Step[_]])].xmap({
-      case (None,      None,        None) => None
-      case (Some(oid), Some(title), step) => Some((Observation(oid, title, None, Nil), step))
+  private implicit val OptionObservationOptionStep: Composite[Option[(Observation[Nothing], Option[Sequence.Id], Option[Step[_]])]] =
```
cquiroz: This looks big. What do you need the `Composite` for?

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -89,21 +90,51 @@ object StepDao {
...
+              sg.shutter,
+              sc.offset_p,
+              sc.offset_q
+       FROM step s
+              LEFT OUTER JOIN step_gcal sg
+                 ON sg.sequence_id = s.sequence_id AND sg.index = s.index
+              LEFT OUTER JOIN step_science sc
+                 ON sc.sequence_id = s.sequence_id AND sc.index = s.index
+       WHERE s.sequence_id = $sid
+    ORDER BY s.index
```
cquiroz: What's the `index` column? is it like the position of the sequence?

sql/V001__Initial_Setup.sql
```diff
@@ -527,9 +527,12 @@ COMMENT ON TABLE semester IS '// TODO: start/end dates for site';
 
 CREATE TABLE step (
     observation_id character varying(40) NOT NULL,
+    sequence_name character varying(20) NOT NULL,
     index smallint NOT NULL,
     instrument identifier NOT NULL,
-    step_type step_type NOT NULL
+    sequence_id character varying(60) NOT NULL,
+    step_type step_type NOT NULL,
+    CONSTRAINT sequence_id_check CHECK (((sequence_id)::text = (((observation_id)::text || '-'::text) || sequence_name)))
```
cquiroz: nice 👍

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -89,21 +90,51 @@ object StepDao {
       }
   }
 
-  def selectAll(oid: Observation.Id): ConnectionIO[List[Step[Instrument]]] =
+  def selectIds(oid: Observation.Id): ConnectionIO[List[Sequence.Id]] =
```
cquiroz: given that the query is `DISTINCT` i guess you could return a `Set`?

modules/db/src/main/scala/gem/dao/ProgramDao.scala
swalker2m: Rob would have a more detailed answer but this enables Doobie to figure out how to map query output where we are expecting that type: `Option[(Observation[Nothing], Option[Sequence.Id], Option[Step[_]])]`.

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: That's right.  I guess I decided to return a list since ids are ordered.

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: Right, the "step" number in today's model.

PR #16, opened 2016-10-28
Remove sequence id concept.

swalker2m:
A straightforward update to event log to remove sequence ids, since we've determined that this complication will not be necessary.

Again, rather than adding a new migration I just edited the previous one which implies a clean/migrate cycle.  This seems better than complicating the setup, at least until there is actual production data to worry about.


tpolecat:
Seems fine. 👍

PR #17, opened 2016-11-07
Groups Discussion

swalker2m:
These brief notes contain my memory of a discussion with Bryan, Andy, and Rob on groups in the new OCS along with some minor embellishments of my own.  My memory may be poor in some cases and my embellishments might be nonsense so I wanted to share this for input.  I probably forgot something important as well.

I'm looking for feedback on whether these group types make sense and what else might be missing. I'm guessing Bryan and/or Andy have documented all of this somewhere as well, so a pointer that information if it exists would be appreciated.

### DIFF
--- a/GROUPS.md
+++ b/GROUPS.md
@@ -1,0 +1,43 @@
+## Introduction
+
+This page contains a few initial notes on observation group requirements and design.  The idea is to ensure that we understand what we're trying to accomplish before we spend too much time doing an implementation.
+
+## Group Concept
+
+In the existing OCS groups are explicitly part of the science program hierarchy, cannot be nested, and have limited impact on scheduling.  An observation can be in exactly zero or one group.  For OCS3, our idea is to capture groups by tagging observations (and other groups?).  This is the same concept behind "labels" in gmail.  What appear to be folders of emails are in fact just collections of emails with the same label.  Groups will work in a similar way in OCS3, except that there will be different types of group each with their own characteristics and scheduling implications.  It remains an open question how to portray groups in the user interface.  Like gmail, one can imagine a view of folders of observations for instance but folders probably won't capture timing constraints in an intuitive way.
+
+## Nested Groups
+
+To make more sophisticated scheduling constraints possible, it should be possible to tag a group itself with a group label.  Unfortunately "nesting" groups also invites the possibility of creating recursive groups.  "Group A" can be tagged with "Group B" and vice versa. It isn't clear what, if anything, should be done to prevent this other than to report any impossible scheduling constraints that result from nested groups.
+
+## Group Types
+
+Groups can be used to solve various problems in science program design.  In this section we attempt to enumerate the various options that will be needed.
+
+* _Organizational Folders._  These are arbitrary groups of observations with no impact on scheduling.  They can be used to organize observations that are logically similar or in some way important to the user.
+
+* _Any Groups._ This is an unordered group with a number (or possibly a percentage) parameter.  The idea is that *n* of the items in the group must be completed before the group is considered complete.  For example, this would cover the possibility of choosing amongst various options of identical value.
+
+  - Putting two items in a group and requiring just one of them to be done is effectively an "or" group.
+
+  - Setting n to the number of items in the group (or setting the percentage to 100 if that's the way these are done) would effectively result in an "and" group.
+
+* _Timing Groups._  We express timing constraints between two observations and/or groups with a timing group.  A timing group has a min/max time bounds specifying how much time must or may pass between execution of two consecutive items.
+
+  - It would be useful to extend this to *n* ordered items but it isn't clear to me how, beyond building up a hierarchy.
+
+  - Maybe this is too simplistic.  Are there other types of timing constraints besides min / max time bounds?
+
+  - Explicit ordering of items without any additional timing consideration could be represented via a timing group with no max time bounds and a min time bounds of 0.
+
+## Acquisitions and Telluric Standards
+
+Two common group use cases that have been mentioned in the past are telluric standard stars and acquisition observations. After discussion with Bryan and Andy we have concluded that neither warrants special group types.
+
+### Acquisitions
+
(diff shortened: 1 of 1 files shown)


GROUPS.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Nested Groups
+
+To make more sophisticated scheduling constraints possible, it should be possible to tag a group itself with a group label.  Unfortunately "nesting" groups also invites the possibility of creating recursive groups.  "Group A" can be tagged with "Group B" and vice versa. It isn't clear what, if anything, should be done to prevent this other than to report any impossible scheduling constraints that result from nested groups.
+
```
tpolecat: If we're going to allow nesting then I think the tagging notion may not be sufficient.

GROUPS.md
```diff
@@ -0,0 +1,43 @@
...
+  - Explicit ordering of items without any additional timing consideration could be represented via a timing group with no max time bounds and a min time bounds of 0.
+
+## Acquisitions and Telluric Standards
+
+Two common group use cases that have been mentioned in the past are telluric standard stars and acquisition observations. After discussion with Bryan and Andy we have concluded that neither warrants special group types.
+
+### Acquisitions
+
+The need for separate acquisition observations is a reflection of poor modeling choices in the existing codebase.  In particular they are necessary so that (differently configured) acquisition and science observations can be found by the OT browser which only takes into account the static instrument configuration.  In the new model, the acquisition sequence can just be included with the science sequence in the same observation and the need for separate observations and therefore a grouping of some kind disappears.
+
```
tpolecat: We might note that steps *are* distinguished by time accounting category, so we can still tell which parts of a sequence are the acquisition and so on.

tpolecat:
I think maybe we do need more structure than what we can get with tagging, because we probably want to prevent for instance an observation from being a member of two independent "any" groups ... or do we?

GROUPS.md
tpolecat: Yeah if we have tagged tags I think it will start to be conceptually hard for users. Either we have nested folders or overlapping tags, but not both.

swalker2m:
> we probably want to prevent for instance an observation from being a member of two independent "any" groups

I was assuming no, but that's a good question for @andrewwstephens and Bryan.  Could we get in trouble by being a bit loose with restrictions?  I suppose it wouldn't be hard to set up constraints that are logically impossible to fulfill.  For example, if observation B must be done right after observation A according to one timing group and yet just the opposite is required in a different timing group.  Here we could presumably report the problems and let users sort it out?

tpolecat:
Maybe we could allow nested constraint groupings of the kinds you describe (sequencing and selection) to specify the program plan, and then maybe use tagging for arbitrary groupings for human use if that seems useful.

I think you can treat the program plan as a kind of regular expression. I played around with that a while back. I'll mess around with it after lunch.

tpolecat:
I played around with a plan data type yesterday. Here's a [gist](https://gist.github.com/tpolecat/b24cac4d8893bd4db7f88d71afb7716f). I think this is orthogonal to tagging.

swalker2m:
````scala
  // Ignore this for now. It's a missing library function that should be in Scalaz.
  def cata[F[_]: Functor, A](ffa: Free[F, A])(f: F[A] => A): A =
    ffa.resume match {
      case \/-(a)  => a
      case -\/(fa) => f(fa.map(cata(_)(f)))
    }
````

It's hard for me to ignore things in the codebase that I don't understand.  It seems to be some kind of reduce or fold?  From the documentation of `resume` I gather that it "evaluates a single layer of the free monad" but unfortunately I'm not sure what that means either.

**Comment by swalker2m** [comment]
> Rather than representing this as an inductive data type it ends up being more convenient to use a flat pattern functor and then use `Free` to apply the recursive structure and leaf values.

I'm certainly not contradicting this, but I also don't immediately see it.  In what way is it more convenient?

**Comment by swalker2m** [comment]
> In any case we have conjunctions of elements, which are ordered and must all be satisfied

I wanted to suggest ordering as well but how is it stored in the database?  We'll have to introduce an index column like sequence steps I suppose.  If you reorder something, sequence steps included, won't potentially all other elements have to be updated as well to change their indices?

**Comment by swalker2m** [comment]
> `PlanF` is a traversable functor.

Maybe a few words about this would help.

**Comment by tpolecat** [comment]
Ok working on a response to all of these :-)

**Comment by swalker2m** [comment]
`Free.roll`ing is new to me too.  I guess I'll stop now.  In general I'm not able to follow the suggestion very well.

**Comment by tpolecat** [comment]
Ok since I need to talk about the telnet code next week I'll put together a presentation on my proposed group model and we can decide if it's dumb or not. It's too much to write up here.

> I wanted to suggest ordering as well but how is it stored in the database? We'll have to introduce an index column like sequence steps I suppose. If you reorder something, sequence steps included, won't potentially all other elements have to be updated as well to change their indices?

I was kind of thinking the whole structure could be treated atomically (a modal editor in the UI) so we just replace the whole thing rather than trying to keep up with little edits that would require us to mess around with indices. If we did it that way then the indexing would be a storage detail and not something that would appear in the model. This would also give us a way to cancel the edit and retain old versions if we felt like it.

But if we do want to support incremental editing of the program plan then yeah we would need a representation of these indices in memory and would need to define all the edit operations pretty carefully.

**Comment by tpolecat** [comment]
I guess I'll say a tiny bit about `Free` that might clarify things a bit, but maybe not. The "real" definition of `Free` is:

```
case class Free[F[_], A](resume: A \/ F[Free[F, A]])`
           ─────┬───────                ─────┬────
                └──────────── <<< ───────────┘
```

Which means it's **either** an `A` **or** it's an `F` of "the same thing". Or in other words, a tree of `F` with `A`s at the leaves. For example a LISP s-expression can be represented as `Free[List, Atom]` ... arbitrarily nested lists of atoms.

In scalaz `resume` lets you inspect the structure and get the `\/` back, which was one of your questions; and `pure/roll` are scalaz's names for the left and right constructors. The structure is kind of obscured and the names are weird.

PR #18, opened 2016-11-14
Remove `tcc_value`

swalker2m:
This PR removes the misnamed `tcc_value` column from the various enumeration tables.  The `tcc_value` was the `String` value passed to the seqexec in today's OCS.  It will not be needed for the new seqexec which will work directly with the new sequence model.  Because the `tcc_value` was also used in translating old program XMLs, this PR also reworks the import code.  It combines the functionality of the `TCC` object with `ConfigReader` since it will only ever be needed during import and provides direct mapping from the old model to the new model without looking for `tcc_value` column entries.

Finally, it removes `create.sql` since we will be using flyway migrations in the `sql` directory.


modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -38,72 +51,186 @@ object ConfigReader3 {
...
+        new Key(name, write, read)
+
+      // Produces a pair of functions, one from old to new (S => A) and one
+      // from new to old (A => S)
+      private def xlat[S <: SequenceableSpType, A](name: String, m: Seq[(S, A)]): (S => A, A => S) = {
+        val oldToNew = m.toMap
+        val newToOld = m.map(_.swap).toMap
+
+        ((s: S) => oldToNew.getOrElse(s, sys.error(s"new value of '$name': $s not found")),
+         (a: A) => newToOld.getOrElse(a, sys.error(s"old value of '$name': $a not found")))
```
cquiroz: Do you think we won't use the old importer. It sounds to me like we may be requested to add forgotten programs from time to time

modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -38,72 +51,186 @@ object ConfigReader3 {
...
+        val key   = Key.this
+        override def toString: String =
+          s"Key.Entry[${clean(ev.tpe.toString)}]($path -> ${write(a)})"
+      }
+
+      def apply(a: A) = new Entry(a)
+    }
+
+    protected object Key {
+      def apply[A: TypeTag](name: String, write: Show[A])(read: Read[A]): Key[A] =
```
cquiroz: any reason why `read` is curried?

project/gen2.scala
```diff
@@ -168,8 +168,8 @@ object gen2 {
       },
 
       enum("Instrument") {
-        type InstrumentRec = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'tccValue -> String, 'obsolete -> Boolean`.T
-        val io = sql"select id, id tag, short_name, long_name, tcc_value, obsolete from e_instrument".query[(String, InstrumentRec)].list
+        type InstrumentRec = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'obsolete -> Boolean`.T
+        val io = sql"select id, id tag, short_name, long_name, obsolete from e_instrument".query[(String, InstrumentRec)].list
         io.transact(xa).unsafePerformIO
```
cquiroz: Feels a bit strange that we call `unsafePerformIO` inside the constructor. Why's that?

tpolecat:
Looks like you have to start from scratch with your db for this change, which is fine. Looks good, thanks. 👍

modules/importer/src/main/scala/ConfigReader.scala
swalker2m: Yeah, you're probably right.  We should keep it around for a little while regardless but at some point not too long after the transition we'll retire this code and stop maintaining it.

modules/importer/src/main/scala/ConfigReader.scala
swalker2m: I anticipate that the `Read` implementations will be involved in some cases and so it will be more convenient to write them this way.

project/gen2.scala
swalker2m: This is actually part of the build.  Each one of these is generating an enum data type from information in the database so it needs to do some IO.

PR #19, opened 2016-11-16
Gcal Lamp Update

swalker2m:
This PR updates the `GcalConfig` lamp type.  Previously, we had a single enumeration of lamps of which one was selected for the gcal observe:

````scala
case class GcalConfig(lamp: GcalLamp, shutter: GcalShutter)
````

In reality, there are two types of lamps, continuum and arc.  A gcal dataset can be produced with one type of lamp or the other.  Furthermore, multiple simultaneous arc lamps are permitted.  The new `GcalLamp` type is therefore `GcalContinuum \/ OneAnd[ISet, GcalArc]` where `GcalContinuum` contains just the continuum lamps and `GcalArc` just the arcs.

````scala
import GcalConfig.GcalLamp

case class GcalConfig(lamp: GcalLamp, shutter: GcalShutter)

object GcalConfig {
  type GcalLamp = GcalContinuum \/ OneAnd[ISet, GcalArc]
}
````

One way to accomplish this in the database model would have been to create a separate arcs table with which we would join when creating steps.  I found this somewhat challenging and decided on another approach which, admittedly, might be a bit lame.  Namely I added a `boolean` column for each of the four arc lamps.

````
CREATE TABLE step_gcal (
    index smallint NOT NULL,
    observation_id character varying(40) NOT NULL,
    continuum identifier,
    ar_arc boolean NOT NULL DEFAULT FALSE,
    cuar_arc boolean NOT NULL DEFAULT FALSE,
    thar_arc boolean NOT NULL DEFAULT FALSE,
    xe_arc boolean NOT NULL DEFAULT FALSE,
    shutter gcal_shutter NOT NULL,
    CONSTRAINT check_lamp CHECK ((continuum IS NULL) = (ar_arc OR cuar_arc OR thar_arc OR xe_arc)));
````

### CODE REVIEW

File: modules/core/src/main/scala/gem/config/GcalConfig.scala [hunk 1]
```diff
@@ -1,7 +1,21 @@
...
 
-case class GcalConfig(lamp: GCalLamp, shutter: GCalShutter)
+import scalaz.{\/, OneAnd}
+
+import GcalConfig.GcalLamp
+
+case class GcalConfig(lamp: GcalLamp, shutter: GcalShutter)
+
+object GcalConfig {
+  type GcalLamp = GcalContinuum \/ OneAnd[Set, GcalArc]
```
tpolecat: Did you consider `ISet` here? We would need to change `Enumerated` to extend `Ordered` but that seems ok to me. What do you think?

tpolecat:
Cool re: `ISet`. I think it's worth trying to be hard-asses about parametricity unless we have a good reason to compromise.

swalker2m:
Okay, switched to `ISet` and made a few other minor adjustments.  I think I'm finally finished tweaking this.  I'll merge it a some point before too long if there are no further comments.

PR #20, opened 2016-11-18
telnet server

tpolecat:
I have been sitting on this branch for a while, so here it is. Rebased, etc.

This adds a telnet server that we can use to interact with gem. In fact right now it's the only `main` implementation we have. My intent is that as we build things out we can add more commands and make this a reasonable way to interact with the server, especially for admin tasks but also for things we're working on that don't yet have a web UI. It is very easy to add a new command.

To run the telnet server you can do:

    sbt> telnetd/run

And them from a shell you can do

    telnet localhost 6666

And explore. You can log in with `root` and no password (you can then use `passwd` to change it if you like). Help, history, tab completion, etc., are built in. Commands are self-describing which hopefully will keep the documentation burden under control.


modules/db/src/main/scala/gem/dao/ProgramDao.scala
```diff
@@ -82,7 +82,7 @@ object ProgramDao {
     sql"""
       SELECT program_id, title
         FROM program
-       WHERE program_id like $pat OR title like $pat
+       WHERE lower(program_id) like lower($pat) OR lower(title) like lower($pat)
```
cquiroz: I guess this is not part of the telnet PR, right?

modules/db/src/main/scala/gem/dao/UserDao.scala
```diff
@@ -33,6 +33,7 @@ object UserDao {
              g.program_role
         FROM gem_user u LEFT OUTER JOIN gem_user_program g ON g.user_id = u.id
         WHERE id = $id
+        AND "md5" = md5($pass)
```
cquiroz: I'd like to avoid storing passwords, instead giving that business to ITS. We may want to have local password for development and eventually for some sort of admin user

In any case md5 hast of the password is likely too weak in this day. May I suggest something like `hmac` with  a key of at least 256 bits. I don't know how to do that in postgres though

modules/telnetd/src/main/scala/command/ls.scala
```diff
@@ -0,0 +1,57 @@
...
+           progs  <- d.queryProgramsByName(p, n + 1)
+           (h, t)  = progs.splitAt(n)
+           cols   <- getColumns
+           _      <- formatProgs(h, cols).traverseU(writeLn(_))
+           _      <- (t.nonEmpty).whenM(writeLn(s"--Limit reached. ($n)--"))
+         } yield d
+      }
+    ).zoom(Session.L.data[GemState])
+
+  // scala.text and kiama are both on the classpath but neither has any doc so I'm just going
```
cquiroz: What's kiama?

modules/telnetd/src/main/scala/main.scala
```diff
@@ -0,0 +1,21 @@
...
+import scalaz.effect._, scalaz.concurrent.Task
+
+/**
+ * Entry point for running Gem with a telnet server. This will go away at some point and the telnet
+ * server will be one of several services.
+ */
+object Main extends SafeApp {
+
+  // We need two transactors, one for things we do inside our telnet session and another for logging.
+  val xa  = DriverManagerTransactor[SessionIO]("org.postgresql.Driver","jdbc:postgresql:gem","postgres","")
```
cquiroz: How would the db password be provided?

modules/telnetd/src/main/scala/command/ls.scala
```diff
@@ -0,0 +1,57 @@
+package gem
+package telnetd
+package command
+
+import gem.dao.ProgramDao
+import net.bmjames.opts.{ intOption, help, metavar, short, long, value, strArgument }
```
cquiroz: What's bmjames?

build.sbt
```diff
@@ -92,3 +92,10 @@ lazy val sql = project
   .in(file("sql"))
   .settings(flywaySettings)
 
+lazy val telnetd = project
+  .in(file("modules/telnetd"))
+  .dependsOn(service)
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies += "org.tpolecat" %% "tuco-core" % "0.1.0"
```
cquiroz: I guess `tuco-core` is where the telnet server is?

modules/service/src/main/scala/gem/Service.scala
```diff
@@ -8,46 +8,45 @@ import gem.enum._
...
-
-  /**
-   * Fetch an observation with full sequence information. The use case is initial population of a
-   * sequence editor.
-   */
-  def getObservation(id: Observation.Id): M[Observation[Step[_ <: Instrument]]] =
-    log.log(user, s"getObservation($id)") {
-      ???
+  def changePassword(oldPassword: String, newPassword: String): M[Boolean] =
+    log.log(user, "changePassword(***, ***)") {
```
cquiroz: This maybe unrleated but what if I want to `log` inside the `log` block?

modules/service/src/main/scala/gem/Service.scala
```diff
@@ -8,46 +8,45 @@ import gem.enum._
...
+    def user[M[_]]: Service[M] @> User[ProgramRole] = Lens.lensu((a, b) => new Service(a.xa, a.log, b), _.user)
   }
 
+  def apply[M[_]](xa: Transactor[M], log: Log[M], user: User[ProgramRole]): Service[M] =
+    new Service(xa, log, user)
+
+  /**
+   * Construct a program that verifies a user's id and password and returns a `Service`.
+   */
+  def tryLogin[M[_]: Monad: Catchable: Capture](
```
cquiroz: Could you expand a bit on `Catchable` and `Capture` they seem to be defined in `scalaz`

modules/db/src/main/scala/gem/dao/UserDao.scala
tpolecat: Yeah that's a good idea. I just wanted something simple for development. We can swap out the methods on the `Service` object to do something else.

modules/telnetd/src/main/scala/command/ls.scala
tpolecat: Ben James is the author of `scala-optparse-applicative` which is the API I'm using to define the commands and parsers.

modules/telnetd/src/main/scala/main.scala
tpolecat: Presumably passed in as a commandline argument but I'm open to suggestions.

modules/service/src/main/scala/gem/Service.scala
tpolecat: You can do that if you want. `log(a)` has the same type as `a` so you could have `log(a *> log(b))` and it would work fine.

swalker2m:
I think the commands look great, but I also think it would take me a very long time to be able to write or significantly modify one on my own.  Perhaps we should schedule a time to go over this code in detail with everyone.

tpolecat:
> Perhaps we should schedule a time to go over this code in detail with everyone.

Good idea, I'll put something together for next week.

tpolecat:
FWIW there is some rudimentary doc on the telnet library [here](https://tpolecat.github.io/tuco/docs/) ... the **Command Shell To-Do List** attempts to explain the command business.

swalker2m:
Excellent, thanks!

sql/V003__Add_Root_User.sql
```diff
@@ -0,0 +1,7 @@
+
+-- Set the root user's password to '' if it hasn't been set. This will allow the
+-- first administrator to log in.
+UPDATE gem_user
+SET    md5 = md5('')
+WHERE  id  = 'root'
+AND    md5 = ''
```
swalker2m: I like flyway but I wonder if we shouldn't just put everything in one big initialization file until the first release.  After that, migrations will obviously be necessary.

modules/telnetd/src/main/scala/interact.scala
```diff
@@ -0,0 +1,73 @@
...
+   * Login loop. Prompt for a password `remainingTries` times and return a Service on success. The
+   * transactors are needed for the `Service`.
+   */
+  def loginLoop(
+    user:     String,
+    maxTries: Int,
+    xa:       Transactor[SessionIO],
+    txa:      Transactor[Task]
+  ): SessionIO[Option[Service[SessionIO]]] = {
+    def go(remaining: Int): SessionIO[Option[Service[SessionIO]]] =
```
swalker2m: `SessionIO[Option[Service[SessionIO]]]` ?

tpolecat: The login loop has telnet effects, so it's in `SessionIO`. 

If the login loop is successful it will yield a `Service` object, which provides service calls in an arbitrary IO-like effect type (could be `Task` for instance, and probably will be for the web front end). Here we're doing service calls from `SessionIO` in our command shell so that's the convenient effect type to use.

PR #21, opened 2016-11-21
Program DAO selectFull

swalker2m:
This PR contains a bug fix for the `StepDao` and a simple proposal for reworking `ProgramDao.selectFull`.    Previously `selectFull` referenced (or pretends to reference) all step fields in all tables in a single query building up a `Program[Observation[Step[_]]]` result.  This is clearly not complete now and not scalable going forward but the right way to do this also isn't obvious to me.  I'm proposing a straightforward change to multiple small queries.  If there is a better way, please let me know.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/ObservationDao.scala
+++ b/modules/db/src/main/scala/gem/dao/ObservationDao.scala
@@ -5,21 +5,31 @@ import gem.enum.Instrument
 
 import doobie.imports._
 
+import scalaz._, Scalaz._
+
 object ObservationDao {
 
   def insert(o: Observation[_]): ConnectionIO[Int] =
     sql"""
-      INSERT INTO observation (observation_id, 
-                              program_id, 
-                              observation_index, 
+      INSERT INTO observation (observation_id,
+                              program_id,
+                              observation_index,
                               title,
                               instrument)
-            VALUES (${o.id}, 
-                    ${o.id.pid}, 
-                    ${o.id.index}, 
+            VALUES (${o.id},
+                    ${o.id.pid},
+                    ${o.id.index},
                     ${o.title},
                     ${o.instrument})
     """.update.run
+
+  /** Select all the observation ids associated with the given program. */
+  def selectIds(pid: Program.Id): ConnectionIO[List[Observation.Id]] =
+    sql"""
+      SELECT observation_id
+        FROM observation
+       WHERE program_id = $pid
+    """.query[Observation.Id].list
 
   def selectFlat(id: Observation.Id): ConnectionIO[Observation[Nothing]] =
     sql"""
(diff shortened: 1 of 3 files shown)


modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -30,7 +40,7 @@ object ObservationDao {
        .unique
        .map { case (t, i) =>
          Observation(id, t, i, Nil)
-       }      
+       }
 
   def select(id: Observation.Id): ConnectionIO[Observation[Step[_]]] =
```
swalker2m: It seems to me that this should return `ConnectionIO[Option[Observation[Step[_]]]]` much like the corresponding select in `ProgramDao`.  No?

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -5,22 +5,32 @@ import gem.enum.Instrument
...
 
+  /** Select all the observation ids associated with the given program. */
+  def selectIds(pid: Program.Id): ConnectionIO[List[Observation.Id]] =
+    sql"""
+      SELECT observation_id
+        FROM observation
+       WHERE program_id = $pid
+    """.query[Observation.Id].list
+
   def selectFlat(id: Observation.Id): ConnectionIO[Observation[Nothing]] =
```
cquiroz: What does it mean to be `Observation[Nothing]`. is that only the observation data is contained?

swalker2m: Right, it is devoid of `Step` information for the sequence.  The idea (from Rob) was to be able to query for just the information you need rather than having to pull out everything any time you want anything.  The same scheme exists for `Program`, as in `Program[Nothing]` if you're not interested in observation details.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
tpolecat: I normally want to treat a bogus id as an error: if you have your hands on an id it must have come from the database originally so lookups shouldn't fail. But I can see it both ways. There could be two versions of the method I guess.

tpolecat:
I think in practice you will never want to call `selectFull` so it was mostly an experiment to see if it's possible. It will be much slower doing doing all the sub-selects but I'm ok with it as a simplifying thing. If it ends up being a performance issue we can deal with it later.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
swalker2m: As long as it is possible to delete observations the possibility of holding an ID that no longer exists is there.  I suppose though the idea is to simply mark them deleted rather than actually delete them.

swalker2m:
What prompted me to make this change was thinking about filling out the gcal information with all the missing fields and seeing the impact that would have on the `selectFull` implementation as it exists.  Add to that `n` instruments with all their fields and it didn't seem maintainable.

tpolecat:
Yep. In any case this looks fine. 👍

PR #22, opened 2016-11-23
Finish Gcal Step Definition

swalker2m:
This PR fills in missing gcal parameters to complete manual Gcal steps (and eventually expanded smart gcal steps).  It also switches the gcal shutter type to an enum table to be consistent with the other fields.

__Note:__ Once again I updated the `V001__Initial_Setup.sql` file rather than adding a new migration.  The reasoning is that (a) it seems better if all the initial program model is in a single file and (b) more importantly though I could add columns to `step_gcal` in a separate migration, programs have to be reimported anyway to pick up the data in the new fields.

### DIFF
--- a/modules/core/src/main/scala/gem/config/GcalConfig.scala
+++ b/modules/core/src/main/scala/gem/config/GcalConfig.scala
@@ -1,14 +1,16 @@
 package gem
 package config
 
-import gem.enum.{GcalArc, GcalContinuum, GcalShutter}
+import gem.enum.{GcalArc, GcalContinuum, GcalDiffuser, GcalFilter, GcalShutter}
+
+import java.time.Duration
 
 import scalaz._
 import Scalaz._
 
 import GcalConfig.GcalLamp
 
-case class GcalConfig(lamp: GcalLamp, shutter: GcalShutter) {
+case class GcalConfig(lamp: GcalLamp, filter: GcalFilter, diffuser: GcalDiffuser, shutter: GcalShutter, exposureTime: Duration, coadds: Int) {
   def continuum: Option[GcalContinuum] =
     lamp.swap.toOption
 
--- a/modules/db/src/main/scala/gem/dao/package.scala
+++ b/modules/db/src/main/scala/gem/dao/package.scala
@@ -4,7 +4,7 @@ import doobie.imports._
 import edu.gemini.spModel.core._
 
 import java.sql.Timestamp
-import java.time.Instant
+import java.time.{Duration, Instant}
 import java.util.logging.Level
 
 import scala.reflect.runtime.universe.TypeTag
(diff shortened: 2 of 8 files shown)


modules/core/src/main/scala/gem/config/GcalConfig.scala
```diff
@@ -1,14 +1,16 @@
...
+
+import java.time.Duration
 
 import scalaz._
 import Scalaz._
 
 import GcalConfig.GcalLamp
 
-case class GcalConfig(lamp: GcalLamp, shutter: GcalShutter) {
+case class GcalConfig(lamp: GcalLamp, filter: GcalFilter, diffuser: GcalDiffuser, shutter: GcalShutter, exposureTime: Duration, coadds: Int) {
```
tpolecat: Is the `exposureTime` a property of the instrument configuration, or part of the step itself? Is it ever absent?

modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -44,6 +44,9 @@ package object dao extends MoreTupleOps with ToUserProgramRoleOps {
   implicit val InstantMeta: Meta[Instant] =
     Meta[Timestamp].nxmap(_.toInstant, Timestamp.from)
 
+  implicit val DurationMeta: Meta[Duration] =
```
tpolecat: I wonder if we want to be using `scala.concurrent.duration.FiniteDuration` here, just to avoid JDK classes in the model since they may not work on Scala-JS. I don't have a strong opinion, just wondering if you had considered it.

cquiroz: `Duration` is certainly available in scala.js' time libs

tpolecat: Which one? JDK or Scala? Or both?

cquiroz: JDK, I assumed we were using java time API

tpolecat: Ok, sounds good. Just wanted to check.

modules/core/src/main/scala/gem/config/GcalConfig.scala
swalker2m: Good question, I think I may be repeating a flaw of the current program model.  In today's Science Program model you can configure the instrument exposure time separately from the calibration step exposure time.  You then actually get three properties in the sequence step:

1. `calibration:exposureTime`
2. `instrument:exposureTime`
3. `observe:exposureTime`

The calibration and instrument exposure times may differ but the `observe:exposureTime` always matches one or the other depending on step type.  So I think this may make more sense as part of the step itself.  It is never absent, though it can be 0 seconds.

The one mitigating factor is thinking ahead to smart gcal.  There we'll want to associate gcal step configurations with subsets of instrument configurations and I guess it would be convenient to have the entire thing in `GcalConfig`.  I suppose we could always pair `GcalConfig` with the exposure time.

Coadds are very similar.  Not all instruments (and therefore not all steps) support coadds though.

If it seems okay, I'll merge this as is and we can rework it as necessary?

tpolecat: Seems fine, yeah. It will become more clear as we build it out more.

modules/db/src/main/scala/gem/dao/package.scala
swalker2m: We use `java.time.Instant` elsewhere and `java.time.Duration` in the F2 config, fwiw.

tpolecat: Good, seems fine then.

PR #23, opened 2016-11-24
Step Index Discussion

swalker2m:
This is another content-free PR with the purpose of facilitating discussion.  In particular, I'm concerned about the step indices in the database model and have been trying to decide what if anything should be done.   My concerns are quite potentially due to ignorance.   Before I go down the wrong path, please weigh in if you have any thoughts.

### DIFF
--- a/StepIndex.md
+++ b/StepIndex.md
@@ -69,11 +69,13 @@ CREATE TABLE step_science (
 Ordering can take advantage of the Logoot index algorithm that allows steps to
 be inserted between any two other steps at any time without the need to update
 any other table row.  To do this, a `location` column with type `integer[]` could
-be added to the `step` table.  The database orders arrays on an element-by-element
-basis such that `[5, 1]` sorts before `[5, 2]` and both before `[10]`.  In fact,
-given sufficient spacing between step locations the need to introduce a second
-array element will rarely if ever arise.
+be added to the `step` table.   For example, to insert a step between steps at
+locations `[5]` and `[6]`, the location `[5, 5]` could be assigned for example.
 
+The database orders arrays on an element-by-element basis such that `[5]` sorts
+before `[5, 5]` and both before `[5, 8]`, or `[6]`.  In fact, given sufficient
+spacing between subsequent step locations the need to introduce a second array
+element will rarely if ever arise.
 
 ````sql
 CREATE TABLE step (
@@ -87,4 +89,6 @@ CREATE TABLE step (
 ````
 
 Note the `UNIQUE` constraint preventing two steps from residing at the same spot.
+There is no need, of course, to repeat the `location` column in the sub tables
+like `step_science`.
 
--- a/StepIndex.md
+++ b/StepIndex.md
@@ -60,7 +60,7 @@ CREATE TABLE step (
 );
 
 CREATE TABLE step_science (
-    step_science id SERIAL                PRIMARY KEY REFERENCES step ON DELETE CASADE,
+    step_science id integer               PRIMARY KEY REFERENCES step ON DELETE CASADE,
     observation_id  character varying(40) NOT NULL,
     ...
 );
--- a/StepIndex.md
+++ b/StepIndex.md
@@ -1,0 +1,90 @@
+## Step Indices Issue
+
+Sequence steps are modeled in Scala with a simple ADT:
+
+````scala
+sealed abstract class Step[A] extends Product with Serializable {
+  def instrument: A
+}
+
+final case class BiasStep   [A](instrument: A)                             extends Step[A]
+final case class DarkStep   [A](instrument: A)                             extends Step[A]
+final case class GcalStep   [A](instrument: A, gcal:      GcalConfig)      extends Step[A]
+final case class ScienceStep[A](instrument: A, telescope: TelescopeConfig) extends Step[A]
+final case class SmartStep  [A](instrument: A, smartCal:  SmartCalConfig)  extends Step[A]
+````
+
+The database model mirrors this simple hierarchy with a "base" `step` table:
+
+````sql
+CREATE TABLE step (
+    observation_id character varying(40) NOT NULL,
+    index          smallint              NOT NULL,
+    instrument     identifier            NOT NULL,
+    step_type      step_type             NOT NULL
+);
+````
+
+and "sub" tables for the various step types:
+
+````sql
+CREATE TABLE step_science (
+    index          smallint              NOT NULL,
+    observation_id character varying(40) NOT NULL,
+    offset_p       double precision      NOT NULL,
+    offset_q       double precision      NOT NULL
+);
+````
+
+Here the primary key of each of the tables is the combination `(observation_id, index)`,
+where matching values are used to stitch together the base and sub type rows into a
+complete step.  The primary key does double duty, uniquely identifying and also ordering
+steps, and yet the order is not fixed.  In fact one of the main selling points of a
+flattened sequence model is that it should permit steps to be arranged, reordered, and
+updated as necessary.  Imagine for example inserting a step at the beginning of a long
+sequence.  This would imply changing the primary key of every step in the sequence that
+follows.
+
+
+## Proposal
+
+The step table needs to track the `observation_id` identifying the observation to which
+it belongs but the primary key could be a unique id that is unrelated to the order of
+the step. This id then serves as a permanant link between the base and sub tables.
+
+````sql
+CREATE TABLE step (
+    step_id        SERIAL                PRIMARY KEY,
+    observation_id character varying(40) NOT NULL,
+    ...
+);
+
+CREATE TABLE step_science (
+    step_science id SERIAL                PRIMARY KEY REFERENCES step ON DELETE CASADE,
+    observation_id  character varying(40) NOT NULL,
+    ...
+);
+```` 
+
+Ordering can take advantage of the Logoot index algorithm that allows steps to
+be inserted between any two other steps at any time without the need to update
+any other table row.  To do this, a `location` column with type `integer[]` could
+be added to the `step` table.  The database orders arrays on an element-by-element
+basis such that `[5, 1]` sorts before `[5, 2]` and both before `[10]`.  In fact,
+given sufficient spacing between step locations the need to introduce a second
+array element will rarely if ever arise.
+
+
+````sql
+CREATE TABLE step (
+    step_id        SERIAL                PRIMARY KEY,
+    observation_id character varying(40) NOT NULL,
+    location       integer[]             NOT NULL,
+    instrument     identifier            NOT NULL,
+    step_type      step_type             NOT NULL, 
+    UNIQUE (observation_id, location)
+);
+````
+
+Note the `UNIQUE` constraint preventing two steps from residing at the same spot.
+
--- a/modules/core/src/main/scala/gem/Location.scala
+++ b/modules/core/src/main/scala/gem/Location.scala
@@ -10,15 +10,12 @@ import scalaz.Ordering.{EQ, GT, LT}
   * `Location` is a proper prefex of another, then it sorts ahead of the other
   * `Location`.
   */
-sealed abstract case class Location(toNel: NonEmptyList[Int]) {
+sealed abstract case class Location(toIList: IList[Int]) {
 
   // Use the sketchy sealed abstract case class technique to control the
   // NonEmptyList values that are passed to the constructor.  We want to
   // ensure that trailing Int.MinValue is always trimmed so that EQ and ==
   // agree.
-
-  def toIList: IList[Int] =
-    toNel.list
 
   def toList: List[Int] =
     toIList.toList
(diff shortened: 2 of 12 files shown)


StepIndex.md
```diff
@@ -0,0 +1,94 @@
...
+````sql
+CREATE TABLE step_science (
+    index          smallint              NOT NULL,
+    observation_id character varying(40) NOT NULL,
+    offset_p       double precision      NOT NULL,
+    offset_q       double precision      NOT NULL
+);
+````
+
+Here the primary key of each of the tables is the combination `(observation_id, index)`,
```
cquiroz: Where is the composite primary key defined?

StepIndex.md
```diff
@@ -0,0 +1,94 @@
...
+CREATE TABLE step_science (
+    step_science id integer               PRIMARY KEY REFERENCES step ON DELETE CASADE,
+    observation_id  character varying(40) NOT NULL,
+    ...
+);
+```` 
+
+Ordering can take advantage of the Logoot index algorithm that allows steps to
+be inserted between any two other steps at any time without the need to update
+any other table row.  To do this, a `location` column with type `integer[]` could
```
cquiroz: Is there an array type in postgres?

StepIndex.md
```diff
@@ -0,0 +1,94 @@
...
+be added to the `step` table.   For example, to insert a step between steps at
+locations `[5]` and `[6]`, the location `[5, 5]` could be assigned for example.
+
+The database orders arrays on an element-by-element basis such that `[5]` sorts
+before `[5, 5]` and both before `[5, 8]`, or `[6]`.  In fact, given sufficient
+spacing between subsequent step locations the need to introduce a second array
```
cquiroz: Right, couldn't you just give large spaces between the steps and avoid the integer array altogether?

StepIndex.md
swalker2m: Sorry, some parts are generated by `pgdump` which places the primary key definition elsewhere in the file:

````sql
ALTER TABLE ONLY step
    ADD CONSTRAINT step_pkey PRIMARY KEY (index, observation_id);
````

File: StepIndex.md (hunk 2 above)
Comment (swalker2m) [inline StepIndex.md]: Yes, postgres [supports array types](https://www.postgresql.org/docs/9.5/static/arrays.html).

File: StepIndex.md (hunk 3 above)
Comment (swalker2m) [inline StepIndex.md]: Since it can happen that you need to insert a step between two with consecutive indices, we have to handle it somehow.  I suppose an alternative to an array would be to use an integer and just renumber everything when necessary.

File: modules/core/src/main/scala/gem/Step.scala [hunk 4]
```diff
@@ -9,6 +9,37 @@ sealed abstract class Step[A] extends Product with Serializable {
...
+
+    private val LocationString = """^\{([0-9]+(?:,[0-9]+)*)\}$""".r
+
+    def parse(s: String): Option[Location] =
+      s match {
+        case LocationString(ds) => Some(Location(ds.split(',').toList.map(_.toInt)))
+        case _                  => None
+      }
+
+    def unsafeParse(s: String): Location =
```
cquiroz: Could/Should this instead return `Option[Location]` or a `\/`?

swalker2m:
Added some exploratory commits along the lines of the discussion document to make the proposal updates concrete.

We'd still some kind of `Sequence` to manage the location values.  Currently these are ignored other than sort order when read back in.

modules/importer/src/main/scala/Importer.scala
```diff
@@ -86,7 +86,7 @@ object Importer extends SafeApp {
         val configs = ss.flatMap(unsafeFromConfig)
 
         ObservationDao.insert(newObs) *>
-        configs.zipWithIndex.traverse { case (c, n) => StepDao.insert(newObs.id, n, c) }.void
+        configs.zipWithIndex.traverse { case (c, n) => StepDao.insert(newObs.id, Step.Location(n*100), c) }.void
```
swalker2m: Here the importer arbitrarily spaces out the step locations when reading in old data.

modules/core/src/main/scala/gem/Step.scala
swalker2m: There is a plain `parse` method for that.  In the end I don't think I really need to parse from `String` anyway so I'd probably just delete these methods altogether.

jdnavarro:
Just letting you know I'm following the conversation even though I don't know enough about the topic to provide any meaningful feedback.

It's not related to this particular PR but I'm specially interested in seeing how the polymorphic type parameter representing an *instrument* in the `Step` ADT will be used.

swalker2m:
@jdnavarro As a minimal example, check out [InstrumentConfig](https://github.com/gemini-hlsw/gem/blob/master/modules/core/src/main/scala/gem/config/InstrumentConfig.scala) which hints at how F2 would be defined.  So you could have, for example, `ScienceStep[F2Config]` for an F2 science step.

If you have your hands on an arbitrary `ScienceStep[A]`, I think you'd need to pull out the instrument and match on the expected instrument types.

modules/core/src/main/scala/gem/Step.scala
```diff
@@ -9,6 +9,37 @@ sealed abstract class Step[A] extends Product with Serializable {
   def instrument: A
 }
 
+object Step {
+  final case class Location(loc: List[Int]) {
+    override def toString: String =
+      loc.mkString("{", ",", "}")
+  }
```
tpolecat: We may end up using the same structure elsewhere (and it will certainly get built out more) so it might be useful to make a top-level data type.

sql/V001__Initial_Setup.sql
```diff
@@ -532,22 +532,22 @@ COMMENT ON TABLE semester IS '// TODO: start/end dates for site';
 --
 
 CREATE TABLE step (
-    observation_id character varying(40) NOT NULL,
-    index smallint NOT NULL,
-    instrument identifier NOT NULL,
-    step_type step_type NOT NULL
+    step_id        SERIAL                PRIMARY KEY,
```
tpolecat: We were talking about using a universal ID on steps so we could associate them with datasets without having to deal with step indexing, which this gives us. So 👍 for this.

tpolecat:
This looks great and it may be useful elsewhere. I do wonder what we should do about imported data since the step id is significant for associating datasets with steps. Presumably the steps with datasets will be marked complete so maybe we can number them consecutively rather than in steps of 100?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -0,0 +1,73 @@
...
+import scalaz.Ordering.{EQ, GT, LT}
+
+/** A sortable value used to indicate relative positions of a set of associated
+  * elements.  `Location`s may be thought of as lists of arbitrary integers
+  * which are be compared element by element such that the first pair of values
+  * that differ determine the ordering of the `Location`s as a whole.  If one
+  * `Location` is a proper prefex of another, then it sorts ahead of the other
+  * `Location`.
+  */
+sealed abstract case class Location(toNel: NonEmptyList[Int]) {
```
tpolecat: This is specialized to `NonEmptyList[Int]` but you could make it work for any `F[A]` where some operations require `Foldable1[F]` and/or `Order[A]`. This would allow you to define `Foldable[Location]` which would give you `.toList` and `.toIList` for free. You may not want to go this far but it's an option.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -0,0 +1,73 @@
...
+    trim(IList(is: _*))
+
+  // Remove trailing Int.MinValue so that == agrees with EQ.
+  private def trim(is: IList[Int]): IList[Int] =
+    is.dropRightWhile(_ == Int.MinValue)
+
+  def apply(h: Int, t: Int*): Location =
+    new Location(NonEmptyList.nel(h, trim(t))) {}
+
+  def fromList(is: List[Int]): Option[Location] =
```
tpolecat: You can define this as `def from[F[_]: Foldable](fi: F[Int])` so you don't need the `List` and `IList` variants, or just hang onto the passed foldable if you're generic in `F` as suggested above.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  def fromNel(n: NonEmptyList[Int]): Location =
+    new Location(NonEmptyList.nel(n.head, trim(n.tail))) {}
+
+  def unsafeFromList(is: List[Int]): Location =
+    fromList(is).get
+
+  private val LocationRegex = """^\{(\s*[-+]?[0-9]+\s*(?:,\s*[-+]?[0-9]+\s*)*)\}$""".r
+
+  def parse(s: String): Option[Location] =
```
tpolecat: Do we have a use case for `parse`?

modules/core/src/main/scala/gem/Location.scala
tpolecat: They're isomorphic but `IList` is invariant and all the methods are total and stacksafe. So it's a better `List`, or at least safer.

swalker2m:
> I do wonder what we should do about imported data since the step id is significant for associating datasets with steps.

I believe we will need a `dataset` table that has a `step_id` foreign key and a `dataset_label` or maybe just `index` column.  The step id doesn't imply anything about the order of the datasets.  The dataset index indicates the order though.  We could even use the start time of the dataset for ordering but the actual fits file will have a dataset label so it is probably useful to be explicit about it in the table.

On import we know which step to associate with each dataset from the observing log record.  Thereafter steps can be executed in any order, presumably.  The `location` field addressed in this PR doesn't come into play, I think. 

> Presumably the steps with datasets will be marked complete so maybe we can number them consecutively rather than in steps of 100?

You should be able to insert steps between two previously executed steps, I suppose.  At any rate I don't think there's a reason to prohibit it?  I'm inclined to just leave them equally spaced whether they are executed or not.

tpolecat:
Ok your answers re: imported data make sense. Thanks.

modules/core/src/main/scala/gem/Location.scala
swalker2m: I did want the non-emptiness of `NonEmptyList` but I suppose it could be `OneAnd`.

modules/core/src/main/scala/gem/Location.scala
swalker2m: Not really, I can just delete it until it is needed.

modules/core/src/main/scala/gem/Location.scala
tpolecat: You could demand `Foldable1` for methods that need the non-emptiness. It's really not a big deal. I suggest you merge it as-is and we can make it generic later if we feel like it.

tpolecat: Or just capture the `Foldable1` as a member I guess. Either way it's probably not worth worrying about right now.

swalker2m: 👍 didn't know about `Foldable1`.

swalker2m:
`Foldable` aside, it isn't clear how/whether to proceed from here.  It would now be possible to make minimal changes to the steps when adding / moving them around but how exactly isn't clear.  In the end I wonder if it is any better than just renumbering all the steps whenever anything about the order is changed.

I considered a `Sequence` class that keeps up with all the steps, their ids and their locations.  It could support operations like `List.patch` and others if useful, computing new locations to keep the order as expected.  Then to write the changes back to the database .... you need to know which steps had their locations updated by your edits.

I could instead expand the `Step` class with explicit knowledge of ids and locations and leave it up to the caller to do the updates.  That doesn't feel exactly right either.

tpolecat:
When we insert/move rows we only have to update the items that were moved, so it seems like a win to me. The "integrate" process in the server would reflect the algorithm directly which seems nice. It would be nice if we could fuse "delete and then re-insert" into a single update but it's just a constant factor … having to renumber for every atomic edit is N^2 and tracking/batching the edits to optimize renumbering sounds hard.

I think maybe I didn't understand your point.

modules/core/src/main/scala/gem/Location.scala
swalker2m: Given a generic `fi: F[Int]`, how do you produce a copy of `fi` with trailing values equal to `Int.MinValue` trimmed?  If `F` is also a `Monoid` then I could see how to do it, I suppose.  Or perhaps having `===` match `==` (which is the point of trimming trailing values) isn't a worthwhile goal?  Or perhaps I should change the `Order` implementation?

swalker2m:
> I think maybe I didn't understand your point.

I suppose we just need to include the `id` and `location` in the results returned with the `StepDao` queries then so that the client can manage locations.

modules/importer/src/main/scala/Importer.scala
```diff
@@ -83,10 +84,26 @@ object Importer extends SafeApp {
...
+        val obsLog  = Option(ObsLog.getIfExists(o)).fold(Map.empty[Int, OldDataset]) {
+                        _.getAllDatasetRecords.asScala.map(_.exec.dataset).map(d => d.getIndex -> d).toMap
+                      }
 
         ObservationDao.insert(newObs) *>
-        configs.zipWithIndex.traverse { case (c, n) => StepDao.insert(newObs.id, Location(n*100), c) }.void
+        configs.zipWithIndex.traverseU { case (c, n) =>
+          for {
+            id <- StepDao.insert(newObs.id, Location(n * 100), c)
+            _  <- obsLog.get(n).fold(Free.point[ConnectionOp, Unit](())) { oldDataset =>
```
swalker2m: I assume there's a better/shorter way than `Free.point[ConnectionOp, Unit](())` but couldn't find it.

sql/V001__Initial_Setup.sql
```diff
@@ -647,6 +658,23 @@ CREATE TABLE step_science (
...
+--
+
+CREATE TABLE dataset (
+    dataset_label   character varying(46)        PRIMARY KEY,
+    observation_id  character varying(40)        REFERENCES observation ON DELETE CASCADE,
+    dataset_index   id_index                     NOT NULL,
+    step_id         integer                      REFERENCES step        ON DELETE CASCADE,
+    filename        character varying(24)        NOT NULL,
+    "timestamp"     timestamp (5) WITH TIME ZONE NOT NULL
+    CONSTRAINT dataset_id_check CHECK ((dataset_label)::text = (((observation_id)::text || '-'::text) || trim(leading from to_char(dataset_index, '9000'))))
```
swalker2m: `trim(leading from to_char(dataset_index, '9000'))` is a bit weird, admittedly, but `to_char(dataset_index, '000')` fails if there are a 1000 or more datasets and I believe we have a few such observations.

modules/importer/src/main/scala/Importer.scala
tpolecat: `F[A]` is a monoid given `Applicative[F]` and `Monoid[A]` so you should be able to say

```scala
_ <- obsLog.get(n).foldMap { oldDataset => ...
```

But in general you can say `x.point[M]` so `().point[ConnectionIO]` would have worked too.

swalker2m:
Barring any objections, I'll go ahead and merge this rather than continuing to add new commits.  I'm not sure [how to switch `Location` to use a generic `Foldable`](https://github.com/gemini-hlsw/gem/pull/23#discussion_r90872354) but probably the biggest open question is about step ids and locations.  The client needs to be aware of them in order to edit, delete, move, or insert steps and yet I'm hesitant to include them in the `Step` Scala model.  I'll leave that TBD for now I suppose unless anybody has an opinion they'd like to share.

tpolecat:
> I'll leave that TBD for now I suppose unless anybody has an opinion they'd like to share.

That's fine. We can figure out how to externalize the id/loc so it doesn't infect the model. Don't worry about genericizing `Location`.

PR #24, opened 2016-11-25
Domains for coadds and exposure time

swalker2m:
A tiny update to provide domains for coadds and exposure time.


cquiroz:
Should these changes be reflected on the code as well?

swalker2m:
> Should these changes be reflected on the code as well?

I definitely think we should make Scala types for coadds and exposure time.  Those would best go into `spModel.core` I think.  The domains can also stand on their own though.  It will allow us to specify constraints on their values such that they don't have to be repeated in every instance.  I think that will be useful since many instruments have coadds and all have exposure time.

I'll add Scala types though and we can move them to `spModel.core` at some point.

PR #25, opened 2016-12-07
Add missing codecs.

tpolecat:
The `json` compilation tests were failing because `GCalConfig` now has a member of type `OneAnd[ISet, GcalArc]` and we didn't have codecs for `OneAnd` or `ISet`. This fixes that.

### DIFF
--- a/modules/json/src/main/scala/gem/json.scala
+++ b/modules/json/src/main/scala/gem/json.scala
@@ -5,6 +5,7 @@ import gem.enum.Instrument
 import gem.config._
 import java.time.Duration
 import argonaut._, Argonaut._, ArgonautShapeless._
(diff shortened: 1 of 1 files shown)


modules/json/src/main/scala/gem/json.scala
```diff
@@ -36,11 +37,19 @@ package object json {
     casecodec2(Observation.Id.apply, Observation.Id.unapply)("program-id", "index")
 
   // OffsetP maps to a signed angle in arcseconds
-  implicit val OffsetPMeta: CodecJson[OffsetP] =
+  implicit val OffsetPCodec: CodecJson[OffsetP] =
```
tpolecat: It can't do it for some reason, I think because `OffsetP` and `OffsetQ` extend `AnyVal`.

PR #26, opened 2016-12-12
Location refactor and find function.

swalker2m:
This PR updates `Location` to

* Add `Beginning` and `End` marker locations.  `Beginning` is a location that comes before all others and `End`, correspondingly, comes after all others.  You use the markers to insert elements at the front or end of a sequence.  For example, to put something at the beginning of the list you find a `Location` that falls between `Beginning` and the `Location` of the current first element.  Similarly you can find `Location`s between `Beginning` and `End` when inserting the initial elements of the sequence.

* Add `Middle` locations for all positions associated with elements in the sequence.

* Provide a `find` method for obtaining evenly spaced `Location`s between two given `Locations`.

### DIFF
--- a/modules/core/src/main/scala/gem/Location.scala
+++ b/modules/core/src/main/scala/gem/Location.scala
@@ -46,15 +46,15 @@ object Location {
 
   /** A Location that falls somewhere between the Beginning and End.
     */
-  sealed abstract case class Middle(posList: OneAnd[IList, Int]) extends Location {
+  sealed abstract case class Middle(posList: OneAnd[Vector, Int]) extends Location {
     def positions: Stream[Int] =
       posList.toStream #::: Stream.continually(Int.MinValue)
 
     def toIList: IList[Int] =
-      posList.head +: posList.tail
+      IList.fromFoldable(posList.head +: posList.tail)
 
     def toList: List[Int] =
-      posList.head :: posList.tail.toList
+      posList.head +: posList.tail.toList
 
     protected def minPrefixLength: Int =
       1 + posList.tail.length
@@ -82,30 +82,36 @@ object Location {
   // Constructors
 
   def apply(is: Int*): Location =
-    fromTrimmedIList(IList(is: _*))
+    fromTrimmedFoldable(is)(new Foldable[Seq] with Foldable.FromFoldr[Seq] {
+      def foldRight[A, B](fa: Seq[A], z: => B)(f: (A, => B) => B): B =
+        fa.foldRight(z) { (a, b) => f(a, b) }
+    })
 
   def fromFoldable[F[_]: Foldable](fi: F[Int]): Location =
-    fromTrimmedIList(IList.fromFoldable(fi))
+    fromTrimmedFoldable(fi)
 
-  private def fromTrimmedIList(is: IList[Int]): Location =
-    is.dropRightWhile(_ === Int.MinValue) match {
-      case INil()      => Beginning
-      case ICons(h, t) => new Location.Middle(OneAnd(h, t)) {}
+  def fromTrimmedFoldable[F[_]: Foldable](fi: F[Int]): Location = {
+    fi.foldRight(Vector.empty[Int]) { (i, v) =>
+      (v.isEmpty && i === Int.MinValue) ? v | i +: v
+    } match {
+      case h +: t => new Middle(OneAnd(h, t)) {}
+      case _      => Beginning
     }
+  }
 
   /** Assuming not all provided Ints are `Int.MinValue`, produces a `Middle`
     * `Location`.
     */
-  def unsafeMiddle(i: Int, is: Int*): Location.Middle =
+  def unsafeMiddle(i: Int, is: Int*): Middle =
     unsafeMiddle(i :: is.toList)
 
   /** Assuming not all provided Ints are `Int.MinValue`, produces a `Middle`
     * `Location`.
     */
-  def unsafeMiddle[F[_]: Foldable](fi: F[Int]): Location.Middle =
+  def unsafeMiddle[F[_]: Foldable](fi: F[Int]): Middle =
     fromFoldable(fi) match {
-      case m: Location.Middle => m
-      case _                  => sys.error("Location arguments do not form a Middle position: " + fi.toList.mkString(", "))
+      case m: Middle => m
+      case _         => sys.error("Location arguments do not form a Middle position: " + fi.toList.mkString(", "))
     }
 
 
@@ -121,7 +127,7 @@ object Location {
     * @return sorted list of `Location` where every element is GT l0 and LT l1
     *         (or vice versa if l0 is GT l1)
     */
-  def find(count: Int, start: Location, end: Location): IList[Location.Middle] = {
+  def find(count: Int, start: Location, end: Location): IList[Middle] = {
     val Zero  = BigInt(0)
     val One   = BigInt(1)
     val Max   = BigInt(Int.MaxValue)
@@ -133,18 +139,18 @@ object Location {
         (acc + (BigInt(i) + Min.abs) * pow, pow * Radix)
       }._1
 
-    def fromBase10(bi: BigInt): Location.Middle = {
+    def fromBase10(bi: BigInt): Middle = {
       @tailrec
-      def go(rem: BigInt, tail: IList[Int]): Location.Middle = {
+      def go(rem: BigInt, tail: Vector[Int]): Middle = {
         val (a, b) = rem /% Radix
         val head   = (b + Min).intValue
-        if (a === Zero) new Location.Middle(OneAnd(head, tail)) {} else go(a, head +: tail)
+        if (a === Zero) new Middle(OneAnd(head, tail)) {} else go(a, head +: tail)
       }
-      go(bi, IList.empty)
+      go(bi, Vector.empty)
     }
 
     @tailrec
-    def go(len: Int): IList[Location.Middle] = {
+    def go(len: Int): IList[Middle] = {
       val start10 = toBase10(start, len)
       val end10   = toBase10(end, len)
       val avail   = end10 - start10 - One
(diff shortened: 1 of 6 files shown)


modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+  // Constructors
 
   def apply(is: Int*): Location =
-    new Location(trim(is)) {}
+    fromTrimmedIList(IList(is: _*))
+
+  def fromFoldable[F[_]: Foldable](fi: F[Int]): Location =
+    fromTrimmedIList(IList.fromFoldable(fi))
+
+  private def fromTrimmedIList(is: IList[Int]): Location =
```
cquiroz: Can you define this from just `Foldable`?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+    is.dropRightWhile(_ === Int.MinValue) match {
+      case INil()      => Beginning
+      case ICons(h, t) => new Location.Middle(OneAnd(h, t)) {}
```
cquiroz: Why is `Middle` and abstract class?

modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -49,8 +49,8 @@ package object dao extends MoreTupleOps with ToUserProgramRoleOps {
   implicit val InstantMeta: Meta[Instant] =
     Meta[Timestamp].nxmap(_.toInstant, Timestamp.from)
 
-  implicit val LocationMeta: Meta[Location] =
-    Meta[List[Int]].nxmap(Location.fromList, _.toList)
+  implicit val LocationMeta: Meta[Location.Middle] =
+    Meta[List[Int]].nxmap(Location.unsafeMiddle(_), _.toList)
```
cquiroz: What's `nxmap`?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+    *
+    * @param count how many locations to find
+    * @param start starting location (exclusive)
+    * @param end ending Location (exclusive)
+    *
+    * @return sorted list of `Location` where every element is GT l0 and LT l1
+    *         (or vice versa if l0 is GT l1)
+    */
+  def find(count: Int, start: Location, end: Location): IList[Location.Middle] = {
+    val Zero  = BigInt(0)
```
cquiroz: What's `BigInt`? Is it related to `BigDecimal`?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+    */
+  case object Beginning extends Location {
+    protected def positions: Stream[Int] =
+      Stream.continually(Int.MinValue)
+
+    protected def minPrefixLength: Int =
+      0
+
+    override def toString: String =
+      "{α}"
```
swalker2m: Yes, I just needed a way to identify the start location.  Middle positions are a sequence of integers separated by commas and bracketed by '{' and '}'.

modules/core/src/main/scala/gem/Location.scala
swalker2m: Definitely but this is an internal method and `dropRightWhile` is just what I needed.

modules/core/src/main/scala/gem/Location.scala
swalker2m: To prevent anybody from directly using the constructor.  I'm trying to ensure that `==` and `===` provide the same answer in all cases.  See the definition of `OrderLocation` to see why that might not be the case otherwise.

modules/core/src/main/scala/gem/Location.scala
swalker2m: Right, it allows integer math without worrying about overflow.  The series of position integers can be thought of as digits in a number system with radix the entire range of `Int` values.

modules/db/src/main/scala/gem/dao/package.scala
swalker2m: From the doc it's an "invariant map".  I would appreciate a theoretical explanation of that myself.  @tpolecat would have a [better explanation](https://github.com/tpolecat/tpolecat.github.io/blob/master/_doobie-0.2.1/10-Custom-Mappings.md), but this is how I understand it in imprecise language:

We're telling doobie how to translate between `Location.Middle` classes and database types, which it can do because it already knows how to handle `List`.  The the 'n' is a reference to the fact that you can supply functions that don't worry about `null` values because if it sees a `null` it will return a `null`.  With this defined, doobie can return values from queries where a `Location.Middle` is expected and can accept a `Location.Middle` to insert in a row.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+  }
+
+  /** A Location that falls somewhere between the Beginning and End.
+    */
+  sealed abstract case class Middle(posList: OneAnd[IList, Int]) extends Location {
```
tpolecat: Remind me again, why not `NonEmptyList` here? We're not worried about serialization anymore I hope?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+      posList.toStream #::: Stream.continually(Int.MinValue)
+
+    def toIList: IList[Int] =
+      posList.head +: posList.tail
+
+    def toList: List[Int] =
+      posList.head :: posList.tail.toList
+
+    protected def minPrefixLength: Int =
+      1 + posList.tail.length
```
tpolecat: We might consider `OneAnd[Vector, Int]` here if this ends up being a performance issue. Looks like it's only used to compute new locations for insert though so maybe it doesn't matter.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,167 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+
+  implicit val OrderLocation: Order[Location] = Order.order(Function.untupled {
+    case (Beginning,  Beginning )                => EQ
+    case (End,        End       )                => EQ
+    case (Middle(m0), Middle(m1)) if (m0 === m1) => EQ
+    case (l0,         l1        )                =>
+      l0.positions.zip(l1.positions)
+        .find { case (a, b) => a =/= b }
+        .fold[Ordering](EQ) { case (a, b) => (a < b) ?[Ordering] LT | GT }
+  })
```
tpolecat: I like how you're using a stream here so you don't have to consider the "unequal length" case.

modules/db/src/main/scala/gem/dao/package.scala
tpolecat: If you have a parameterized type `case class Foo[A](f: A => A)` for instance, where `A` appears in both covariant ("return value") and contravariant ("argument") position and you then wish to derive a `Foo[B]` from it, you need to provide both `g: B => A` and `h: A => B` from which you can construct a `B => B` as `g andThen f andThen h`. Functors of this type are called exponential or invariant functors, which is where `xmap` came from … in cats it's called `imap`.

In any case because `Meta` provides mapping to read *and* write a column value it's an invariant functor like `Foo` above.

The `nxmap` is just a convenience that allows you to ignore the `null` case, since JDBC speaks in terms of `null` and we don't. The next-higher abstraction `Atom` doesn't allow you to see `null`; from `Meta[A]` you get both `Atom[A]` and `Atom[Option[A]]` for the non-nullable and nullable cases, respectively, and user-level doobie code never observes `null`.

More generally you can see invariant functors as a supertype, where covariant and contravariant functors as specializations that ignore one or the other of the `xmap` arguments. There is a fourth type called phantom functors which ignore *both* arguments and are thus both covariant *and* contravariant, which would be the case for something like `case class Phantom[A](i: Int)` where `A` is never used so it can vary in either direction.

swalker2m:
Switched to `OneAnd[Vector, Int]`, though in reality the `tail` of the integer position list should usually be empty.  If it every got deep enough to matter, either the sequence is astronomically long or else something is wrong.

tpolecat:
> Switched to OneAnd[Vector, Int], though in reality the tail of the integer position list should usually be empty. If it every got deep enough to matter, either the sequence is astronomically long or else something is wrong.

If that's the case then I might prefer `NonEmptyList` since it performs better for very small sizes. Fine either way though.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -10,46 +11,171 @@ import scalaz.Ordering.{EQ, GT, LT}
...
+    override def toString: String =
+      "{ω}"
+  }
+
+  // Constructors
+
+  private implicit object FoldableSeq extends Foldable[Seq] with Foldable.FromFoldr[Seq] {
+    def foldRight[A, B](fa: Seq[A], z: => B)(f: (A, => B) => B): B =
+      fa.foldRight(z) { (a, b) => f(a, b) }
+  }
```
swalker2m: Moved back to `NonEmptyList` so I'll merge this now.

One question though, for my own edification, is about `FoldableSeq`.  `Seq` is not a great sequence since it can be anything, infinite, mutable, etc. but it is what you get when you use varargs.   I went ahead and defined a basic `Foldable[Seq]` since it doesn't seem to exist (?) and avoid converting to a `List` or something but wanted to check why this is necessary and whether it's a good idea or not.  I keep it private here so it doesn't leak out.

tpolecat: Yes, scalaz has no instances for `Seq` because it's mutable. It's less of an issue for `Foldable` because it's hard to identify laws, but still doesn't really fit with the way scalaz thinks about things.

In 2.13 the default `Seq` is going to become an alias for `immutable.Seq` so with some luck this will be fixed. Anyway not a big deal if it's local.

PR #27, opened 2016-12-13
Smart GCal  Steps

swalker2m:
Adds a table, etc., for smart gcal steps.  Normalizes the various combinations of spelling and capitalization to SmartGcal for consistency.

__Note:__ we cannot distinguish smart steps from normal gcal steps in imported data, but I propose that the import just treat all existing smart steps as manual gcals.  If that doesn't turn out to be sufficient, we'll need an update to the old sequence model to inject some kind of "smart" key into the smart steps.

### DIFF
--- a/modules/core/src/main/scala/gem/config/SmartGcalConfig.scala
+++ b/modules/core/src/main/scala/gem/config/SmartGcalConfig.scala
@@ -1,0 +1,6 @@
+package gem
+package config
+
+import gem.enum.SmartGcalType
+
+final case class SmartGcalConfig(smartGcalType: SmartGcalType)
--- a/modules/core/src/main/scala/gem/config/SmartGcalConfig.scala
+++ b/modules/core/src/main/scala/gem/config/SmartGcalConfig.scala
@@ -1,6 +1,0 @@
-package gem
-package config
-
-import gem.enum.SmartGcalType
-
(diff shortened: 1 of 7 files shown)


tpolecat:
> I propose that the import just treat all existing smart steps as manual gcals.

👍 if we can get away with this. The objection will be "now I can't copy/paste from my old program" but I guess we can cross that flaming bridge when we get to it.

modules/core/src/main/scala/gem/config/SmartGcalConfig.scala
```diff
@@ -0,0 +1,6 @@
+package gem
+package config
+
+import gem.enum.SmartGcalType
+
+final case class SmartGcalConfig(smartGcalType: SmartGcalType)
```
tpolecat: It seems a little strange to have a case class that wraps an enum. If there is no other structure associated with `SmartGcalConfig` could we just use `SmartGcalType` directly rather than wrapping it? If not it seems like this type might work better as an enumerated ADT and then only use `SmartGcalType` for serialization. I don't have a strong opinion. What do you think?

swalker2m:
> It seems a little strange to have a case class that wraps an enum.

Agreed.  I just removed it.  If we add more structure to smart gcal in the future we can revive it.

PR #28, opened 2016-12-13
Warning Removal

swalker2m:
This tiny PR removes all compile warnings (in conjunction with the SmartGcal step PR which handles a couple "match may not be exhaustive" warnings).  Most warnings were about using higher kinded types so I just enabled that everywhere instead of importing it into individual classes.  We're all about high kinds.

I think we should soon enable `-Xfatal-warnings` as well.  Warnings tend to pile up if you don't deal with them, as you can see in the "ocs" project build.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,9 +16,19 @@ lazy val testLibs = Seq(
 lazy val commonSettings = Seq(
   scalaVersion := "2.11.8",
   scalacOptions ++= Seq(
-    "-feature",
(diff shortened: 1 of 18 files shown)


swalker2m:
Went ahead and turned on fatal warnings and most of the others from [this guy's blog post](https://tpolecat.github.io/2014/04/11/scalac-flags.html).

tpolecat:
👍

PR #29, opened 2016-12-13
Enable idea to find generated source code

cquiroz:
This little change on the build let's idea on my machine to find the generated code automatically. Would be good that somebody else could test it


tpolecat:
The generated sources show up for me but there's no Scala SDK, whatever that means, so it won't compile. I'm giving up but Shane may want to try.

PR #30, opened 2016-12-14
WIP: docker support

tpolecat:
Work in progress on docker deployment. Still limited but finally I think it's starting to make sense. This doesn't affect Gem all that much with the exception of the migration scripts, which I moved into a resources directory so the app (telnetd for now) can run migrations on startup.

If you click "View" on the markdown doc you can see what I have been up to.

I have never written a shell script before so it's probably terrible. Sorry.

TODO:
- [ ] Add health checks for containers.
- more


cquiroz:
I think there is a need of a general conversation between the approach on seqexec (rpms) and this one (docker). I'd rather have a single one unless there is a very good reason not

I pressume ITS would rather have rpms but perhaps docker is better for us? I'm not sure

tpolecat:
The nice thing about docker is that it permits safe database migrations. If an upgrade fails or is found to be buggy we can just turn it off and turn on the old version and we're back up and running in a few seconds. I don't know how you would manage that with RPMs without building out a lot of tooling.

docker.md
```diff
@@ -0,0 +1,184 @@
...
+
+This document explains how to deploy Gem inside docker on localhost. A next step is getting this tooling to deploy and manage remote machines, but it doesn't do that yet.
+
+### About Docker
+
+Docker ("docker engine") is a system service that manages **containers**. A container is a program that runs with its own Linux environment, but all containers on a machine share the same kernel. So it's kind of like a VM but it's better to think of as a unit of isolation for applications.
+
+Containers are based on **images** which are immutable filesystem layers that define an application's environment. The application itself is added on top of an existing base image like `ubuntu` or `centos` for instance. These are publicly available.
+
+The running state of our system is two docker containers, one running Gem (just telnetd for now) and one running Postgres. The Postgres container uses the `postgres:9.6.0` image which is maintained by Postgres. The Gem container uses the `java:latest` base image with our application added on via sbt-native-packager, yielding (for now) an image called `telnetd:0.1-SNAPSHOT`.
```
swalker2m: I was surprised to not see a Dockerfile anywhere.  It's all done through the magic of the `JavaAppPackaging` plugin and a bit of config in `build.sbt`?

jdnavarro:
I still don't fully understand the constraints we have with ITS, but do we have control over the physical hardware we could deploy to? If we could install any OS we want, we could opt for a specialized container host to take advantage of a nicer API to manage the lifecycle of containers (I have something like [CoreOS](https://coreos.com/) or [SmartOS](https://www.joyent.com/smartos) in mind). I heard lots of cumbersome workarounds related to systemd and docker daemon fighting over the control over of the running services in the host.

Perhaps they are still too immature, but there are standardization efforts for container APIs so we don't lock ourselves to Docker, the implementation. See https://www.opencontainers.org/ https://containerd.io/. In the future, if there is anything already, I'm pretty sure someone will create the JVM bindings for this standard APIs and we could use a sane language like Scala, instead of relying on shell script subprocesses.

Regarding the guest OS, do we need a full blown OS with all the RPMs and everything that comes with it? I'm successfully running several experiments at home several containers with [Alpine Linux](https://alpinelinux.org/) on SmartOS. There are more initiatives of distros getting rid of packages and coming with deployment workflows involving just image diffs. For example [Ubuntu Snappy](https://developer.ubuntu.com/en/snappy/) and [Project Atomic](https://www.projectatomic.io/) from RedHat, but I don't really know how much real world use these distros have had.

The PostgreSQL container might still require a full blown OS. I have in mind the application code requiring just the JVM and little more for these minimal image containers without RPMs.

swalker2m:
I agree with Carlos that whatever we do, it has to be consistent across all of our applications.  I'm new to docker and containers in general, but I think this looks very promising.  It will be running on the VM setup that IT provides us, so we need to check that that will be kosher but otherwise it looks good to me.  Will this help with setting up environments for testing?

tpolecat:
> Will this help with setting up environments for testing?

Yes, I'm thinking maybe one production VM and one test VM per site, and of course anyone can spin up an environment locally (and the nice thing about Docker is that they're identical so you shouldn't have weird OS/Charset/Timezone/JVM-version issues to deal with). The test setup can automatically deploy on every push to master and production could be on-demand (although ideally it will track master pretty closely).

docker.md
tpolecat: Right, it does all that for you. You can customize it as much as you like but the default works so I haven't messed with it that much.

tpolecat:
> I still don't fully understand the constraints we have with ITS, but do we have control over the physical hardware we could deploy to?

My understanding is that we must use CentOS and don't have root access. Docker frees us from that constraint (or any others that might be imposed) and lets us deploy whatever we want.

> Perhaps they are still too immature, but there are standardization efforts for container APIs so we don't lock ourselves to Docker

It's certainly worth keeping an eye on this. Hopefully our usage will remain so simple that we're not all that locked in.

> Regarding the guest OS, do we need a full blown OS with all the RPMs and everything that comes with it?

I don't think we have much choice about what's provisioned on the VM but the containers only have the bare minimum.

tpolecat:
> I agree with Carlos that whatever we do, it has to be consistent across all of our applications.

This is a prototype and I think it's worth experimenting with different approaches. It may not work. But I think our current setup doesn't work well either. Deployments are very complicated and risky and I want to make it routine rather than terrifying.

cquiroz:
I'd be quite happy to get rid of the rpms. As we have recently seen they can be quite unpredictable. Also they require a lot of work on the build

swalker2m:
I don't know if you wanted to evolve this more before merging, but if not 👍.  Since it changes the structure of the flyway stuff, it might be better to have it happen sooner than later.

cquiroz:
I agree with @swalker2m, code-wise seems ok. If we are going this way, @jdnavarro should probably stop working on rpms and continue with this idea for the seqexec.

@arturog8m How could we decide whether it is ok/desirable to use docker for our apps?

tpolecat:
> I don't know if you wanted to evolve this more before merging, but if not 👍. Since it changes the structure of the flyway stuff, it might be better to have it happen sooner than later.

Oh, a note on that. Deployments rely on schema evolution and will fail unpredictably if you change an old script rather than adding a new one. So when we get auto-deployments working we will need to start using evolution to make schema changes (or destroy the world and start over when this happens). Long before we intend to ship this software we need to be deploying it as we intend to do in production which means getting used to schema evolution.

But re: merging, yes I will do that since it will block you on schema changes,

tpolecat:
> If we are going this way, @jdnavarro should probably stop working on rpms and continue with this idea for the seqexec.

> @arturog8m How could we decide whether it is ok/desirable to use docker for our apps?

My intent was to experiment with this and see how it goes, and not change anything else unless it works well and we're confident that it does what we want.

tpolecat:
Merging, and this is good enough to play with but I'm continuing work on it.

PR #31, opened 2016-12-16
SmartGcal F2

swalker2m:
A work-in-progress PR for building out SmartGcal.  To summarize

* I envision a drastically simpler smart gcal infrastructure in the new model.  We'll keep smart gcal information in database tables rather than CSV files in SVN and there will be no upload servlet or verification checks, etc.

* Before it can be executed, a smart gcal step has to be expanded into a normal/manual gcal step.  At that point the smartness disappears and the mapping in the database can change at any time without impacting executed steps.

This PR,

* Moves gcal columns of `step_gcal` into a table of just `gcal` data.  `step_gcal` just links a step index with a `gcal` id in the `gcal` table.

````
gem=# \dS step_gcal
      Table "public.step_gcal"
    Column    |  Type   | Modifiers 
--------------+---------+-----------
 step_gcal_id | integer | not null
 gcal_id      | integer | not null
Indexes:
    "step_gcal_pkey" PRIMARY KEY, btree (step_gcal_id)
Foreign-key constraints:
    "step_gcal_gcal_id_fkey" FOREIGN KEY (gcal_id) REFERENCES gcal(gcal_id) ON DELETE CASCADE
    "step_gcal_step_gcal_id_fkey" FOREIGN KEY (step_gcal_id) REFERENCES step(step_id) ON DELETE CASCADE
````

* Creates a `smart_f2` to hold F2 information pertinent to expanding smart gcal steps.  It too links to a `gcal` row.

````
gem=# select * from smart_f2;
 lamp | baseline |  disperser  | filter |    fpu    | gcal_id 
------+----------+-------------+--------+-----------+---------
 Arc  | Night    | R1200JH     | JH     | LongSlit1 |       1
 Arc  | Night    | R1200JH     | JH     | LongSlit2 |       2
...
````

* Adds an importer of smart gcal data from the existing csv files for F2.  Others should be "easy" to add.  This means that `importer/run` now prompts for which importer to run, but we can make the program importer the default again if that's annoying.

### CODE REVIEW

File: modules/db/src/main/scala/gem/dao/SmartGcalDao.scala [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+import gem.config.{F2SmartGcalKey, GcalConfig, SmartGcalKey}
+import gem.enum.{GcalBaselineType, GcalLampType}
+import doobie.imports._
+
+object SmartGcalDao {
+  def insert(l: GcalLampType, b: GcalBaselineType, k: SmartGcalKey, g: GcalConfig): ConnectionIO[Int] = {
+    def insertSmartF2(gcalId: Int, k: F2SmartGcalKey): ConnectionIO[Int] =
+      sql"""
+        INSERT INTO smart_f2 (lamp, baseline, disperser, filter, fpu, gcal_id)
+             VALUES ($l :: gcal_lamp_type, $b :: gcal_baseline_type, ${k.disperser}, ${k.filter}, ${k.fpu}, $gcalId)
```
cquiroz: Can you explain a bit the syntak `$1 :: gcal_lamp_type` is it a custom postgress type?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -0,0 +1,24 @@
...
+      """.update.run
+
+    for {
+      id <- GcalDao.insert(g)
+      r  <-
+        k match {
```
cquiroz: Why do you need the match, isn't k already a `F2SmartGcalKey`?

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -141,22 +143,24 @@ object StepDao {
...
+             gc.shutter,
+             gc.exposure_time,
+             gc.coadds,
              sc.offset_p,
              sc.offset_q,
              ss.type
         FROM step s
              LEFT OUTER JOIN step_gcal sg
                 ON sg.step_gcal_id    = s.step_id
+             LEFT OUTER JOING gcal gc
```
cquiroz: There maybe a typo on `JOING`

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -141,22 +143,24 @@ object StepDao {
...
+             gc.filter,
+             gc.diffuser,
+             gc.shutter,
+             gc.exposure_time,
+             gc.coadds,
              sc.offset_p,
              sc.offset_q,
              ss.type
         FROM step s
              LEFT OUTER JOIN step_gcal sg
```
cquiroz: It still worries me having so many joins. Are we still checking for performance?

modules/importer/src/main/scala/SmartGcalImporter.scala
```diff
@@ -0,0 +1,145 @@
...
+  // -------------------------------------------------------------------------
+  // Implementation Details
+  // -------------------------------------------------------------------------
+
+  val dir = new File("smartgcal")
+  val xa  = Importer.xa
+  val lxa = Importer.lxa
+
+  val checkSmartDir: IO[Unit] =
+    IO(dir.isDirectory).flatMap { b =>
```
cquiroz: Nice!

project/gen2.scala
```diff
@@ -175,6 +175,26 @@ object gen2 {
         io.transact(xa).unsafePerformIO
       },
 
+      enum("GcalBaselineType") {
+        type GcalBaselineTypeRec = Record.`'tag -> String`.T
```
cquiroz: Will this generate the enumeration from the values on the db?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
swalker2m: `gcal_lamp_type` is an `ENUM`:

````
CREATE TYPE gcal_lamp_type AS ENUM (
    'Arc',
    'Flat'
);
````

In `smart_f2` the `lamp` column is given type `gcal_lamp_type` but here I'm inserting text so a cast is needed.  I'm not sure if there's a way to work around this.  There seems to exist a `CREATE CAST` statement that might be used to make this implicit?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
swalker2m: The key is a `SmartGcalKey`.  Right now, there's only the one instance, `F2SmartGcalKey`, but presumably there will be one for every instrument that can do smart gcal.  They are subsets of the full instrument config.  Only values that go into determining the calibration configuration are needed.  In the case of F2, that is the disperser, filter and FPU.

So, every instrument that supports smart gcal would need to be added to the match statement and an `INSERT` statement specific to its type will be needed.  Of course, if you create an instance of `SmartGcalKey` but fail to add it to the `match` statement the compiler will flag it as an error.

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: It worries me too.  I don't think this is maintainable going forward.  I think this was just done as an initial experiment.

Going forward I assume we'll need two queries.  The first will determine the step type and then the second will read the information specific to that step type.  Already we aren't really doing a "selectAll" -- note that only the instrument name is included not all the instrument configuration.  It might be feasible with just one instrument, but with just a handful we won't be able to keep this implementation.

project/gen2.scala
swalker2m: Right, this is the code generation for `ENUM` types in the database.  We have a second kind of enumeration that are based on normal database tables.  For example `F2Disperser` is one of these.  The code generation is slightly different for these but the resulting Scala code is similar.

The distinction is a bit blurry but I am using `ENUM` types for concepts that shouldn't change like Night or Day baseline calibration.  Table-based enumerations are used for instrument features that can be obsoleted or extended with new values in the future.  I'm not sure though whether to keep that distinction going forward or to just always use the table-based enumerations.  As you noticed, the `ENUM` types require explicit casting.  The table-based enums on the other hand lead to text columns with constraints on the value.

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
tpolecat: I think this is the only way to do it in a way that JDBC understands. The value is sent as a string.

modules/db/src/main/scala/gem/dao/StepDao.scala
tpolecat: The join will be much more efficient than doing a subquery for each step. I was thinking we would do a second query for the instrument slices because it will be the same instrument for the whole sequence. So, 2 selects rather than N.

project/gen2.scala
tpolecat: I was using `ENUM` for types that have no structure other than being distinct from one another (i.e., when you would use `Symbol` in Scala). But it may be easier in the end to use lookup tables everywhere.

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: Thanks, that makes sense.  As long as the instrument stuff is kept to a separate select, the query shouldn't get much worse than this.

PR #32, opened 2016-12-19
Smart gcal select

swalker2m:
A small PR that updates the `SmartGcalDao` to add a select by key and `SmartGcalType`.  The type is one of `Arc`, `Flat`, `DayBaseline`, or `NightBaseline`.  Each entry in a smart gcal table like `smart_f2` is tagged with a `lamp` type which is `Arc` or `Flat`, and a `baseline` type which is `Day` or `Night`.  Depending on the `SmartGcalType` (and the instrument fields), the row is selected or not.

````
gem-# select * from smart_f2;
 lamp | baseline |  disperser  | filter |    fpu    | gcal_id 
------+----------+-------------+--------+-----------+---------
 Arc  | Night    | R1200JH     | JH     | LongSlit1 |       1
 Arc  | Night    | R1200JH     | JH     | LongSlit2 |       2
 Arc  | Night    | R1200JH     | JH     | LongSlit3 |       3
...
````

This PR highlights a problem that I don't know how to address.  The issue is that the most efficient query would join the instrument key table (e.g., `smart_f2`) with `gcal` to produce an immediate result.  Unfortunately that would require duplicating the 10 fields of `gcal` for each instrument and type (lamp vs. baseline).  In other words, duplicating 10 gcal fields (2 x #instruments) times in the code.

Rather than do that, I just run separate queries:  an initial query of `smart_f2` for all the matching `gcal_id` and then the `gcal` table with the appropriate ids.  I think there must be a better way?

### DIFF
--- a/modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
+++ b/modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
@@ -1,18 +1,11 @@
 package gem.dao
 
 import gem.config.{F2SmartGcalKey, GcalConfig, SmartGcalKey}
-import gem.enum.{GcalBaselineType, GcalLampType, SmartGcalType}
+import gem.enum._
 import doobie.imports._
 import scalaz._, Scalaz._
 
 object SmartGcalDao {
-  def split(t: SmartGcalType): GcalLampType \/ GcalBaselineType =
-    t match {
-      case SmartGcalType.Arc           => GcalLampType.Arc.left
-      case SmartGcalType.Flat          => GcalLampType.Flat.left
-      case SmartGcalType.DayBaseline   => GcalBaselineType.Day.right
-      case SmartGcalType.NightBaseline => GcalBaselineType.Night.right
-    }
 
   def selectF2(k: F2SmartGcalKey, t: SmartGcalType): ConnectionIO[List[Int]] = {
     def byLamp(l: GcalLampType): ConnectionIO[List[Int]] =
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: modules/db/src/main/scala/gem/dao/SmartGcalDao.scala [hunk 1]
```diff
@@ -1,10 +1,45 @@
...
+      sql"""
+        SELECT gcal_id
+          FROM smart_f2
+         WHERE baseline  = $b :: gcal_baseline_type
+           AND disperser = ${k.disperser}
+           AND filter    = ${k.filter}
+           AND fpu       = ${k.fpu}
+      """.query[Int].list
+
+    t.fold(byLamp, byBaseline)
```
swalker2m: Here the `byLamp` and `byBaseline` queries are nearly identical, so it's a shame not to abstract this somehow.

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -1,10 +1,45 @@
...
+  }
+
+  def select(k: SmartGcalKey, t: SmartGcalType): ConnectionIO[List[GcalConfig]] =
+    for {
+      ids <- k match {
+              case f2: F2SmartGcalKey => selectF2(f2, t)
+            }
+      gcs <- ids.traverseU { GcalDao.select }.map(_.flatten)
+    } yield gcs
```
swalker2m: Here a more efficient query would join `smart_f2` and `gcal` for an immediate result.  Not sure how to do this w/o duplicating all the information in `GcalDao.select` for each instrument / type combination.

tpolecat: I just added composable statement fragments to doobie so we may be able to remove the repetition.

tpolecat:
This looks fine. As I noted I think we can get rid of the copy/paste (and also add some checking that will catch typos as in your previous PR). I'll work on this in the new year.

PR #33, opened 2016-12-21
Make StepDao.selectAll select all

swalker2m:
This PR updates the `StepDao.selectAll` method to return steps containing their corresponding instrument configuration.  It also adds a sample application to exercise the `selectAll` function.  Outstanding issues include:

* Each observation is associated with an instrument, and all steps will have a corresponding instrument configuration.  You supply the expected instrument in the call to `selectAll`.  If you are correct all is well.  If you are wrong the query will generate an exception.   There are at least two ways to fix that:

  1. Don't specify the instrument and instead query the observation to read it.

  2. Amend the instrument-specific queries like `selectF2` to anticipate `null` values in the join.  Null rows could be skipped resulting in an empty list of steps instead of an exception.

* Sample applications are fine for development but we need to figure out a testing environment to use with the database.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/StepDao.scala
+++ b/modules/db/src/main/scala/gem/dao/StepDao.scala
@@ -139,7 +139,12 @@ object StepDao {
       }
   }
 
-  def selectAll(oid: Observation.Id): ConnectionIO[List[Step[Instrument]]] =
+  /** Selects `Step`s with an `Instrument` element but without any instrument
+    * configuration information.
+    *
+    * @param oid observation whose steps should be selected
+    */
+  def selectEmpty(oid: Observation.Id): ConnectionIO[List[Step[Instrument]]] =
     sql"""
       SELECT s.instrument,
              s.step_type,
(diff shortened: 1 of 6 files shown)


modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -158,15 +163,88 @@ object StepDao {
...
+             i.filter,
+             i.lyot_wheel,
+             i.disperser,
+             i.window_cover
+        FROM step s
+             LEFT OUTER JOIN step_f2 i
+               ON i.step_f2_id = s.step_id
+       WHERE s.observation_id = $oid
+    ORDER BY s.location
+    """.query[F2Config].list
```
swalker2m: For example, this works but is kind of grim:

````scala
    sql"""
      SELECT i.fpu,
             i.mos_preimaging,
             i.exposure_time,
             i.filter,
             i.lyot_wheel,
             i.disperser,
             i.window_cover
        FROM step s
             LEFT OUTER JOIN step_f2 i
               ON i.step_f2_id = s.step_id
       WHERE s.observation_id = $oid
    ORDER BY s.location
    """.query[(Option[F2FpUnit],
               Option[Boolean],
               Option[Duration],
               Option[F2Filter],
               Option[F2LyotWheel],
               Option[F2Disperser],
               Option[F2WindowCover])].map {
        case (Some(u), Some(m), Some(e), Some(f), Some(l), Some(d), Some(w)) =>
          Some(F2Config(u, m, e, f, l, d, w))
        case _                                                               =>
          None
      }.list.map(_.flatten)
````

**Comment by tpolecat** [comment]
I don't have a good solution to the "where do we specify the instrument" problem. We can come back to it I think. This looks fine for now.

doobie provides some mechanisms for checking queries and once we have the deployment thing working we can have test deployments run benchmarks. So I think we'll be ok here, we're just not there yet.

File: modules/db/src/main/scala/gem/dao/StepDao.scala [hunk 2]
```diff
@@ -240,11 +240,19 @@ object StepDao {
...
-      case _                     => selectAllGeneric(oid)
-    }).map(_.map(_.widen[InstrumentConfig]))
+  def selectAll(oid: Observation.Id): ConnectionIO[List[Step[InstrumentConfig]]] = {
+    def instrumentConfig(ss: List[Step[Instrument]]): ConnectionIO[List[InstrumentConfig]] =
+      ss.headOption.fold(List.empty[InstrumentConfig].point[ConnectionIO]) { s =>
+        (s.instrument match {
+          case Instrument.Flamingos2 => selectF2(oid)
+          case _                     => selectGeneric(oid)
+        }).map(_.map(c => c: InstrumentConfig))  // List isn't a scalaz Functor? So no widen on List?
+      }
```
tpolecat: You should be able to do something like

````scala
  ss.headOption.map(_.instrument).foldMap[List[InstrumentConfig]] {
    case Instrument.Flamingos2 => selectF2(oid)
    case _                     => selectGeneric(oid)
  }
```

tpolecat: Also `List` is absolutely a functor ... `widen` should work.

swalker2m: Hmm, I must be doing something wrong:

````
import scalaz._
import Scalaz._
Welcome to Scala version 2.11.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_45).
Type in expressions to have them evaluated.
Type :help for more information.

scala> trait Animal { def name: String }
defined trait Animal

scala> case class Cat(name: String) extends Animal
defined class Cat

scala> val cats = List(Cat("A"), Cat("B"))
cats: List[Cat] = List(Cat(A), Cat(B))

scala> val animals = cats.widen[Animal]
<console>:17: error: value widen is not a member of List[Cat]
       val animals = cats.widen[Animal]
````

tpolecat: Where are you running that? The REPL is broken in gem at the moment because of `-Xfatal-warnings`. The `.widen` op is definitely in scalaz 7.2.4 which is what we're using.

swalker2m: The snippet comes from a shell command that starts up the REPL with scalaz.   It was using scalaz 7.1.1 but when updated to 7.2.4 `widen` starts working.

Now, as long as I apply it to every branch of the `match` it works:

````scala
    ss.headOption.fold(List.empty[InstrumentConfig].point[ConnectionIO]) { s =>
      s.instrument match {
        case Instrument.Flamingos2 => selectF2(oid).map(_.widen[InstrumentConfig])
        case _                     => selectGeneric(oid).map(_.widen[InstrumentConfig])
      }
    }
````

but if I try to map the result to a widened list the compiler complains:

````
value widen is not a member of _187
}).map(lst => lst.widen[InstrumentConfig])
                  ^
````

Presumably since `List` is covariant there's not even really a need to explicitly widen.  This awkwardness works:

````scala
      ss.headOption.fold(List.empty[InstrumentConfig].point[ConnectionIO]) { s =>
        (s.instrument match {
          case Instrument.Flamingos2 => selectF2(oid)
          case _                     => selectGeneric(oid)
        }).map((lst: List[InstrumentConfig]) => lst)
      }

```` 

Finally, the `foldMap` version also works with a few adjustments but doesn't seem much better, if I'm doing it right:

````scala
      ss.headOption.map(_.instrument).foldMap[ConnectionIO[List[InstrumentConfig]]] {
        case Instrument.Flamingos2 => selectF2(oid).map(_.widen[InstrumentConfig])
        case _                     => selectGeneric(oid).map(_.widen[InstrumentConfig])
      }
````

So after all of that, what's the proper way to do this?

tpolecat: I guess the last way is the way I would do it. You won't need the type argument for `foldMap`. Covariance would make this easier in some ways but harder in others.

PR #34, opened 2016-12-22
F2 Read Mode

swalker2m:
This PR makes a few adjustments, all related to adding a Flamingos2 read mode field to the model.

1. Sorted `F2Config` class and `step_f2` table fields since the order was seemingly random.  Since there are various places where the fields are listed, this makes it easier to line everything up and easier to insert new fields.

2. Switched exposure time database domain definition to milliseconds since F2 supports fractional second exposure times and the bright read mode minimum exposure time is 1.5 seconds.  This was actually a bug in the import in which we were truncating fractional second exposure times.

3. Added read mode.  This required updates to `gen2` to support `java.time.Duration`.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/package.scala
+++ b/modules/db/src/main/scala/gem/dao/package.scala
@@ -53,7 +53,7 @@ package object dao extends MoreTupleOps with ToUserProgramRoleOps {
     Meta[List[Int]].nxmap(Location.unsafeMiddle(_), _.toList)
 
   implicit val DurationMeta: Meta[Duration] =
-    Meta[Long].xmap(Duration.ofSeconds, _.getSeconds)
+    Meta[Long].xmap(Duration.ofMillis, _.toMillis)
 
   def capply2[A, B, T](f: (A, B) => T)(
     implicit ca: Composite[(Option[A], Option[B])]
--- a/project/gen2.scala
+++ b/project/gen2.scala
@@ -2,6 +2,8 @@ import doobie.imports._
 
 import java.io.File
 
+import java.time.Duration
+
 import scalaz._, Scalaz._
 import scalaz.effect._
 import scala.reflect.runtime.universe.TypeTag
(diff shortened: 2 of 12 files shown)


modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -53,7 +53,7 @@ package object dao extends MoreTupleOps with ToUserProgramRoleOps {
     Meta[List[Int]].nxmap(Location.unsafeMiddle(_), _.toList)
 
   implicit val DurationMeta: Meta[Duration] =
-    Meta[Long].xmap(Duration.ofSeconds, _.getSeconds)
+    Meta[Long].xmap(Duration.ofMillis, _.toMillis)
```
cquiroz: I guess this allows to use `Duration` on queries

project/gen2.scala
```diff
@@ -22,21 +24,26 @@ object gen2 {
     ""
   )
 
+  implicit val DurationMeta: Meta[Duration] =
+    Meta[Long].xmap(Duration.ofMillis, _.toMillis)
+
   object ToDeclaration extends Poly1 {
```
cquiroz: Is this for shapeless?

modules/db/src/main/scala/gem/dao/package.scala
swalker2m: Right, you can add [custom mappings](https://tpolecat.github.io/doobie-0.2.1/10-Custom-Mappings.html) for your types as explained in the Book of Doobie.

project/gen2.scala
swalker2m: I wanted a couple of `Duration` fields to be part of the generated class for Flamingos2 `ReadMode`.  Part of the process of generating the class is reading the `e_f2_read_mode` class where the data is stored as an integral type.  Since it needs to convert that into a `Duration` it has to be told how.  This is the same thing that is done over in the `dao` package as you noted above.  It's repeated here because it's not available otherwise.  Maybe it makes sense to take some of these mappings for common things like `Duration` and `Timestamp` and move it to somewhere that it can be shared, if possible.  I'm not sure where exactly though because this is needed for code generation before the rest of the code is built.

PR #35, opened 2016-12-28
Smart Gcal Expansion

swalker2m:
The goal of this work-in-progress PR is to implement a simple smart gcal expansion API like:

````scala
  def lookup(oid: Observation.Id, loc: Location.Middle): ConnectionIO[LookupResult]

  def expand(oid: Observation.Id, loc: Location.Middle): ConnectionIO[LookupResult]
````

where `LookupResult` is a disjunction: error code if it fails or `List[GcalStep[InstrumentConfig]]` if successful.  `lookup` just finds the gcal configuration matching the smart gcal step at the given location if there is one and `expand` actually replaces the smart gcal step with the gcal config.

The PR also changes the `StepDao.selectAll` methods to return Scalaz maps `Location ==>> Step[A]` in order to have access to the location information.  In order to expand the smart gal step, we have to create locations that fall between the previous and next steps.

I'm not quite ready to merge this yet:

- [x] Document, comment
- [x] Complete test cases
- [ ] Decide whether `SmartGcalDao` is really the right place for this higher-level API. If not, where?

Presumably at some point, this can be cleaned up with SQL fragments in a newer version of doobie.  As is, there is a bit of duplication in some of the queries.

### DIFF
--- a/modules/core/src/main/scala/gem/Location.scala
+++ b/modules/core/src/main/scala/gem/Location.scala
@@ -79,8 +79,8 @@ object Location {
       "{ω}"
   }
 
-  def beginning: Location = Beginning
-  def end: Location       = End
+  val beginning: Location = Beginning
+  val end: Location       = End
 
   // Constructors
 
@@ -181,7 +181,7 @@ object Location {
   })
 
   implicit val OrderMiddle: Order[Location.Middle] =
-    Order.orderBy(m => m: Location)
+    OrderLocation.contramap[Location.Middle](lm => lm: Location)
 
   implicit val ShowLocation: Show[Location] = Show.shows(_.toString)
 }
--- a/modules/core/src/main/scala/gem/Location.scala
+++ b/modules/core/src/main/scala/gem/Location.scala
@@ -11,7 +11,7 @@ import scalaz.Ordering.{EQ, GT, LT}
   * `Location` is a proper prefex of another, then it sorts ahead of the other
   * `Location`.
   */
-sealed trait Location {
+sealed trait Location extends Product with Serializable {
 
   // These functions aren't of any use to clients.  Instead they are involved
   // in the cacluation of Locations that fall between two other locations.
@@ -78,6 +78,9 @@ object Location {
     override def toString: String =
       "{ω}"
   }
+
+  def beginning: Location = Beginning
+  def end: Location       = End
 
   // Constructors
 
@@ -176,6 +179,9 @@ object Location {
         .find { case (a, b) => a =/= b }
         .fold[Ordering](EQ) { case (a, b) => (a < b) ?[Ordering] LT | GT }
   })
+
+  implicit val OrderMiddle: Order[Location.Middle] =
+    Order.orderBy(m => m: Location)
 
   implicit val ShowLocation: Show[Location] = Show.shows(_.toString)
 }
--- a/modules/core/src/main/scala/gem/config/GcalConfig.scala
+++ b/modules/core/src/main/scala/gem/config/GcalConfig.scala
@@ -22,31 +22,33 @@ object GcalConfig {
   type GcalArcs = OneAnd[ISet, GcalArc]
   type GcalLamp = GcalContinuum \/ GcalArcs
 
-  def mkLamp(continuum: GcalContinuum): GcalLamp =
-    continuum.left
+  object GcalLamp {
+    def fromConfig(continuum: Option[GcalContinuum], arcs: (GcalArc, Boolean)*): Option[GcalLamp] = {
+      // Extract the arc lamps to include, if any.
+      val as = arcs.filter(_._2).unzip._1.toList
 
-  def mkLamp(arc0: GcalArc, arcs: GcalArc*): GcalLamp =
-    OneAnd(arc0, ISet.fromList(arcs.toList)).right
+      // Extract the continuum lamp, assuming there are no arcs.
+      val co = continuum.flatMap { c => as.isEmpty option c.left[GcalArcs] }
 
-  def mkLamp(continuum: Option[GcalContinuum], arcs: (GcalArc, Boolean)*): Option[GcalLamp] = {
-    // Extract the arc lamps to include, if any.
-    val as = arcs.filter(_._2).unzip._1.toList
+      // Prepare the arc lamps, assuming there is no continuum.
+      val ao = as match {
+        case h :: t if continuum.isEmpty => Some(OneAnd(h, ISet.fromList(t)).right[GcalContinuum])
+        case _                           => None
+      }
 
-    // Extract the continuum lamp, assuming there are no arcs.
-    val co = continuum.flatMap { c => as.isEmpty option c.left[GcalArcs] }
-
-    // Prepare the arc lamps, assuming there is no continuum.
-    val ao = as match {
-      case h :: t if continuum.isEmpty => Some(OneAnd(h, ISet.fromList(t)).right[GcalContinuum])
-      case _                           => None
+      co orElse ao
     }
 
-    co orElse ao
+    def fromContinuum(continuum: GcalContinuum): GcalLamp =
+      continuum.left
+
+    def fromArcs(arc0: GcalArc, arcs: GcalArc*): GcalLamp =
+      OneAnd(arc0, ISet.fromList(arcs.toList)).right
+
+    def unsafeFromConfig(continuum: Option[GcalContinuum], arcs: (GcalArc, Boolean)*): GcalLamp =
+      fromConfig(continuum, arcs: _*).getOrElse {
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: modules/db/src/main/scala/gem/dao/SmartGcalDao.scala [hunk 1]
```diff
@@ -56,4 +60,55 @@ object SmartGcalDao {
...
+    // Find the previous and next location for the smart gcal step that we are
+    // replacing.  This is needed to generate locations for the steps that will
+    // be inserted.
+    def bounds(steps: Location.Middle ==>> Step[InstrumentConfig]): (Location, Location) =
+      steps.split(loc) match {
+        case (prev, next) => (prev.findMax.map(_._1).widen[Location] | Location.beginning,
+                              next.findMin.map(_._1).widen[Location] | Location.end)
+      }
+
+    def insert(before: Location, gcal: ExpandedSteps, after: Location): ConnectionIO[Unit] =
```
cquiroz: I wonder why the result is `ConnectionIO[Unit]` Would you want to e.g. obtain the inserted ids?

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -171,16 +212,10 @@ object StepDao {
              LEFT OUTER JOIN step_smart_gcal ss
                 ON ss.step_smart_gcal_id = s.step_id
        WHERE s.observation_id = $oid
-    ORDER BY s.location
-    """.query[StepKernel].map(_.toStep).list
+    """.query[(Loc, StepKernel)].map(_.map(_.toStep)).list.map(==>>.fromList(_))
```
cquiroz: Is the `fromList` reading a list of tuples?

modules/db/src/test/scala/gem/dao/SmartGcalSpec.scala
```diff
@@ -0,0 +1,85 @@
...
+// - Unknown location
+// - Not smart gcal
+// - No mapping
+// - valid mapping, first step
+// - valid mapping, in between step
+// - mapping arc, flat, night, day
+
+
+
+class SmartGcalSpec extends FlatSpec with Matchers {
```
cquiroz: Nice to have some tests. Should we run them on travis?

modules/db/src/test/scala/gem/dao/StepDaoSample.scala
```diff
@@ -1,18 +1,20 @@
...
+  type Result = Location.Middle ==>> Step[InstrumentConfig]
 
   val oid = Observation.Id.unsafeFromString("GS-2016A-Q-102-108")
 
   override def runl(args: List[String]): ConnectionIO[Result] =
     StepDao.selectAll(oid)
 
   override def format(r: Result): String =
-    r.mkString(", \n")
+    r.toAscList.map { case (l, s) => f"$l%10s -> $s" }.mkString(", \n")
```
cquiroz: `toAscList`?

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -11,7 +11,7 @@ import scalaz.Ordering.{EQ, GT, LT}
   * `Location` is a proper prefex of another, then it sorts ahead of the other
   * `Location`.
   */
-sealed trait Location {
+sealed trait Location extends Product with Serializable {
```
swalker2m: I'm open to suggestions.  It's used to sort steps.  The absolute value is meaningless but you can compare one with another to know which comes first and can always find a value that falls between any two others.  In the Logoot paper it's called a line id.

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: Yes, it takes a list of tuples where the first value is used as the key (and so has to have an `Order` instance in scope).  In this case, the keys are `Location.Middle` so I had to define an `Order` instance for that even though `Location` itself is already ordered.

modules/db/src/test/scala/gem/dao/StepDaoSample.scala
swalker2m: Ascending, which I think is the default if you use `toList` but I figured why not be explicit.

tpolecat:
I'll take a look at this over the weekend, sorry for the delay.

modules/core/src/main/scala/gem/config/GcalConfig.scala
```diff
@@ -22,6 +22,12 @@ object GcalConfig {
   type GcalArcs = OneAnd[ISet, GcalArc]
   type GcalLamp = GcalContinuum \/ GcalArcs
 
+  def mkLamp(continuum: GcalContinuum): GcalLamp =
+    continuum.left
+
+  def mkLamp(arc0: GcalArc, arcs: GcalArc*): GcalLamp =
+    OneAnd(arc0, ISet.fromList(arcs.toList)).right
+
```
tpolecat: I think I would prefer something like this to avoid awkward naming and overloading.

```scala
object GcalLamp {

  // most general constructor
  def apply(continuum: Option[GcalContinuum], arcs: (GcalArc, Boolean)*)

  // others
  def fromContinuum(continuum: GcalContinuum): GcalLamp =...
  def fromArcs(arc0: GcalArc, arcs: GcalArc*): GcalLamp = ...

}
```

There's a point at which we'll want to newtype the `\/`.

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -56,4 +60,85 @@ object SmartGcalDao {
     } yield r
   }
 
+  private def lookup(step: Option[Step[InstrumentConfig]], loc: Location.Middle): ConnectionIO[LookupResult] = {
```
tpolecat: These computations might be easier to express in terms of `EitherT`, which would let you avoid some of the `.left/.right/.fold/.point`ing, but it's fine as-is. Just something to keep in mind.

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -193,65 +228,137 @@ object StepDao {
...
+      s.instrument match {
+        case Instrument.Flamingos2 => oneF2Only(oid, loc)     .map(_.widen[InstrumentConfig])
+        case _                     => oneGenericOnly(oid, loc).map(_.widen[InstrumentConfig])
+      }
+
+    for {
+      so <- selectOneEmpty(oid, loc)
+      io <- so.fold(Option.empty[InstrumentConfig].point[ConnectionIO]) { instrumentConfig }
+    } yield (so |@| io)((s, i) => s.as(i))
+  }
```
tpolecat: Similarly using `OptionT` here could prevent some awkwardness.

swalker2m:
Thanks for the feedback.  It all seemed good to me, so I took a stab at implementing it.  The switch to `OptionT` made things a bit nicer but I may not have managed the `EitherT` correctly.  Also I'm not sure about the syntax added to the `gem.dao` package object.  Does it belong there?  Is there something like this (or likely better) defined already that I'm missing?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -129,16 +139,14 @@ object SmartGcalDao {
...
-                 for {
-                   _ <- StepDao.delete(oid, loc)
-                   _ <- insert(locBefore, gs, locAfter)
-                 } yield ())
+      gcal  <- lookup(MaybeConnectionIO.fromOption(steps.lookup(loc)), loc)
+      _     <- for {
+                 // replaces the smart gcal step with the expanded manual gcal steps
+                 _ <- StepDao.delete(oid, loc).right[ExpansionError]
+                 _ <- insert(locBefore, gcal, locAfter).right[ExpansionError]
+               } yield ()
```
tpolecat: You should be able to flatten this out to

```
gcal  <- lookup(MaybeConnectionIO.fromOption(steps.lookup(loc)), loc)
_     <- StepDao.delete(oid, loc).right[ExpansionError]
_     <- insert(locBefore, gcal, locAfter).right[ExpansionError]
```

modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -14,6 +14,59 @@ import Scalaz._
...
+      EitherT.fromDisjunction(eab)
+  }
+
+  implicit class ConnectionIOOps[T](c: ConnectionIO[T]) {
+    def left[B]: EitherConnectionIO[T, B] =
+      EitherConnectionIO.left[T, B](c)
+
+    def right[A]: EitherConnectionIO[A, T] =
+      EitherConnectionIO.right[A, T](c)
+  }
```
tpolecat: I would rename these because you're shadowing the `.left/.right` syntax provided by scalaz.

PR #36, opened 2017-01-04
Seqexec / Gem Discussion

swalker2m:
This document is a result of a conversation with Carlos and Arturo about the possibility of having the seqexec switch to work directly with the new sequence model.  I'm not 100% sure this is possible but I think we're not too far off and it seemed worth discussing in the context of merging the `ocs3` and `gem` builds.  Please let me know if you have any thoughts on this.

### DIFF
--- a/SeqexecGem.md
+++ b/SeqexecGem.md
@@ -41,26 +41,34 @@ All the above is type safe and straightforward to use.
 
 * Avoids eventually having to circle back and rewrite parts of the seqexec to migrate to the new model.
 * Exercises the new sequence model well before it would otherwise be available for production, partially addressing the governing board concerns.
-* Exposes the new program model and database to the rest of the high-level team.
+* Exposes the new program model and database to the rest of the high-level team and gets everybody working on the same codebase.
 * Necessitates combining the current `ocs3` and `gem` builds and deployment models, which keeps them from diverging.
 * Concentrates all OCS2 to OCS3 translation code in one place, instead of having both seqexec and "gem" projects translating and depending on OCS2.
 * Allows us to flesh out dataset record, observing events, and logs in the new program model and postgres database.
+* Prioritizes future development, focusing on seqexec needs like full instrument support.
 
 
-## Sketching Out a Course
+## Sketching Out a Proposal
 
 The technical issue preventing the new seqexec from working with the new sequence model today is that it must continue to read sequences from the
 existing Observing Database.  The new science program model is not complete and there is no UI for editing it.  A potential solution is to offer
-an enqueue service that:
+an enqueue/translator service that:
 
-* Contacts the ODB and requests the OCS2 `ConfigSequence` over trpc. 
+* Contacts the ODB and requests the OCS2 `ConfigSequence` over trpc, or perhaps asks for the XML representation if that helps avoid dependency
+issues.
 * Translates the config sequence into a minimal science program with just enough information to contain the sequence.
 * Writes the sequence to postgres database tables, replacing any existing sequence information for the observation.
 
 The seqexec could then work directly with the new sequence in the postgres database.  The enqueue service could be developed separately from the
 rest of the combined seqexec/gem project because it will be dependent on OCS2 versions of pervasive libraries like `scalaz`.  The rest of the new
 codebase is then freed of its OCS2 shackles.  (Though of course, the seqexec will still need to use the XMLPRC OCS2 `wdba` services to record
-execution events in the Observing Database.)
+execution events in the Observing Database.)  This same translation project would eventually be responsible for migration from the OCS2 science
+program as well.
+
+Eventually as observations are edited any sequence information traslated and added to the postgres database will of course be out-of-date with
+the ODB.  This is why enqueueing observations or refetching from the ODB must always replace any existing sequence information.  Eventually when
+everything is running on OCS3 we can simply delete the enqueue service since the seqexec will then always have the program to work with in the
+postgres database.
 
 As Carlos suggested, pressing the "Queue" button in today's OT could be made to trigger the enqueue service to perform these steps.
 
--- a/SeqexecGem.md
+++ b/SeqexecGem.md
@@ -1,0 +1,71 @@
+## Seqexec / Gem Proposal
+
+The new seqexec currently under development works with an OCS2 `ConfigSequence` which it obtains from the existing Observing Database over 
+our custom RPC protocol.  The `ConfigSequence` is essentially a `List[Config]` where each `Config` is roughly a `Map[String, Object]`.  The
+`Object`s in the sequence vary from primitive values to specific enumerations like `Flamingos2.Filter`.  Client code has to know the expected 
+type of the `Object` associated with each key in order to work with it.
+
+At the same time, a new sequence model is being developed as part of the OCS3 effort.  The new sequence model avoids the indignity and risk of
+having to downcast from `Object` to the "known" type for a key.  Eventually the new seqexec will abandon the OCS2 sequence in favor of the new
+model.  This brief document explores the possibility of developing the new seqexec directly on the OCS3 sequence, bypassing the intermediate step
+of targeting OCS2 first.
+
+
+## Current Status
+
+The new seqexec is slated for production well before the OCS3 code will be deployed in full. Nevertheless the sequence model itself along with
+Flamingos 2 and GCal configurations are fairly well baked.  I believe that with comparatively little effort enough of the sequence would be
+available to comply with the needs of the seqexec.
+
+### What We Have
+
+* Sequence as a `List[Step]`.
+* `Step` as an ADT with bias, dark, gcal, science, and smart gcal options.
+* Telescope offset configuration for science steps.
+* Complete gcal configuration for gcal steps.
+* Smart gcal expansion for Flamingos 2.
+* Instrument configuration for Flamingos 2.
+* Data-access code for reading/writing the sequence to a postgres database.
+
+All the above is type safe and straightforward to use.
+
+### What We're Missing
+
+* All other instrument configurations.
+* "Guide With" values.
+* Ready/Complete status.
+* others?
+
+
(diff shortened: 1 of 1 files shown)


cquiroz:
I'm totally on board 👍

tpolecat:
I'm a little nervous about writing the program to the database because it makes seqexec dependent on a lot more than just the model. Why not just hand it to the seqexec directly?

arturog8m:
I kind of like the idea of handing the sequence directly to the seqexec. In addition of missing the opportunity to use the database early in an operational environment, what other disadvantages this might have?

cquiroz:
I don't know what you mean handling the sequence directly. The seqexec needs to search for sequences (either by id, or sequences scheduled to be run). Sounds more like a pull model than a push.

We could have an in memory cache with all the sequences to be ran though that's very similar to just having the db

A general drawback is having to depend on the db though

swalker2m:
> I'm a little nervous about writing the program to the database because it makes seqexec dependent on a lot more than just the model. Why not just hand it to the seqexec directly?

> I kind of like the idea of handing the sequence directly to the seqexec. In addition of missing the opportunity to use the database early in an operational environment, what other disadvantages this might have?

I guess I saw using more of the machinery and API as soon as possible as a goal instead of a bug.  It would be developing a seqexec that will work more like we envision it working in the end.  We could also phase it in though when everyone is more comfortable with the state of things.  It's still a win to start exercising at least part of the codebase.

As noted one part of the sequence that is missing from the new sequence model is the completed state information.  I've been assuming that will work by checking whether there is a matching dataset in a table of dataset records.  The seqexec would be responsible for updating that table, presumably, as datasets are written.  It's also where it will store exec events and presumably its logs.  Do you think we need to avoid the database altogether or could it still be used for these kinds of things?  I'm not sure which part we are worried about in particular.

arturog8m:
What I mean is that the 'enque' service will pass the sequence directly to the seqexec (instead of writing it to the postgres database). We will need to figure out at what time, probably when the sequence is 'queued'. The seqexec will need to store this information (probably in memory) - but it won't need to have all the science program, just the sequence. I guess once the new OCS is in place, we might want to have the seqexec reading the sequences directly from the postgres database, so this code will be replaced, but the actual manipulation of the sequences by the seqexec should be the same, as it will be the same model.

swalker2m:
The client API for extracting a sequence from the database is a one liner, but you're right that line could change a bit over time.  The other nice thing about writing the sequence to the database is that it enables the OCS2 translator to be independent from the seqexec.  It writes an updated sequence to the database and the seqexec reads it from there using the client API.  I guess we could concentrate on a JSON representation or something though to keep it out of the seqexec.

> Somehow that information will need to be sent to the current OCS.

Absolutely.  Unfortunately it has to continue to use the WDBA as well until the current ODB is retired.

cquiroz:
I think this is a good result for the general ocs3 project. Since we need to show concrete deliverables, having the new model and db access used by the seqexec is a clear milestone in my view

tpolecat:
We have the notion of a unique step ID in the new system and were going to use this to associate data sets with sequence steps and do away with consecutive numbering. How do we reconcile this with OCS2 which expects steps to be numbered consecutively?

tpolecat:
I just have a general fear that this is going to end up being much harder than we think, because everything with OCS always is, and every time you have duplicated information it ends up causing trouble. So I think it's worth investigating this in much more detail before committing to it.

swalker2m:
> So I think it's worth investigating this in much more detail before committing to it.

Agreed, that's what this PR is meant to kick off if some basic level of agreement can be reached that this is an idea worth pursuing. 

> We have the notion of a unique step ID in the new system and were going to use this to associate data sets with sequence steps and do away with consecutive numbering. How do we reconcile this with OCS2 which expects steps to be numbered consecutively?

This is not reconcilable.  Regardless of what we do, as long as we have to work with the ODB and the OCS2 sequence model at some level, only 1:1 consecutive numbering is supported.  The advantage of consulting a dataset table in the seqexec is that this is the way it will eventually work presumably.  Otherwise, we'll need to invent something else to pass along with the sequence that informs the seqexec about completion state.  For example, I guess it could get a `List[Step[(InstrumentConfig, Boolean)]]` where the `Boolean` flags whether the step is completed.

The point about duplicated information is certainly true.  It adds a bit of fragility until we can get away from OCS2 altogether.  The further away we get from how things will eventually work, the less advantage there is to doing this though.  I suppose to avoid duplication the translation service could update the dataset table itself when a sequence is read from OCS2 to make it consistent with what it reads from the sequence.  Then it is duplicated but made to be consistent every time a sequence is fetched.

Then there's just the question of whether to write the sequence itself to the database or return, say, a JSON representation of the sequence to the seqexec so it can instantiate the sequence from that?  That would be work that eventually would no longer be needed when we're more sure about the model at some point.  Perhaps that's for the best but I'm not sure I understand yet the risk of reading the sequence from the database.

tpolecat:
> This is not reconcilable. Regardless of what we do, as long as we have to work with the ODB and the OCS2 sequence model at some level, only 1:1 consecutive numbering is supported. 

I think I don't understand how we deal with this. The Gem model has `step_id` which is an ever-growing integer value that's distinct from the `location` tht determines the desired order of execution. My understanding of our plan was that the unique id would be written in the FITS header and would be associated with events from the Seqexec, so that each step essentially has an independent execution log and we can run them in whatever order we want without risk of screwing things up.

So, if we rewrite the sequence every time we re-import the sequence from OCS2 we get a new set of ids and invalidate any existing event log.

We could add an `ocs2_index` column and only replace the unexecuted portion of the sequence but there's nothing guaranteeing that this will be consistent since you OCS2 lets you change the past. I guess we have been saying "don't do that" for long enough that it probably won't happen but it still bothers me.

I think I'm probably missing something.

swalker2m:
Let me try to explain the idea with completion state in a bit less of a stream of consciousness style.  As long as we are using the ODB as the ultimate source of truth for the seqexec, we are limited by the fact that sequence step ordering for completed steps cannot change.  Without changing the OCS2 model, I can’t think of any way to fix that.

At the same time, we need completion information in the new seqexec and it won’t work by this 1:1 association with step order.  Instead, one idea is to consult a table of executed datasets where each entry is tied to the unique sequence step id.  If there is an entry in the table, the step has been executed.  Or perhaps we could examine sequence events looking for dataset start/end events if that’s preferable for some reason.

It seems like it would be possible to develop the new seqexec to work by checking a dataset table for completion information right now, even though we haven’t abandoned the ODB.  I’m proposing a separate service that handles the translation of OCS2 information from the ODB to the corresponding tables.  The seqexec then works with those tables just as it would if the postgres database were the ultimate source of truth.  So when a sequence is requested from the ODB the translation service reads from the ODB and writes the data to tables replacing whatever is there.  That keeps it consistent with what is in the ODB. In the case of the dataset table, that means taking the newly created unique step id along with completion information in the OCS2 sequence and writing out datasets.  As the seqexec works, it also records datasets in the table because that’s what it should be doing eventually.  The next time it needs the sequence though, whatever it wrote will be replaced with information from the ODB.  For now, the seqexec has to continue sending sequence events to the ODB since it really is the ultimate source of truth.

There are a couple of goals for all of this:

1. Combine the seqexec and gem efforts so that we’re all working on one new system instead of developing two distinct systems.

2. Produce a seqexec that works as much as possible like we envision it working in the end so that we’re not setting ourselves up for a big merge/refactoring project in the future.

What I’m hearing from you and Art though is that the seqexec schedule and the current state of the gem project aren’t compatible with those goals yet?

cquiroz:
Isn't that when we import the steps would be ordered sequentially on the DB as it is now?
Since we don't edit or reorder the steps on the seqexec, could we just use that ordering?

swalker2m:
> Isn't that when we import the steps would be ordered sequentially on the DB as it is now?
Since we don't edit or reorder the steps on the seqexec, could we just use that ordering?

That's correct.  The issue is about determining which steps have been executed.  In the ODB program model, if you have dataset GS-2017A-Q-1-2-003 that means that the 3rd step is executed.  You cannot insert a step before step 3 because we'd no longer know that the current step 3 is done and instead assume that the new step you inserted is complete.   So it's fragile.  We only support adding new steps at the end which is complicated with iterators -- think about increasing the loop count in a top-level iterator for an executed sequence.  We want to fix this by identifying each step with a unique id that doesn't depend upon where it currently appears in the sequence.

cquiroz:
@swalker2m What you propose sounds workable to me. We are going to have some impedance between the old and the new regardless.

Correct me if I'm wrong, but the issue is if somebody edits he sequence once it has been partially executed? If so we'd need to re import, potentially producing new step locations?

Then we'd need, upon import, to check the current set of executed steps and remap them to the new sequence? The only editor would be the OT so you couldn't insert steps before only append new ones, right?

swalker2m:
> Correct me if I'm wrong, but the issue is if somebody edits he sequence once it has been partially executed? If so we'd need to re import, potentially producing new step locations?

Right, if you edit the sequence with the OT you can change the ordering of the steps.  It even lets you insert new steps before already executed steps which is a problem we have with the existing model.

> Then we'd need, upon import, to check the current set of executed steps and remap them to the new sequence? 

After an edit, you have to import from the ODB again.  I propose that this replaces any step information for the observation, datasets etc. reading the information from the ODB version.  That way the seqexec works with data that is consistent with the ODB.  Once the ODB is turned off, we stop importing and reimporting and less needs to change in how the seqexec works, in theory anyway.

tpolecat:
>  I propose that this replaces any step information for the observation, datasets etc. reading the information from the ODB version.

Ok, so we also replace the log entries when we import. I guess that's ok, as long as the ODB is storing enough information to fill out the log in the way we want to represent it in OCS3. 

If the Seqexec is writing log entries to the database *and* sending them to OCS there is a chance that they will get out of sync but I guess we're no worse off than we are now.

So, **I think I'm ok with this** as long as we're very clear that we are never going to do any round-trip syncing of data between OCS2 and OCS3.

PR #37, opened 2017-01-06
Double to numeric?

swalker2m:
Thinking about bringing targets over to the new model, I find myself wondering what to do about floating point values.  For example we store offsets as `double precision` values in arcsecs:

````
gem=# select * from step_science;
 step_science_id |      offset_p       |      offset_q       
-----------------+---------------------+---------------------
             281 |                   0 |                   0
             282 |    10.0000000000591 |                   0
             283 |                   0 |                   0
             284 |                   0 |                   0
             285 |                   0 |                   0
             286 |                   0 |    2.99999999999727
...
````

A query for offsets in p `<= 10` clearly misses step 282.  I’m wondering if we shouldn’t store these as `numeric` and even model them as `BigDecimal`.  It comes up in the context of targets because basically everything is floating point and it would be nice to get it (closer to) right.

````
 step_science_id | offset_p | offset_q 
-----------------+----------+----------
               1 |    0.000 |    0.000
               2 |   10.000 |    0.000
               3 |    0.000 |    0.000
               4 |    0.000 |    0.000
               5 |    0.000 |    0.000
               6 |    0.000 |    3.000
...
````

Picking the size is bothersome, I suppose.  Here I went with (9,3) thinking that in a signed angle in arcsecs the largest value possible is 648,000 which is 6 digits.  I don’t know if milliarsec is fine grained enough though.  I suppose the next step would be to update the core `Angle` to store `BigDecimal` values instead of `Double` or do we just round when we read/write the values as in this PR?

### DIFF
--- a/modules/sql/src/main/resources/db/migration/V001__Initial_Setup.sql
+++ b/modules/sql/src/main/resources/db/migration/V001__Initial_Setup.sql
@@ -705,9 +705,9 @@ ALTER TABLE step_smart_gcal OWNER TO postgres;
 --
 
 CREATE TABLE step_science (
-    step_science_id integer          PRIMARY KEY REFERENCES step ON DELETE CASCADE,
-    offset_p        double precision NOT NULL,
-    offset_q        double precision NOT NULL
+    step_science_id integer      PRIMARY KEY REFERENCES step ON DELETE CASCADE,
+    offset_p        numeric(9,3) NOT NULL,
+    offset_q        numeric(9,3) NOT NULL
 );
 
 

### CODE REVIEW

**Comment by tpolecat** [comment]
Yes, I think fixed-point is probably what we want everywhere. Definitely in the database, as you suggest. In the model it would be nice if we could say

```scala
case class OffsetP(mas: Fixed[9,3])
```

and I actually think it's possible with the new singleton support in the Typelevel compiler. I'll play around with it over the weekend.

cquiroz:
I don't have much knowledge about it, but could spire help here?

tpolecat:
@cquiroz Spire has a fixed-point type that takes an implicit scale, which isn't quite what we want. But it's close. I'm looking at it right now.

tpolecat:
So, I played around with singleton types and Spire's `FixedPoint` and it's certainly doable. I'm going to look at it a little more this morning and maybe we can decide whether it's worth trying.

cquiroz:
@tpolecat Is there a problem using spire and scala.js? We want to reuse the code on the frontend

tpolecat:
It doesn't work on scala-js, I assume because numerics are totally different on JS. But we can pull out that one class. Messing with it now.

tpolecat:
Ok here's a prototype that wraps `java.math.BigDecimal` which is available in Scala-JS.

https://gist.github.com/tpolecat/e30aec4272c247b0b5d899c74d675fbd

cquiroz:
Looks very interesting. What is ValueOf? I guess we should try to compile this to scala.js to make sure it works

cquiroz:
BigDecimal is supposed to be very slow but probably doesn't matter for us, WDYT?

tpolecat:
> What is ValueOf?

`ValueOf` is a typeclass that provides a value-level witness of a literal type, so you can get your hands on a value that otherwise exists only at compile-time. It's kind of like a `ClassTag` in that you can ask for it and the compiler will forge it for you if the type is known.

> BigDecimal is supposed to be very slow

I was thinking we would convert to double for nontrivial calculations and then convert back.

tpolecat:
Yeah if you could give it a try in scala-js that would be great.

cquiroz:
I'm trying to compile it but I get:

```
[error] /Users/cquiroz/Projects/ocs3/modules/edu.gemini.seqexec.model/src/main/scala/Fixed.scala:19: not found: type ValueOf
[error]   def rescale[NN <: Int : ValueOf]: Fixed[NN] =
```

tpolecat:
@cquiroz you changed the `ScalaOrganization` in your build?

cquiroz:
I did change it, How can I tell that's TL compiler is being used?

tpolecat:
Hm I'm not sure. Are you setting it `in ThisBuild`? You need to do that if you have subprojects.

cquiroz:
Yeah, that did it, it Compiles fine let me see if I can call it from .js

cquiroz:
So it compiles and it can be called in js tests, but I'm seeing a huge increase in compilation time and runtime execution. I'll try to isolate if it is about `Fixed` or just using the typelevel compiler

tpolecat:
Huh, weird.

tpolecat:
Is the slowdown reproducible? Could it have been a 1-time thing due dealing with the new compiler and runtime?

cquiroz:
Will check it again tomorrow morning, it is very hot today and it makes the
laptop work slowly

cquiroz:
So it seems the slowdown was a one-time thing. In summary, today I can compile Fixed in scala.js and execute in the form of tests with very little difference in time

Perhaps as part of the port to sbt 0.13.13 I should also switch to the typelevel compiler

swalker2m:
I think this looks great.  What are the implications of switching to the type level compiler though?  There's no guarantee that support for this will ever be officially part of Scala, correct?

cquiroz:
@tpolecat knows a bit better but from what I understand the implications are fairly low impact as the binary format must be compatible with Lightbend's compiler output.
There is no guarantee that the changes will be integrated but apparently so far all of them have been accepted

tpolecat:
Everything in the Typelevel compiler must first be submitted to Lightbend as a PR and it must have a reasonable likelihood of being merged. So far this has happened 100% of the time, and this change is scheduled for 2.12.2 so I think it's low risk.

As a bonus we also get `-Ypartial-unification` which makes the unapply trick unnecessary (for instance you can just use `.sequence` instead of `.sequenceU`).

tpolecat:
I messed around with also encoding the precision, so `Fixed[9,3]` which would map to `NUMERIC(9,3)` and overflow for numbers ≥ 10^6 but I'm not sure it's worthwhile … we can get close to the same behavior with regard to overflows and stuff but it would take a lot of work to match them up exactly. I think the fixed fractional part is really what we care about.

tpolecat:
In any case I propose merging this as-is and opening a GH issue to implement `Fixed` (would `Decimal` be better?) as another PR. It will need a lot of tests.

tpolecat:
Anyway I think this change is fine as-is. Implementing `Fixed` and figuring out where to use it (do we want `Angle[N <: Int]` as well) will take some thought and care.

swalker2m:
Agreed, that all sounds great.  I'll merge this tiny change as suggested since switching to `Fixed` will be a bigger job.

PR #39, opened 2017-01-12
Sequence Import (WIP)

swalker2m:
This PR provides a first pass at importing sequences directly from the ODB using the WDBA.  There are a few potentially controversial elements though so I thought I'd offer it for feedback now.  It provides an HTTP service that can be used to show or import a sequence, for example:

    http://localhost:8989/seq/import/GS-2016B-Q-33-222

to import the sequence for GS-2016B-Q-33-222 and

    http://localhost:8989/seq/show/GS-2016B-Q-33-222

to get a simple `String` dump of the sequence.  I could provide a telnet command instead for this, and that may be a good idea anyway.  The idea of publishing a service is that the OT, for example, could call this when the Queue button is pressed.

Since we work with the OCS2 sequence XML, we are limited to the items that you can read from the sequence.  That means that the datasets that it writes out have fake filenames and timestamps for example.  More on this below.

The service creates a stubbed out `Program` and `Observation` because they are required in order to satisfy database constraints.  If there is an existing program or observation in the database with the same id, nothing is done.  See the `ProgramDao` and `ObservationDao` where `ON CONFLiCT DO NOTHING` was added.  This was expedient but probably not what we want in the end?  I'm guessing a better idea is to do a query first to check for the existence of a program or observation. 

Finally it adds a completely new `ConfigReader` that is pretty much like the original `ConfigReader` except that it works with the `String`s in the OCS2 sequence XML instead of using OCS2 enums and objects.  The reason for this is that it completely avoids any dependencies on OCS2 classes and the libraries they use.  My proposal is to:

* Create a new OCS2 service that writes program XML (I suppose) for programs, groups, observations, etc. but replaces the iterator hierarchy with the sequence XML -- a hybrid PIO/sequence XML of some kind.  In other words, produce program XML that is closer to what we're representing in the new model with the unrolled sequence.  

* Update the importer to use the new OCS2 service to directly load programs from an ODB instead of going through WDBA.  That will allow us to fill out program and observation, datasets, etc. since we aren't limited to what is in the sequence XML.  We can also eventually use this to do a daily dump of the ODB into a read-only database for reporting, odb browser replacement, etc.

* Remove the existing importer that works with XML files since it requires OCS2 classes and all their dependencies.


modules/seqimporter/src/main/scala/gem/seqimporter/ConfigReader.scala
```diff
@@ -0,0 +1,279 @@
...
+      val P = Key[OffsetP]("p")(Read.arcsec andThen OffsetP.apply)
+      val Q = Key[OffsetQ]("q")(Read.arcsec andThen OffsetQ.apply)
+    }
+
+    case object Observe extends System("observe") {
+      val ObserveType  = Key[String  ]("observeType" )(identity    )
+      val ExposureTime = Key[Duration]("exposureTime")(Read.durSecs)
+      val Status       = Key[Boolean ]("status"      ) {
+        case "complete" => true
+        case _          => false
```
swalker2m: It's a bit lazy to use a `Boolean` for this.  I'll either replace it with an actual type or else, if we write a more complete ODB importer as I propose, do away with this in favor of writing out dataset information directly from the obslog.

modules/core/src/main/scala/gem/config/InstrumentConfig.scala
```diff
@@ -13,6 +13,15 @@ sealed abstract class InstrumentConfig extends Product with Serializable {
       case f2: F2Config     => Some(f2.key)
       case GenericConfig(_) => None
     }
+
+  def instrument: Instrument = {
+    import gem.enum.Instrument._
+
+    this match {
```
cquiroz: Does this mean that we get the F2 config if we have read an Instrument form the db?

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -21,6 +21,7 @@ object ObservationDao {
                     ${o.id.index},
                     ${o.title},
                     ${o.instrument})
+       ON CONFLICT DO NOTHING
```
cquiroz: I guess this is debatable. If you want to have the latest snapshot from ocs2 you probably want to delete what you have and overwrite

But yes it can be dangerous

modules/seqimporter/src/main/scala/gem/seqimporter/ConfigReader.scala
```diff
@@ -0,0 +1,279 @@
...
+        sys.error(s"Unexpected observeType: $x${m.toList.mkString("\n>  ", "\n  ", "")}")
+    }
+
+    step.strengthR(m.uget(Legacy.Observe.Status))
+  }
+
+  def unsafeInstConfig(i: Instrument, m: ConfigMap): InstrumentConfig =
+    i match {
+      case Instrument.Flamingos2 =>
+        val disperser     = m.uget(Legacy.Instrument.F2.Disperser)
```
cquiroz: Why do we need to use the unsafe get so much?

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
```diff
@@ -0,0 +1,152 @@
...
+  private val client = PooledHttp1Client()
+
+  /** Requests a sequence from the WDBA using XML-RPC, parses the result into
+    * sequence XML.
+    */
+  private def fetchSequenceXml(oid: Observation.Id): Task[Elem] = {
+    // WDBA implements an XML-RPC service for fetching sequences.  We post the
+    // following XML and it returns the matching sequence.
+    val req: Task[Request] =
+      Request(Method.POST, Uri.unsafeFromString("http://gsodb.gemini.edu:8442/wdba")).withBody {
```
cquiroz: nice 👍

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
```diff
@@ -0,0 +1,152 @@
...
+  def fetchSequence(oid: Observation.Id): Task[List[Step[(InstrumentConfig, Boolean)]]] =
+    fetchSequenceXml(oid).map { parseSequenceXml(oid, _) }
+
+  private val configureLogging: Task[Unit] =
+    Task.delay(List(
+      "edu.gemini.spModel.type.SpTypeUtil"
+    ).map(Logger.getLogger).foreach(_.setLevel(Level.OFF)))
+
+  // We need a minimal program/observation in order to insert steps.  The
+  // seqexec doesn't actually use these but they have to be there to satisfy
```
cquiroz: Do you need to read that back?

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
```diff
@@ -0,0 +1,152 @@
...
+      _ <- ObservationDao.insert(o)
+    } yield ()).transact(lxa)
+  }
+
+  // Writes the sequence to the database.
+  private def writeSequence(oid: Observation.Id, s: List[Step[(InstrumentConfig, Boolean)]]): Task[Unit] = {
+    // Create datasets for the complete steps. Fake filename and timestamp.
+    // Seqexec doesn't need to read these values anyway.
+    val datasets = s.map(_.instrument._2).zipWithIndex.map { case (complete, i) =>
+      complete option Dataset(Dataset.Label(oid, i+1), s"unused", Instant.now)
```
cquiroz: I don't get why do you need the fake datalabels

modules/core/src/main/scala/gem/config/InstrumentConfig.scala
swalker2m: It's just a way to go from the specific `InstrumentConfig` back to the `Instrument` enum.  Right now we only have two `InstrumentConfig`s.  `F2Config` for Flamingos2 and `GenericConfig` for everything else.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
swalker2m: At the moment, I don't have any program or observation information other than the id.  Maybe what we want is to do a query and check whether the program and observation already exist before trying to create stubbed out versions.

modules/seqimporter/src/main/scala/gem/seqimporter/ConfigReader.scala
swalker2m: It's all unsafe since it is converting `String`s.  For example if a `Filter` value shows up in the sequence XML and we haven't added an entry for it in the corresponding map, it will blow up.  When the final ODB import is done someday we won't need any of this anymore.

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
swalker2m: I don't think the seqexec will need to read the program or observation, though if the information were correct it could certainly provide more information like the observation title.  These need to exist in the database though because there are constraints such as the one that says each step's `observation_id` references the `observation_id` column of the `observation` table.

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
swalker2m: The labels are real, just as they are in the ODB.  It's the other information that I don't have access to looking just at the sequence XML.  To get the actual filename and timestamp I'd need to have access to the observation's obslog node.  I think we should create a new OCS2 service that provides the full program/observation configuration along with its unrolled sequence.  At that point, I can populate the `dataset` table with all real information.  Even now though, the seqexec could read this table to determine whether a step is complete.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
tpolecat: Yeah I don't think this is what we want. This should be an error. You can delete from the program table and it will cascade everything.

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -362,4 +362,34 @@ object StepDao {
...
+    // configuration if we're deleting a gcal step.
+
+    val delGcal =
+      sql"""
+        DELETE FROM gcal g
+              USING step_gcal sg, step s
+              WHERE s.observation_id = $oid
+                AND s.step_id        = sg.step_gcal_id
+                AND sg.gcal_id       = g.gcal_id
+      """.update.run
```
tpolecat: Can we not set this up to cascade as well?

tpolecat:
> Create a new OCS2 service that writes program XML (I suppose) for programs, groups, observations, etc. but replaces the iterator hierarchy with the sequence XML …

This seems good. The existing importer needs the OCS2 model mostly for this reason. There will be a lot of duplication of key constants and stuff but the clean separation seems worth it.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
swalker2m: I don't think we want to limit the seqexec to a single observation per program, which is what we'd end up with if we delete the entire program every time we fetch an observation.  I basically want to leave the program and observation stub alone if they exist, but create them otherwise.  If a query to check and then create if necessary is done in a single transaction, that should do it?

tpolecat: I think I would just do the insert and catch the duplicate key exception. Something like

```scala
import doobie.contrib.postgresql.syntax._

doInsert(...).onUniqueViolation(0.point[ConnectionIO]) // ConnectionIO[Int]
```

modules/db/src/main/scala/gem/dao/StepDao.scala
swalker2m: I certainly may have set it up incorrectly.  There is a `gcal` table which has gcal configurations with unique ids. Each gcal configuration corresponds to either a step in `step_gcal` or else a smart gcal mapping like those in `smart_f2`.  The reason I made a `gcal` table was to avoid duplicating all the gcal config columns in both `step_gcal` and every `smart_*` table like `smart_f2`.

So the `gcal` entries don't reference steps or smart gcal mappings but steps and smart gcal mappings reference them.  To make it cascade I'd need to add an optional step id reference in the `gcal` table?

modules/db/src/main/scala/gem/dao/ObservationDao.scala
swalker2m: That's a great idea.  I didn't know about that.  If I do it this way though:

````scala
    (for {
      _ <- ProgramDao.insert(p).onUniqueViolation(0.point[ConnectionIO])
      _ <- ObservationDao.insert(o).onUniqueViolation(0.point[ConnectionIO])
    } yield ()).transact(lxa)
````

I get an exception: 

    org.postgresql.util.PSQLException: ERROR: current transaction is aborted, commands ignored until end of transaction block

If I `transact` each individual insert though, it works:

````scala
    for {
      _ <- ProgramDao.insert(p).onUniqueViolation(0.point[ConnectionIO]).transact(lxa)
      _ <- ObservationDao.insert(o).onUniqueViolation(0.point[ConnectionIO]).transact(lxa)
    } yield ()
````

This was surprising to me for some reason.  Am I doing it wrong?

Comment (tpolecat) [inline modules/db/src/main/scala/gem/dao/ObservationDao.scala]: Oh ok if you need to keep going then you have to roll back the transaction at that point.

```scala
blah.onUniqueViolation(HC.rollback as 0)
```

If you have done stuff prior to that point you can use a Savepoint to only roll back to the point before you attempt `blah`.

```scala
HC.setSavepoint.flatMap(s => blah.onUniqueViolation(HC.rollback(s) as 0))
```

tpolecat: The reason I'm avoiding "check first and then insert" is that it's a race condition unless you have your transaction isolation set to `SERIALIZABLE` which is very expensive. We can however turn it on for just this one interaction if it ends up being hard to manage pessimistically.

modules/seqimporter/src/main/scala/gem/seqimporter/SequenceImporter.scala
```diff
@@ -0,0 +1,154 @@
...
+  private def createMinimalContext(oid: Observation.Id, s: List[Step[(InstrumentConfig, Boolean)]]): Task[Unit] = {
+    val i = s.headOption.map(_.instrument._1.instrument)
+    val p = Program[Nothing](oid.pid, "", Nil)
+    val o = Observation[Nothing](oid, "", i, Nil)
+
+    for {
+      _ <- ProgramDao.insert(p).onUniqueViolation(0.point[ConnectionIO]).transact(lxa)
+      _ <- ObservationDao.insert(o).onUniqueViolation(0.point[ConnectionIO]).transact(lxa)
+    } yield ()
+  }
```
tpolecat: Hi, sorry to keep picking at this but I would prefer to see it as a single transaction so that if `ObservationDao.insert` fails for a reason other than unique violation the `ProgramDao.insert` will get rolled back as well. 

We can define a combinator that rolls back a single operation on unique violation

```scala
def ignoreUniqueViolation(fa: ConnectionIO[Int]): ConnectionIO[Int] =
  for {
    s <- HC.setSavepoint
    n <- fa.onUniqueViolation(HC.rollback(s).as(0))
    _ <- HC.releaseSavepoint(s)
  } yield n
```

which lets us write the insert as

```scala
val prog = 
  ignoreUniqueViolation(ProgramDao.insert(p)) *> 
  ignoreUniqueViolation(ObservationDao.insert(o))
  
prog.transact(xa)
```

swalker2m: Thanks, that looks good to me and I appreciate the help.  I'll make the corresponding changes but I'm going to cancel this PR and rework it now that I have an "OCS3" XML format.  Instead of going through the WDBA and being limited to what is in the sequence, I plan for it to be able to import the entire program or observation and to replace the existing import code that works with OCS2 libraries and their dependencies.

PR #40, opened 2017-02-05
Dependency Free OCS2 Importers

swalker2m:
This PR adds an `ocs2` module that handles all migration duties from the existing OCS2 model.  The idea is to replace the `importer` with this version because it requires no OCS2 dependencies.  Three applications are included:

1. `FileImporter` - imports from OCS2 XML files generated by the `exportOcs3` OSGi shell command or from the `Ocs3ExportServlet` at `http://`_odbhost_`:8442/ocs3/fetch/`_programId_

2. `ImportServer` - an http service that accepts requests to import directly from an OCS2 ODB (via the `Ocs3ExportServlet` that runs in the ODB).  This is intended to be used, for example, by the OT's "Queue" button and/or the new seqexec.  It can also be used from a script to perform a full update import from an ODB without requiring an export to files.

3. `SmartGcalImporter` - imports smart gcal CSV files into smart gcal tables.


tpolecat:
Wow, this is impressive. I need to step out for a little while right now but will go through this before lunch.

modules/ocs2/src/main/scala/gem/ocs2/pio/PioError.scala
```diff
@@ -0,0 +1,15 @@
...
+object PioError {
+  final case class MissingKey(name: String)                    extends PioError
+  final case class NullValue(name: String)                     extends PioError
+  final case class ParseError(value: String, dataType: String) extends PioError
+  final case class GeneralError(dataType: String)              extends PioError
+
+  def missingKey(name: String): PioError                    = MissingKey(name)
+  def nullValue(name: String): PioError                     = NullValue(name)
+  def parseError(value: String, dataType: String): PioError = ParseError(value, dataType)
+  def generalError(dataType: String)                        = GeneralError(dataType)
```
tpolecat: Did you forget `: PioError ` here?

modules/ocs2/src/main/scala/gem/ocs2/pio/PioDecoder.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.xml.Node
+import scalaz.Scalaz._
+import scalaz._
+
+/** Typeclass for decoders of PIO-ish XML exported from an OCS2 database.
+  */
+trait PioDecoder[A] { outer =>
+  def decode(n: Node): PioError \/ A
+}
```
tpolecat: `outer` isn't used here. This is isomorphic to `Kleisli[PioError \/ ?, Node, A]` which is a monad and a bunch of other things, so we might want to add at least `Functor` here.

modules/ocs2/src/main/scala/gem/ocs2/pio/PioParse.scala
```diff
@@ -0,0 +1,42 @@
...
+import java.time.{Duration, Instant}
+
+import scalaz._
+import Scalaz._
+
+object PioParse {
+  implicit val FunctorPioParse = new Functor[PioParse] {
+    def map[A, B](pa: PioParse[A])(f: A => B): PioParse[B] =
+      pa andThen (_.map(f))
+  }
```
tpolecat: This is problematic because `PioParse` is a type alias so `object PioParse` isn't really a companion; the instance isn't in implicit search scope. And `String => ?` is already a functor which is probably going to cause trouble. So I suggest newtyping instead of aliasing:

```scala
case class PioParse[A](run: String => Option[A]) {
  def apply(s: String): Option[A] = run(s) // so we can still treat it like a function
}
```

This will let us define a distinct set of instances (it's isomorphic to `Kleisli[Option, String, A]` so it will have a bunch, but I think `Functor` is probably good enough for now).

modules/ocs2/src/main/scala/gem/ocs2/pio/PioPath.scala
```diff
@@ -0,0 +1,276 @@
...
+  *
+  * Node with an optional <instrument>` element that, if it exists, has a
+  * required "type" attribute.  When decoded, any path that included a `?`
+  * but not a `*` will produce an `Option` result.  Any path including a
+  * `*` will produce a `List`.
+  *
+  * If a required element is missing, the result will be `MissingKey`
+  * `PioError` when decoded.
+  */
+object PioPath {
```
tpolecat: Neato.

modules/ocs2/src/main/scala/gem/ocs2/pio/PioDecoder.scala
```diff
@@ -0,0 +1,39 @@
...
+  def apply[A](f: Node => PioError \/ A): PioDecoder[A] =
+    new PioDecoder[A] {
+      def decode(n: Node): PioError \/ A =
+        f(n)
+    }
+
+  def enum[A](dataType: String)(m: (String, A)*): PioDecoder[A] =
+    fromParse[A](dataType) { m.toMap.lift }
+
+  def fromParse[A](dataType: String)(parse: PioParse[A]): PioDecoder[A] =
```
tpolecat: The `dataType` here is just for error reporting?

modules/ocs2/src/main/scala/gem/ocs2/DoobieClient.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def configureLogging(): Unit = List(
+    "edu.gemini.spModel.type.SpTypeUtil"
+  ).map(Logger.getLogger).foreach(_.setLevel(Level.OFF))
+
+  def ignoreUniqueViolation(fa: ConnectionIO[Int]): ConnectionIO[Int] =
+    for {
+      s <- HC.setSavepoint
+      n <- fa.onUniqueViolation(HC.rollback(s).as(0))
+      _ <- HC.releaseSavepoint(s)
```
tpolecat: N.B. this will leak the savepoint if there is a failure other than unique violation. This is probably fine because the whole transaction will be rolled back in that case, but the pedantic way to do this would be `… as(0)) ensuring HC.releaseSavepoint …`.

swalker2m:
Thanks for the feedback and pointers.  It's very much appreciated.  I think I made all the suggested changes. Also  I removed a couple of error cases since it turned out I wasn't using them anymore.

modules/ocs2/src/main/scala/gem/ocs2/pio/PioDecoder.scala
swalker2m: Correct, I couldn't see another way of making it clear what it was you were trying to parse.  Though I suppose we could require `A: ClassTag` and read the runtime type to place in the error?  It seems like that might clean it up a bit.

modules/ocs2/src/main/scala/gem/ocs2/pio/PioDecoder.scala
```diff
@@ -23,16 +24,16 @@ object PioDecoder {
         f(n)
     }
 
-  def enum[A](dataType: String)(m: (String, A)*): PioDecoder[A] =
-    fromParseFunction[A](dataType) { m.toMap.lift }
+  def enum[A: ClassTag](m: (String, A)*): PioDecoder[A] =
```
tpolecat: I just learned about this ... I think a better way to do this is with `shapeless.Typeable` which doesn't use the reflection API at runtime so it will work in scala-js and isn't prone to threading issues. It also does the name shortening for you, which can actually fail at runtime because Scala uses names that violate the JVM spec in some places.

```
scala> def foo[A](implicit ev: shapeless.Typeable[A]) = s"the type was ${ev.describe}"
foo: [A](implicit ev: shapeless.Typeable[A])String

scala> foo[gem.Dataset.Label]
res0: String = the type was Label
```

swalker2m: I tend to forget about scala-js.  Thanks for pointing this out.  It might have been okay in this case because nothing here is intended to be used from a client app but still this is better.

tpolecat:
👍

PR #41, opened 2017-03-31
deployment control program

tpolecat:
This adds a `ctl` project and `gemctl` command to sbt that lets you deploy/upgrade/rollback to a local or remote docker host. It will need to be built out a bit more as the requirements for tooling become more clear but I think it proves that we can do this in good style. If you view the `README.md` you'll get a sense for what it can do so far.

This is a big PR but it doesn't impact any existing code other than the build.


cquiroz:
This is a lot to digest, I'll check it later but somehow I'm still missing the discussion we were going to have about the pros/cons of docker over our current more primitive system with plain RPMs. And also see how it stands with respect to ITS

Could you think a bit how could we use this in the seqexec?

tpolecat:
The pros that I see are that it gives us very fine control over the environment and ensures that it's identical everywhere, it makes the environment part of the source code rather than an independent variable, makes it very easy to spin up new environments for testing, etc., and it makes upgrades and rollbacks very simple. The drawback is that it's another piece of software to deal with.

But we should continue to discuss. I just think I have gotten enough done here to show that it works, and I need to do something else for a while ;-)

cquiroz:
Well, rpms is also another piece of software to deal with (And probably a worse one) so we are just trading one for the other

It would be good to validate the concept with another application. I'd be very keen that we use the seqexec as test case. We already have test servers where we deploy the code

modules/ctl/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Packaging the Application
+
+In order to deply the app we must first create a **docker image** of our application. We do this via sbt-native-packager as follows:
+
+```
+> telnetd/docker:publishLocal
+```
+
+Image names have the form `<repo>/<artifact>:<version>`. Ours will be `geminihlsw/gem-telnetd:<version>` where the version is the git commit hash, or the hash followed by `-UNCOMMITTED` if there are uncommitted changes. This give us very fine-grained versioning and a reliable partial ordering to know that an upgrade is legal. Note that if the filesystem version isn't the same as the version computed by sbt you'll need to reload the project. The prompt will include a message to this effect.
```
cquiroz: Do you have access to this version at the application level if e.g. you want to display it on the UI

modules/ctl/README.md
```diff
@@ -0,0 +1,197 @@
...
+[info]    Deploying Gem.
+[info]      Creating Gem container from image 31f970aef7c7
+[info]        Container is gem-3 8a6a3e54ef47ba3e79ce1b473939d74cd83c1c786d137a2be30d54cb226d872f.
+[info]      Service is available at localhost:1234.
+
+[success] Total time: 18 s, completed Mar 30, 2017 4:09:32 PM
+>
+```
+
+Log in as root with no password, then change the password just to commit a change to the database.
```
cquiroz: Can you force the pwd change? This looks very risky to me

modules/ctl/README.md
```diff
@@ -0,0 +1,197 @@
...
+```
+> telnetd/docker:publishLocal
+```
+
+Image names have the form `<repo>/<artifact>:<version>`. Ours will be `geminihlsw/gem-telnetd:<version>` where the version is the git commit hash, or the hash followed by `-UNCOMMITTED` if there are uncommitted changes. This give us very fine-grained versioning and a reliable partial ordering to know that an upgrade is legal. Note that if the filesystem version isn't the same as the version computed by sbt you'll need to reload the project. The prompt will include a message to this effect.
+
+The image will be stored in docker's internal storage area, which shares "slices" with other images built from the same base to save space.
+
+For remote deployments we must publish and push the image to a repository where it can be seen by other machines. For now we're using docker-hub but it probably makes sense to set up an internal repository. To do this you need to create a docker-hub account and be added to the `geminihlsw` organization (talk to Rob), and then you can do `telnetd/docker:publish` to push the image up to the cloud. Note that you need to do this in order to perform remote deployments, even if they're to a VM on your own machine.
```
cquiroz: Can you explain more about docker-hub? Can anybody download an image? Is that a security risk?

modules/ctl/src/main/scala/command.scala
```diff
@@ -0,0 +1,149 @@
...
+          |  $progName --help            To see available commands.
+          |  $progName COMMAND --help    To see help on a specific command.
+         """.trim.stripMargin) *>
+         IO.putStr(Console.RESET) as None
+    }
+
+  // Parser implementation below
+
+  private lazy val machine: Parser[Boolean] =
+    switch(
```
cquiroz: Can you explain this syntax? does it com from the applicatives library

modules/ctl/src/main/scala/free/ctl.scala
```diff
@@ -0,0 +1,124 @@
...
+    server.map {
+      case Server.Remote(_, _) => true
+      case Server.Local        => false
+    }
+
+  def gosub[A](msg: String)(fa: CtlIO[A]): CtlIO[A] =
+    Free.liftF(CtlOp.Gosub(Level.Info, msg, fa))
+
+  def logMessage(level: Level, msg: String): CtlIO[Unit] =
+    Free.liftF(CtlOp.Gosub(level, msg, ().point[CtlIO]))
```
cquiroz: I have to wonder why we need all this machinery to write a log message

modules/ctl/src/main/scala/free/interp.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+  object InterpreterState {
+    val initial = InterpreterState(0, Map.empty)
+  }
+
+  /**
+   * Construct an interpreter of `CtlIO` given a `Config` and an `IORef` for the initial state. We
+   * carry our state in an `IORef` because it needs to be visible from multiple threads.
+   */
+  def interpreter(c: Config, state: IORef[InterpreterState]) = λ[CtlOp ~> EitherT[IO, Int, ?]] {
```
cquiroz: What's the lambda?

modules/ctl/src/main/scala/hi/Common.scala
```diff
@@ -0,0 +1,45 @@
...
+    gosub("Finding current running Gem container.") {
+      getUniqueRunningContainerWithLabel("edu.gemini.gem")
+    }
+
+  def getRunningPostgresContainer: CtlIO[Container] =
+    gosub("Finding current running Postgres container.") {
+      getUniqueRunningContainerWithLabel("edu.gemini.db")
+    }
+
+  def getDeployCommitForContainer(k: Container): CtlIO[DeployCommit] =
```
cquiroz: What's a `DeployCommit`

modules/ctl/src/main/scala/hi/deploy.scala
```diff
@@ -0,0 +1,267 @@
...
+            _ <- error(s"There is already a running deployment on ${c.server}")
+            _ <- exit[Unit](-1)
+          } yield ()
+      }
+    }
+
+  def createDatabaseContainer(nDeploy: Int, cDeploy: DeployCommit, iPg: Image, kPrev: Option[Container]): CtlIO[Container] =
+    gosub(s"Creating Postgres container from image ${iPg.hash}") {
+      for {
+        r <- isRemote // irritating … see below
```
cquiroz: it seems `r` isn't used here?

modules/ctl/src/main/scala/low/git.scala
```diff
@@ -0,0 +1,58 @@
...
+  case class Commit(hash: String)
+  object Commit {
+    implicit val OrderCommit: Order[Commit] = Order.orderBy(_.hash)
+  }
+
+  case class Tag(tag: String)
+
+  // Yields most recent tag matching the given pattern, if any
+  def mostRecentTag(pattern: Regex): CtlIO[Option[Tag]] =
+    shell("git", "tag").require {
```
cquiroz: This is a hard problem but how do we go accounting for differences on the command versions if e.g. different users have slightly different git configs that result in a different output?

Could/Should we have used a git library if such thing exists?

modules/ctl/src/main/scala/low/io.scala
```diff
@@ -0,0 +1,24 @@
...
+object io {
+
+  case class Output(exitCode: Int, lines: List[String])
+
+  def exec(cmd: String \/ List[String], f: String => IO[Unit] = (_ => IO.ioUnit)): IO[Output] =
+    IO {
+      import scala.sys.process._
+      import collection.mutable.ListBuffer
+      val b = ListBuffer[String]()
+      val x = cmd.fold(_.cat, _.cat).run(ProcessLogger { s =>
```
cquiroz: what's `_.cat`?

cquiroz:
This seems like a good idea in general and I'd really like to have an example closer to operation. The new seqexec is a prime candidate in my opinion. We could use it as a test bed to polish the concept and it is another reason to merge gem and ocs3

I have some concerns about the code which is seems very deep in FP, or at least very deep for my level. While the higher level operations are easy to understand I don't think I can follow the lower level code, it may be hard for me to maintain without external help

build.sbt
```diff
@@ -9,12 +9,42 @@ lazy val scalaTestVersion         = "3.0.0"
...
+  "> "
+}
+
 lazy val testLibs = Seq(
   "org.scalatest"  %% "scalatest"  % scalaTestVersion  % "test",
   "org.scalacheck" %% "scalacheck" % scalaCheckVersion % "test"
 )
 
 lazy val commonSettings = Seq(
+  scalaOrganization := "org.typelevel",
```
tpolecat: No downside. It gives us `-Ypartial-unification` in 2.11.8 which will be in the mainline compiler in 2.11.9 and we can switch back if we want. This flag enables the fix for SI-2717 which ends up fixing a lot of inference problems.

build.sbt
```diff
@@ -139,9 +175,24 @@ lazy val telnetd = project
...
+
+lazy val ctl = project
+  .in(file("modules/ctl"))
+  .settings(commonSettings)
+  .settings (
+    resolvers += "bmjames Bintray Repo" at "https://dl.bintray.com/bmjames/maven",
+    libraryDependencies ++= Seq(
+      "org.scalaz"  %% "scalaz-core"   % scalazVersion,
+      "org.scalaz"  %% "scalaz-effect" % scalazVersion,
+      "net.bmjames" %% "scala-optparse-applicative" % "0.5"
```
tpolecat: `gemctl` takes command-line arguments and switches and stuff, and this library provides a parser that also composes the help message you see when you do `gemctl deploy -h` for instance. It's also used by the telnet server.

It's a small totally pure lib and I'm a maintainer so I'm not too worried about pulling it in. We're going to port it to cats pretty soon.

docker.md
```diff
@@ -1,184 +0,0 @@
...
-    Goodbye.
-    Connection closed by foreign host.
-    gem$
-
-We can look at the running containers.
-
-    gem$ docker ps
-    CONTAINER ID        IMAGE                  COMMAND                  CREATED              STATUS              PORTS                    NAMES
-    e4b079b1c2a7        telnetd:0.1-SNAPSHOT   "bin/telnetd"            About a minute ago   Up About a minute   0.0.0.0:1234->6666/tcp   gem-1
-    2231cdc6ef9e        postgres:9.6.0         "/docker-entrypoint.s"   About a minute ago   Up About a minute   5432/tcp                 db-1
```
tpolecat: No, the port is local to the container. All the containers are on a private virtual network called `gem-net` and they can see each other and refer to each other by name. So the `gem-X` container refers to Postgres host `db-X` as if it were another machine.

This is how upgrade deployments work. We start up a new database server `db-42` and then stream all the data from `db-41` (or whatever the previous one is called) before shutting it down.

If we want to expose ports to the outside world on non-random ports then yes you can have a conflict, as it the case of the public telnet service mapped to localhost:6666. If you try to start up another gem container it will fail because the public port is taken.

modules/ctl/README.md
tpolecat: Not at the moment but we can use `sbt-buildinfo` or similar to make it available. Good idea.

modules/ctl/README.md
tpolecat: Yeah it's public but our organization is hidden so I'm not sure how anyone would find the images. I don't think it's a security risk since we're the only people who can publish to that org and it's all open source code anyway, but just as a logistical thing it makes more sense to have a local instance. I just haven't set it up yet.

modules/ctl/README.md
tpolecat: Sure, we could do that. This is the root user on Gem (not the server) just to be clear. In any real deployment the root password hash will be copied over with the rest of the application data, so this is really just an issue for initial deployments.

modules/ctl/src/main/scala/command.scala
tpolecat: Yeah it's from the optparse library. It constructs a "switch" like `--force` that's just an off/on argument.

modules/ctl/src/main/scala/free/ctl.scala
tpolecat: It's constructing a program that will log the message when it's interpreted, and logging a simple message is a special case of logging a message and then running a sub-program at a higher indentation so there's some extra cruft in there. But it keeps the algebra simple because there's only one log constructor.

modules/ctl/src/main/scala/free/interp.scala
tpolecat: This is a new thing in kind-projector that makes it easier to construct natural transformations and things like it.

```scala
λ[F ~> G]{ ... }
```

Is more or less the same thing as 

```scala
new (F ~> G) {
  def impl[A]: F[A] => G[A] = ...
  def apply[A](fa: F[A]): G[A] = impl[A](fa)
}
```

There's more info in the kind-projector doc. This is new but already pretty standard now. It's used in cats and other libs.

modules/ctl/src/main/scala/hi/Common.scala
tpolecat: It's a commit hash along with a tag that indicates whether there are uncommitted files or not. So if you deploy a version with local changes you can distinguish it from the HEAD commit.

```scala
case class DeployCommit(commit: Commit, uncommitted: Boolean) {
  def imageVersion = if (uncommitted) s"${commit.hash}-UNCOMMITTED" else commit.hash
}
```

modules/ctl/src/main/scala/hi/deploy.scala
tpolecat: It's used below to decide whether or not to include quote marks around an argument, because in the remote cases the command is added to an `ssh` command and it needs different quoting for args that contain spaces. It's kind of hacky and can probably be fixed at a lower level. Originally I was treating local deploys uniformly and just `ssh`ing back to localhost but that seemed needlessly complicated so I got rid of that but now I have to deal with differences in quoting.

modules/ctl/src/main/scala/low/git.scala
tpolecat: This is a tough call. I wanted to keep it as simple as possible (like a shell script) so you can look at the code and the `-v` log output and know what's going on. We can switch to a library if there end up being problems, but there's an awful lot of automation out there that depends on git output so I'm not *that* worried about it. But yeah it's a potential concern. Docker as well.

modules/ctl/src/main/scala/low/io.scala
tpolecat: It's on the syntax from `scala.sys.process._` … it turns a string or list of strings into a `ProcessBuilder`. No idea where the name came from.

tpolecat:
@cquiroz if you can point to the code you don't understand I can explain it. The intent is for this to be very straightforward.

build.sbt
cquiroz: We should likely discuss this a bit and if we are going to use it, then move ocs3 too along

modules/ctl/src/main/scala/free/ctl.scala
cquiroz: I presume this is going to be a lot slower than `log.info(msg)`

modules/ctl/src/main/scala/free/interp.scala
cquiroz: It is probably very standard if you are very deep into the FP stack but becomes very obscure if you are not. So the question is how do we get there. At least I have a very vague idea of what a natural transformation is so having more syntax on top doesn't make it simpler

modules/ctl/src/main/scala/hi/Common.scala
cquiroz: Could we somehow limit to forbid production/test deployment with uncommited changes?

modules/ctl/src/main/scala/low/git.scala
cquiroz: I see your point. This is a problem also when the shell tools evolve and their syntax change. It would probably be good to say, this works with this version at least.

cquiroz:
It is probably very straightforward when you know very well your way around with free monads. My concern is that for people who don't (like me), it requires a substantial amount of training to get there.

This is a concern in general, how do we decide what everybody needs to know to work in a project and when do we allocate time to learn

build.sbt
jdnavarro: 👍 for switching to typelevel compiler

FWIW, we already discarded some code in the core of the seqexec with type lambdas because what we gained with them wasn't worth the readability mess we were introducing. However that would have been a different story with kind projectors, which I'm not mistaken are builtin in the typelevel compiler.

cquiroz: I think kind projectors syntax comes from the `kind-projector` plugin and are independent of the compiler. AFAIK the typelevel compiler brings smaller changes like the fix @tpolecat mentioned above but should change the language too much

modules/ctl/README.md
```diff
@@ -0,0 +1,197 @@
...
+Connection to 192.168.99.100 closed.
+$ _
+```
+
+You can do `docker-machine stop` to shut down the VM and `docker-machiner start` to start it back up again. Anyway we're now all set for a "remote" deployment.
+
+## Deploying and Upgrading Gem
+
+Ok so `gemctl` is an alias in the build that lets you run deployment stuff from the sbt prompt. If you just type `gemctl` at the sbt prompt you'll get a help message. `gemctl -h` will list available commands, and `gemctl <command> -h` will show details on any specific command. If the help messages aren't helpful, please improve them.
+
```
jdnavarro: I'm really interested in the interaction between sbt and docker. Are you using some existing sbt plugin for this? If this so which one? Or does `gemctl` interacts with docker through the docker CLI only?

modules/ctl/src/main/scala/free/ctl.scala
jdnavarro: One advantage of this approach, as opposed to using plain `log.info(msg)`, is that I think it gives more flexibility handling *control operations* *side-effect*s.

We can have different interpreters for this opertions *mini-DSL*, for example we can have one interpreter for testing where all the operations and its side effects are mocked and captured, and other where the side-effects happen for real.

This also makes it easier to combine it with other *things that have effects* (poor euphemism for other monads), instead of shove it all on `Task` or `IO` as the common denominator for combination of anything that has some side effect even if it does really nothing in the real world.

swalker2m:
> It is probably very straightforward when you know very well your way around with free monads. My concern is that for people who don't (like me), it requires a substantial amount of training to get there.
>
> This is a concern in general, how do we decide what everybody needs to know to work in a project and when do we allocate time to learn

I'm not sure what the answer is here, but it is a concern for me too.  Arturo and I have been discussing it as well.  I too often struggle with advanced FP concepts and find it hard to track down explanations that I can digest.  At the same time, I know there is a lot of value in these approaches and don't want to sell ourselves short or limit us to just what is available in the standard Scala library.  Part of the solution is looking at the code and asking questions as we have been doing.  An example like this in the context of our project is invaluable.  Still, it feels like we need to make space for discussion and leveling-up across the team from time to time.

modules/ctl/src/main/scala/free/interp.scala
jdnavarro: From by biased background from Haskell, this looks like a normal type signature that you'd like to easily express in any pure FP language. So I'm glad that kind projectors makes this cleaner, because without them something that should be straightforward is made more difficult because of syntactic mess.

But this this discussion, as many others before, seems to be leading to do we use Scala as a more modern Java with some styling borrowed from FP or are we invested on truly using the Scala as a true FP language.

modules/ctl/src/main/scala/free/ctl.scala
cquiroz: I sort of see the point, I guess you could for example store the logs and replay them later if you need with the correct interpreter

modules/ctl/src/main/scala/free/interp.scala
swalker2m: @jdnavarro I think there is consensus that using Scala just as a more modern Java is out of the question.  That wouldn't improve our current situation where we are afraid to make changes to the codebase because we cannot control all the unintended side-effects.  I think it's more a question of how do we all reach the level of expertise that we need, do we all understand the benefits and justification of doing something one way or another, etc.

build.sbt
tpolecat: Yes, that's right. We're only using it for `-Ypartial-unification` at this point, which is in Lightbend Scala 2.12 but not in 2.11 yet. When 2.11.10 is out (soon!) we can switch back.

modules/ctl/src/main/scala/free/ctl.scala
tpolecat: Ok.

When you use `Free` you construct a program that is a value, and interpret it later. It's exactly the same idea as using `IO` or `Task` but instead of being able to do absolutely anything you can restrict operations to a particular domain, "shell scripting" in this case, by providing a small number of primitive operations and building things up. That's what we're doing here. Everything you can do in this language is defined in `ctl.scala`:

```scala
/** ADT of low-level control operations. */
sealed trait CtlOp[A]
object CtlOp {
  case class  Exit[A](exitCode: Int)                              extends CtlOp[A]
  case object GetConfig                                           extends CtlOp[Config]
  case class  Gosub[A](level: Level, msg: String, fa: CtlIO[A])   extends CtlOp[A]
  case class  Shell(remote: Boolean, cmd: String \/ List[String]) extends CtlOp[Output]
}
```

So all the deployment code, etc., is define in terms of these 4 operations. Logging is defined in terms of a `Gosub` operation that says "log the execution of this sub-program". And the interpreter `interp.scala` implements this by building a tree of log statements that shows the logical execution. The `logMessage` operation above uses an empty sub-program, so the logged message has no child messages.

And @cquiroz yes, exactly, you can change the interpreter to store the logs to a file or a JSON object or whatever. The *meaning* of these operations is entirely up to the interpreter.

modules/ctl/src/main/scala/free/interp.scala
tpolecat: I will rewrite this without the lambda. It's not a big deal. If we end up doing a lot of higher-order functions like this then the convenience will outweigh the cost of learning new syntax.

modules/ctl/src/main/scala/hi/Common.scala
tpolecat: Yes, exactly. That was my thinking. There's no notion of deployment environments ("production") yet but this gives us a way to do it. I think we would also want to confirm that the production deploy commit has been merged into `master` and is on GH.

modules/ctl/src/main/scala/free/interp.scala
cquiroz: I don't think it is necessary to rewrite. We just need to find a way to distribute the knowledge

modules/ctl/src/main/scala/low/git.scala
tpolecat: Yeah, in fact we could check the version as part of the startup if we wanted to.

modules/ctl/README.md
tpolecat: No `gemctl` is just an alias that run the program in this PR. It interacts with docker through the CLI (via ssh). But packaging the application into an image *is* done via an sbt plugin (sbt-native-packager).

tpolecat:
Would it be helpful if I did a talk about the free monad just so we're all on the same page? Most of the code in this PR is really no different than what we have done before with `Task` and `IO`, it's just done in a more narrowly-defined effect type `CtlIO`. The translation of this type to `IO` in the end is the only bit that we haven't done a lot of (but you have a Free language somewhere don't you @cquiroz?).

cquiroz:
I would certainly be happy with a talk. Actually we could consider doing more talks in a bunch of topics (e.g. scalajs-react or applicatives) I think it would be useful in general though of course we'd need time to prepare them

I did the ldap connector using `Free` but I don't think I have a good enough grasp of the concepts. For example I have no idea how could I combine my ldap algebra with e.g. some authorization algebra if that were needed

cquiroz:
Sorry for hijacking this PR with the discussion of code details, and certainly the high level API you have built is beautiful.

In general I'm quite interested on this work as it is potentially a big deal on how we could provide fixes or features faster to production, and wouldn't mind stop fuzzing with rpms every time a new CentOS release is installed.

A couple of small questions:
* Is docker a userland application? If we can just install it as user software, deployment would require less intervention from IT
* A problem we've had in general is where to store sensitive data, e.g. db passwords. The solution in ocs and ocs3 is to have private svn repo which is not very satisfactory. Any other ideas?
* With talked a bit about deploying after every merge (at least for test). Could we do that on travis?

tpolecat:
> Is docker a userland application? If we can just install it as user software, deployment would require less intervention from IT

You have to install it as root but thereafter anyone in the `docker` group can control it without `sudo`. The containers themselves are isolated from each other and from the host OS, but because you can mount host filesystems and bind host sockets you can't let arbitrary users control `dockerd`. There's somewhat more info [here](https://docs.docker.com/engine/security/security/). 

But yeah this is one of the big reasons people use docker. It lets the software people run their systems in the environment they want, and create/manage them freely without having to deal with corporate IT groups.

> A problem we've had in general is where to store sensitive data, e.g. db passwords. The solution in ocs and ocs3 is to have private svn repo which is not very satisfactory. Any other ideas?

I don't have a better idea offhand but everybody has this problem, so we should ask around.

> With talked a bit about deploying after every merge (at least for test). Could we do that on travis?

If we wanted to do it from Travis we would need a hole in our network that Travis could call back into, which I think is probably a bad idea. I was thinking we could set up local Jenkins to do this but I haven't tried to set it up.

cquiroz:
On the contrary. I'm very keen on using and understanding docker and how it helps, and I'm very amazed that @tpolecat was able to get it to work without using bash magic scripts. I'm eager that we try this ASAP on the seqexec but we should probably merge the projects first.

My concerns were that following the code required a good grasp on advanced FP techniques beyond my abilities. As you mentioned given the need to maintain it, we may need to take some time to understand it better

tpolecat:
> Honestly I have to wonder if I'm capable of maintaining it, but I think so with a bit of time.

So, this is really just a normal monadic program that could have been written entirely in `EitherT[IO, Int, ?]`, but by using `Free` we get simpler types, we limit the allowed operations, and we can easily change the back-end implementation because it's all in one place.

I wrote a talk on free things and tested it out at our local Scala group and it needs a bit of work, but I can give it again for our group, maybe next week. I want to get the asterism stuff in the can first.

jdnavarro:
>> With talked a bit about deploying after every merge (at least for test). Could we do that on travis?

> If we wanted to do it from Travis we would need a hole in our network that Travis could call back into, which I think is probably a bad idea. I was thinking we could set up local Jenkins to do this but I haven't tried to set it up.

👍 for the local Jenkins machine. 3rd party CI services like Travis save us from having to maintain yet another local machine, but for our particular needs, I think it's worth the investment, we get more observability when debugging broken builds.

tpolecat:
Ok this is compiling again so it's ready to merge.

PR #43, opened 2017-04-07
Add Schema Checks

tpolecat:
This is a kind of a big PR, sorry, but it takes care of something I glossed over when we were bootstrapping the project. This adds tests that verify that all the SQL literals compile, and that all the type mappings we're asserting are consistent with the schema. So this will let us do schema changes more confidently because most breaking changes will be caught by these tests.

So here's what I did.
- I factored out all the `Query/Update` expressions for each DAO into a nested `Statements` object. We need these available because this is the interface that understand how to do schema checks. This is the most disruptive change but it's entirely mechanical.
- I added tests for every statement in every DAO. In a few cases these caught errors so there are some small changes in the code and schema for those. In other cases the mapping is correct in principle but disagreed with the way Postgres reports the metadata (specifically with domain types and columns with check constraints) so there are a handful of new mappings to silence the errors.
- I upgraded to doobie 0.4.2-SNAPSHOT which provides logging facilities that are helpful when diagnosing failures. This will be a real release soon.
- A few of the tests in `SmartGcalDao` are failing on the master branch so I just commented those out.

I'm still cleaning up but you can have a look if you like.

```
> db/test
[info] ObservationCheck:
[info] ObservationDao.Statements
[info] - should insert
[info] - should selectIds
[info] - should selectFlat
[info] - should selectAllFlat
[info] LogCheck:
[info] LogDao.Statements
[info] - should insert
[info] SmartGcalCheck:
[info] SmartGcalDao.Statements
[info] - should selectF2byLamp
[info] - should selectF2byBaseline
[info] - should insertSmartF2
[info] EventLogCheck:
[info] EventLogDao.Statements
[info] - should insertEvent
[info] - should selectAll
[info] SemesterCheck:
[info] SemesterDao.Statements
[info] - should canonicalize
[info] SmartGcalSpec:
[info] SmartGcalDao
[info] - should fail when there is no corresponding mapping
[info] - should fail when the location is not found
[info] - should fail when the location is not a smart gcal step
[info] ProgramCheck:
[info] ProgramDao.Statements
[info] - should selectFlat
[info] - should selectBySubstring
[info] - should insertArbitraryProgramIdSlice
[info] - should insertDailyProgramIdSlice
[info] - should insert
[info] UserCheck:
[info] UserDao.Statements
[info] - should select
[info] - should selectWithRoles
[info] - should changePassword
[info] GcalCheck:
[info] GcalDao.Statements
[info] - should insert
[info] - should select
[info] DatasetCheck:
[info] DatasetDao.Statements
[info] - should selectAll
[info] - should insert
[info] StepCheck:
[info] StepDao.Statements
[info] - should delete (1)
[info] - should delete (2)
[info] - should allGenericOnly
[info] - should oneGenericOnly
[info] - should allF2Only
[info] - should selectAllEmpty
[info] - should selectOneEmpty
[info] - should insertF2Config
[info] - should insertScienceSlice
[info] - should insertSmartGcalSlice
[info] - should insertGcalStep
[info] - should insertDarkSlice
[info] - should insertBiasSlice
[info] - should insertBaseSlice
[info] ScalaTest
[info] Run completed in 3 seconds, 719 milliseconds.
[info] Total number of tests run: 40
[info] Suites: completed 11, aborted 0
[info] Tests: succeeded 40, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 40, Failed 0, Errors 0, Passed 40
[success] Total time: 6 s, completed Apr 7, 2017 4:23:17 PM
```


modules/db/src/test/scala/gem/dao/check/Check.scala
```diff
@@ -0,0 +1,64 @@
...
+import doobie.imports._
+import doobie.scalatest.imports._
+import org.scalatest._
+
+import scalaz._, Scalaz._
+
+/** Trait for tests that check statement syntax and mappings. */
+trait Check extends FlatSpec with Matchers with QueryChecker {
+
+  def transactor = DriverManagerTransactor[IOLite](
```
cquiroz: I guess this means that the tests go to the database? I suppose that being the syntax tied to `postgress` there is no sense on using e.g. H2 to make them faster

modules/db/src/test/scala/gem/dao/check/Check.scala
```diff
@@ -0,0 +1,64 @@
...
+import edu.gemini.spModel.core.{ Site, ProgramType, Semester, _ }
+
+import doobie.imports._
+import doobie.scalatest.imports._
+import org.scalatest._
+
+import scalaz._, Scalaz._
+
+/** Trait for tests that check statement syntax and mappings. */
+trait Check extends FlatSpec with Matchers with QueryChecker {
```
cquiroz: Is `QueryChecker` from doobie? What does it do?

build.sbt
```diff
@@ -22,7 +24,7 @@ lazy val commonSettings = Seq(
     "-feature",
     "-language:higherKinds",
     "-unchecked",
-    "-Xfatal-warnings",
+    // "-Xfatal-warnings",
```
cquiroz: Why do you need to disable this? I'm liking this option more and more

build.sbt
```diff
@@ -80,7 +83,8 @@ lazy val db = project
       |  "jdbc:postgresql:gem",
       |  "postgres",
       |  "")
-      |import xa.yolo._
+      |val y = xa.yolo
```
cquiroz: What's `yolo` here? Is this to initialize the console?

modules/db/src/main/scala/gem/dao/LogDao.scala
```diff
@@ -6,19 +6,26 @@ import java.util.logging.Level
...
+
+  object Statements {
+
+    def insert(user: User[_], level: Level, pid: Option[Program.Id], msg: String, t: Option[Throwable], elapsed: Option[Long]): Update0 =
+      sql"""
+        INSERT INTO log (user_id, "timestamp", level, program, message, stacktrace, elapsed)
+             VALUES (${user.id}, now(), $level :: log_level, $pid, $msg, ${t.map(stack)}, $elapsed)
+       """.update
+
+    private def stack(t: Throwable): String = {
```
cquiroz: I think this will store the whole stack trace into the db. Do you see any troubles if the trace is very long

modules/db/src/main/scala/gem/dao/UserDao.scala
```diff
@@ -13,29 +13,10 @@ object UserDao {
...
   def select(id: String): ConnectionIO[User[Nothing]] =
-    sql"""
-      SELECT id, first, last, email, staff
-      FROM gem_user
-      WHERE id = $id
-    """.query[(String, String, String, String, Boolean)]
-       .unique.map { case (i, f, l, e, s) =>
-         User(i, f, l, e, s, Map.empty)
-       }
+    Statements.select(id).unique
```
cquiroz: There is a pattern where you have moved the sql from the `Dao` level to an internal `Statements` object. Any reason for that or is it just a style change?

modules/db/src/main/scala/gem/Log.scala
```diff
@@ -7,7 +7,7 @@ import java.util.logging._
 import scalaz._, Scalaz._, scalaz.concurrent.Task
 
 /** Logger that logs to the database and a JDK logger, on its own thread, in effect type M. */
-class Log[M[_]: Monad: Catchable] private (name: String, xa: Transactor[Task], delay: Capture[M]) {
+class Log[M[_]: Monad: Catchable] private (name: String, xa: Transactor[Task, _], delay: Capture[M]) {
```
cquiroz: Is this a change with the doobie update?

cquiroz:
If we were to run these tests with travis we'd need to install `postgress` and initialize the db right?

tpolecat:
> If we were to run these tests with travis we'd need to install postgress and initialize the db right?

Yes, I will set that up in another PR.

modules/db/src/test/scala/gem/dao/check/Check.scala
tpolecat: Yeah it's just a trait that provides a utility method to create schema checks and print out errors in a nice way. Not much to it.

modules/db/src/test/scala/gem/dao/check/Check.scala
tpolecat: Correct. The point is to check the queries and type mappings against the actual schema, which is very specific to the database.

build.sbt
tpolecat: Ah sorry, I'll turn that back on.

build.sbt
tpolecat: It's some convenience syntax for the REPL that lets you run queries and print the results without having to mention which database you're using.

modules/db/src/main/scala/gem/dao/LogDao.scala
tpolecat: It's stored in a `TEXT` field which can be arbitrarily long. We'll need to keep an eye on storage and maybe have a periodic task that deletes old log entries.

modules/db/src/main/scala/gem/dao/UserDao.scala
tpolecat: Yeah, checking needs to see a more precise structure than `ConnectionIO`.

`sql" ... $a ... $b ... ".query[Foo]` gives you a `Query0` which has enough information to perform a metadata check. That's what lives in the `Statements` modules and what is used by the tests.

When you then say `.list` you have a `ConnectionIO` which no longer knows it's a query; it's just a program that produces a `List[Foo]`. This is what you care about as a user and what you want to expose on the DAO.

Analogous for updates.

modules/db/src/main/scala/gem/Log.scala
tpolecat: Yes, `Transactor` has a second type parameter now.

modules/db/src/main/scala/gem/dao/UserDao.scala
cquiroz: So when you run the query it will not only read the data but also verify that the data types between the tuple in `query` and the db match? that would be neat

Does it do it with JDBC metadata?

tpolecat: Yeah doobie fetches all the metadata and checks all the parameter/column mappings with the JDBC spec and complains about mismatched types and nullity. So if you try to read a `VARCHAR` into an `Int` it will fail, for instance.

modules/db/src/main/scala/gem/dao/DatasetDao.scala
```diff
@@ -7,30 +7,48 @@ import doobie.imports._
 import java.time.Instant
 
 object DatasetDao {
```
swalker2m: Since the DAOs don't do much other than delegate to `Statements`, does it make sense to just get rid of them and call `.run`, `.list` etc. from the places where the statements are used?

modules/db/src/main/scala/gem/dao/GcalDao.scala
```diff
@@ -10,48 +10,62 @@ import java.time.Duration
...
-    filter:    GcalFilter,
-    diffuser:  GcalDiffuser,
-    shutter:   GcalShutter,
-    expTime:   Duration,
-    coadds:    Int)
-  {
-    def toGcalConfig: Option[GcalConfig] =
-      GcalLamp.fromConfig(continuum, ArArc -> ar_arc, CuArArc -> cuar_arc, ThArArc -> thar_arc, XeArc -> xe_arc).map { lamp =>
-        GcalConfig(lamp, filter, diffuser, shutter, expTime, coadds)
+    // CoAdds has a DISTINCT type due to its check constraint so we need a fine-grained mapping
```
swalker2m: Having a constraint actually creates a new type?

modules/db/src/main/scala/gem/dao/DatasetDao.scala
tpolecat: I think as we build it out DAOs will do more operations that are compositions of multiple statements (we see a bit of this already) so I'd prefer to keep it as-is for now, and treat `Statements` modules as an implementation detail. Could change my mind but I think it will end up making sense.

modules/db/src/main/scala/gem/dao/GcalDao.scala
tpolecat: It's a different type from JDBC's point of view, so we need to tell doobie that it's not an error. That's really all that's going on here. I did it case by case but may be able to abstract some stuff out.

tpolecat:
Ok everything is passing now. All our SQL is consistent with the schema.

PR #44, opened 2017-05-23
OCS2-dependency-free Import Wrapup

swalker2m:
This PR finishes work on the OCS2 importer that doesn't require OCS2 jar files.  In particular, it

* updates the `README.md` file with information on how to use the import options

* updates the Smart Gcal data importer to clean `gcal` table data and to run the clean and import steps in a single transaction after the `.csv` files are successfully read

* adds a data migration to import Smart Gcal data for F2, making the test cases work out of the box again

* deletes the old `importer` module which required OCS2 jar files


README.md
```diff
@@ -63,12 +68,73 @@ Right now just a sketch is imported:
...
+```
+sbt> ocs2/runMain gem.ocs2.SmartGcalImporter
+```
+
+#### Import Server
+
+You can run an import server which will import programs or observations on demand directly from a running OCS2 Observing Database.  You start it with
+
+```
+sbt> ocs2/runMain gem.ocs2.ImportServer [odb-hostname]
```
cquiroz: Have you tried importing all the database? Any errors?

modules/importer/src/main/scala/ConfigReader.scala
```diff
@@ -1,308 +0,0 @@
-package gem
```
cquiroz: Why did you remove these classes?

swalker2m: These are part of the old `importer` module.  It's been replaced with a version that doesn't rely upon having access to OCS2 classes.

README.md
swalker2m: Actually, I probably need to update it.  I think there are new F2 filters this semester?

sraaphorst: Yes, there are two new K filters. I can't remember offhand if they are `K_SHORT` and `K_LONG` or `K_RED` and `K_BLUE`. I think the latter?

tpolecat:
I think the fastest way from here to there is to merge the Travis PR, then you can rebase this branch and `git revert 7d6c7dc` to re-enable those tests.

tpolecat:
Otherwise we will have to turn Travis off since the check is required now but the config doesn't exist in this branch.

PR #45, opened 2017-05-29
Add GMOS (Part A)

swalker2m:
This is the first part of a larger project to add GMOS to the `gem` model.  I wanted to stop and check-in with everyone before continuing though because this has required a few adjustments that I'm not sure about:

* The definition of an identifier has been expanded to accept underscore characters.

* Adds our core model `Angle` class to the set of types available in the generated enums.  This is the first non-standard Java/Scala class we've used in the enums.  This required adding the core jar file to the `project/lib` directory.  It also exists in `modules/core` which is suboptimal. I played with it for a few minutes but couldn't figure out how to just have the jar file in one place.  Suggestions?

* Removes bogus filter option`None` and disperser `Mirror` from the list of choices.  These are values that represent the lack of a value.  For that we already have the `Option` type which seems more appropriate.  So GMOS would have an `Option[Filter]` instead of a `Filter`, etc.  I'm not sure whether this will pan out in practice.


modules/sql/src/main/resources/db/migration/V001__Initial_Setup.sql
```diff
@@ -61,7 +61,7 @@ ALTER TYPE f2_decker OWNER TO postgres;
 --
 
 CREATE DOMAIN identifier AS character varying(32)
-	CONSTRAINT identifier_check CHECK (((VALUE)::text ~ '^[A-Z][A-Za-z0-9]*$'::text));
+	CONSTRAINT identifier_check CHECK (((VALUE)::text ~ '^[A-Z][A-Za-z0-9_]*$'::text));
```
cquiroz: Is this program id? I've never seen `_` used but I guess it is valid

swalker2m: It's the name of any enumeration element like a GMOS disperser:

```
CREATE TABLE e_gmos_north_disperser (
    id             identifier            PRIMARY KEY,
...
);
```

The check verifies that these names follow the given regex.  Since GMOS filter options combine various filters like `ZPrime` and `CaT` into `ZPrime_CaT`, I added support for underscore.

tpolecat:
> I played with it for a few minutes but couldn't figure out how to just have the jar file in one place. Suggestions?

I think this will work.

```scala
unmanagedJars in Compile +=
  baseDirectory.value / "../modules/core/lib/edu-gemini-spmodel-core_2.11-2016001.1.1.jar"
```

Otherwise 👍

PR #46, opened 2017-05-31
travis

tpolecat:
This enables Travis CI with local Postgres 9.5, set up as the project `README.md` instructs so anything we do locally should work here.

Note that a few tests failed and I had to disable them. I did this by cherry-picking the first commit from #43, which I would like to merge before you try to fix the test failures so we can avoid conflicts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+language: scala
+
+dist: trusty
+
+jdk:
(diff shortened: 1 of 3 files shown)


tpolecat:
Ok so next in line is #44 which will revert 7d6c7dc.

PR #49, opened 2017-06-05
initial static config

tpolecat:
Initial implementation of static config. This adds a second type parameter to `Observation`, intended for a `StaticConfig` value. OCS2 import and Gem db stuff are all stubbed out; `StaticConfig` is isomorphic to `Instrument` right now.

We can represent an observation with "matching" static and dynamic configurations thus, where the existential ensures that they line up.

```scala
type RealizedObservation =
  Observation[StaticConfig.Aux[A], Step[DynamicConfig.Aux[A]]] forSome { type A }
```

### DIFF
--- a/modules/core/src/main/scala/gem/Observation.scala
+++ b/modules/core/src/main/scala/gem/Observation.scala
@@ -1,14 +1,12 @@
 package gem
 
-import gem.enum.Instrument
-
 import scalaz._, Scalaz._
 
-case class Observation[S](
+case class Observation[S, D](
   id: Observation.Id,
   title: String,
-  instrument: Option[Instrument], // redundant? this is on the steps too
-  steps: List[S])
+  staticConfig: S,
+  steps: List[D])
 
 object Observation {
 
@@ -32,10 +30,10 @@ object Observation {
 
   }
 
-  implicit def ObservationTraverse[T]: Traverse[Observation[?]] =
-    new Traverse[Observation[?]] {
-      def traverseImpl[G[_]: Applicative, A, B](fa: Observation[A])(f: A => G[B]): G[Observation[B]] =
-        fa.steps.traverse(f).map(ss => fa.copy(steps = ss))
-    }
+  // implicit def ObservationTraverse[T]: Traverse[Observation[?]] =
+  //   new Traverse[Observation[?]] {
+  //     def traverseImpl[G[_]: Applicative, A, B](fa: Observation[A])(f: A => G[B]): G[Observation[B]] =
+  //       fa.steps.traverse(f).map(ss => fa.copy(steps = ss))
+  //   }
 
 }
--- a/modules/core/src/main/scala/gem/Observation.scala
+++ b/modules/core/src/main/scala/gem/Observation.scala
@@ -30,10 +30,10 @@ object Observation {
 
   }
 
-  // implicit def ObservationTraverse[T]: Traverse[Observation[?]] =
-  //   new Traverse[Observation[?]] {
-  //     def traverseImpl[G[_]: Applicative, A, B](fa: Observation[A])(f: A => G[B]): G[Observation[B]] =
-  //       fa.steps.traverse(f).map(ss => fa.copy(steps = ss))
-  //   }
+  implicit val ObservationBitraverse: Bitraverse[Observation] =
+    new Bitraverse[Observation] {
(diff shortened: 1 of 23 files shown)


tpolecat:
Ok I think this is ready for review.

modules/core/src/main/scala/gem/Observation.scala
```diff
@@ -32,10 +30,10 @@ object Observation {
 
   }
 
-  implicit def ObservationTraverse[T]: Traverse[Observation[?]] =
-    new Traverse[Observation[?]] {
-      def traverseImpl[G[_]: Applicative, A, B](fa: Observation[A])(f: A => G[B]): G[Observation[B]] =
-        fa.steps.traverse(f).map(ss => fa.copy(steps = ss))
+  implicit val ObservationBitraverse: Bitraverse[Observation] =
```
swalker2m: A few words on `Bitraverse` and why we need it would be much appreciated.

modules/core/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -7,22 +7,22 @@ import java.time.Duration
 
 sealed trait SmartGcalKey
 
-sealed abstract class InstrumentConfig extends Product with Serializable {
+sealed abstract class DynamicConfig extends Product with Serializable {
+
+  type I <: Instrument with Singleton
```
swalker2m: I know, or thought I knew, what a `Singleton` is but what is it here?  I couldn't find the documentation for the trait.

modules/core/src/main/scala/gem/enum/package.scala
```diff
@@ -1,21 +1,10 @@
...
 
-import gem.config._
-
 // The members of this package are generated from database tables, which are the source of truth.
 // See project/gen2.scala for details. Associations with other model types, as needed, are provided
 // here as implicit classes wrapping the generated companion objects.
 package object enum {
 
-  /** Add mapping from InstrumentConfig to Instrument. */
-  implicit class InstrumentCompanionOps(companion: Instrument.type) {
```
swalker2m: We get this now by asking for `instrument` directly on the static or dynamic config?

modules/core/src/main/scala/gem/Observation.scala
tpolecat: `Observation[A, B]` is a functor in two ways … over the `A` and the `B`, so it can be a `Bifunctor` with `leftMap` and `rightMap`.

Even better, it's a *traversable* functor in two ways, which means that in addition to `.map` you can also `.traverse` in two ways, or `.bitraverse` to do both at once. Here we take an `Observation` whose children are (fake) database ids, and then we `bitraverse` to turn it into a `ConnectionIO` that materializes both the static and dynamic components.

```scala
// An observation with database references
val o: Observation[Int, Int] =
  Observation(Observation.Id.fromString("GS-2011A-Q-3-12").get, "Test", 123, List(456, 789))

// Programs to resolve a reference into a fetched value
def loadStaticConfig(id: Int): ConnectionIO[StaticConfig] = ???
def loadDynamicConfig(id: Int): ConnectionIO[DynamicConfig] = ???

// The result type of bitraversal
scala> :t o.bitraverse(loadStaticConfig, loadDynamicConfig)
ConnectionIO[Observation[StaticConfig, DynamicConfig]]

// Just do the left side
scala> :t Bitraverse[Observation].leftTraverse.traverse(o)(loadStaticConfig)
doobie.imports.ConnectionIO[gem.Observation[gem.config.StaticConfig,Int]]
```

There's no syntax for `.leftTraverse`/`.rightTraverse` so we might add that. If we run into a type with more than two kinds of foreign keys we'll need to define something more general.

modules/core/src/main/scala/gem/enum/package.scala
tpolecat: Yes.

modules/core/src/main/scala/gem/config/DynamicConfig.scala
tpolecat: It's a marker trait that the compiler knows, for singleton types. So this means `I` can only be a type like `Instrument.GmosSouth.type` (or a bogus subtype like `... with SomethingElse` or `Nothing` but these types are uninhabited). It can't be `Instrument` for instance. I don't know that we really need it here, it's mostly documentation that you should expect `I` to be a singleton type.

modules/core/src/main/scala/gem/Observation.scala
swalker2m: Thanks, very nice.

PR #50, opened 2017-06-05
Update to Scala 2.12

tpolecat:
This updates Gem to Scala 2.12


project/build.sbt
```diff
@@ -6,5 +6,3 @@ libraryDependencies ++= Seq(
 )
 
 addCompilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
-
-unmanagedJars in Compile += baseDirectory.value / "../modules/core/lib/edu-gemini-spmodel-core_2.11-2016001.1.1.jar"
```
tpolecat: Sorry, had to undo this change because we need both the 2.11 and 2.12 versions of spModel/core … when sbt 1.0 is available we can switch.

Actually I don't know how this ever worked because sbt runs in Scala 2.10 … hmm.

tpolecat: I guess we just got lucky and `Angle` happens to be binary compatible with 2.10

PR #52, opened 2017-06-12
F2 Static Config

swalker2m:
This is a preliminary PR for early feedback.  The purpose is to account for the separation of static vs. dynamic instrument config, using F2 as the initial instrument.  It is probably easiest to understand by looking first as the [data migration](https://github.com/swalker2m/gem/blob/d8b2e77cc9de0c833bce947580fb1b55e7c000e6/modules/sql/src/main/resources/db/migration/V012__Static_Config.sql).  That is also the part I'm least sure about.

At this point the schema update, queries, and OCS2 importer have been written but it needs test cases before merging.

### DIFF
--- a/modules/core/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/src/main/scala/gem/config/DynamicConfig.scala
@@ -21,8 +21,26 @@ sealed abstract class DynamicConfig extends Product with Serializable {
 }
 object DynamicConfig {
   type Aux[I0] = DynamicConfig { type I = I0 }
+
+  sealed abstract class Impl[I0 <: Instrument with Singleton](val instrument: I0) extends DynamicConfig {
+    type I = I0
+  }
 }
 
+final case class AcqCamDynamicConfig()   extends DynamicConfig.Impl(Instrument.AcqCam)
+final case class BhrosDynamicConfig()    extends DynamicConfig.Impl(Instrument.Bhros)
+final case class GmosNDynamicConfig()    extends DynamicConfig.Impl(Instrument.GmosN)
+final case class GmosSDynamicConfig()    extends DynamicConfig.Impl(Instrument.GmosS)
+final case class GnirsDynamicConfig()    extends DynamicConfig.Impl(Instrument.Gnirs)
+final case class GpiDynamicConfig()      extends DynamicConfig.Impl(Instrument.Gpi)
+final case class GsaoiDynamicConfig()    extends DynamicConfig.Impl(Instrument.Gsaoi)
+final case class MichelleDynamicConfig() extends DynamicConfig.Impl(Instrument.Michelle)
+final case class NiciDynamicConfig()     extends DynamicConfig.Impl(Instrument.Nici)
+final case class NifsDynamicConfig()     extends DynamicConfig.Impl(Instrument.Nifs)
+final case class NiriDynamicConfig()     extends DynamicConfig.Impl(Instrument.Niri)
+final case class PhoenixDynamicConfig()  extends DynamicConfig.Impl(Instrument.Phoenix)
+final case class TrecsDynamicConfig()    extends DynamicConfig.Impl(Instrument.Trecs)
+final case class VisitorDynamicConfig()  extends DynamicConfig.Impl(Instrument.Visitor)
 
 final case class F2SmartGcalKey(
   disperser: F2Disperser,
--- a/modules/core/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/src/main/scala/gem/config/DynamicConfig.scala
@@ -14,8 +14,8 @@ sealed abstract class DynamicConfig extends Product with Serializable {
 
   def smartGcalKey: Option[SmartGcalKey] =
     this match {
-      case f2: F2Config     => Some(f2.key)
-      case _ => None
+      case f2: F2DynamicConfig => Some(f2.key)
+      case _                   => None
     }
 
 }
@@ -48,7 +48,7 @@ final case class F2SmartGcalKey(
   fpu:       F2FpUnit
 ) extends SmartGcalKey
 
-final case class F2Config(
+final case class F2DynamicConfig(
   disperser:     F2Disperser,
   exposureTime:  Duration,
   filter:        F2Filter,
--- a/modules/core/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/src/main/scala/gem/config/DynamicConfig.scala
@@ -36,7 +36,6 @@ final case class F2Config(
   filter:        F2Filter,
   fpu:           F2FpUnit,
   lyotWheel:     F2LyotWheel,
-  mosPreimaging: Boolean,
   readMode:      F2ReadMode,
   windowCover:   F2WindowCover
 ) extends DynamicConfig {
--- a/modules/db/src/main/scala/gem/dao/ObservationDao.scala
+++ b/modules/db/src/main/scala/gem/dao/ObservationDao.scala
@@ -1,7 +1,7 @@
 package gem
 package dao
 
-import gem.config.StaticConfig
+import gem.config.{DynamicConfig, StaticConfig}
 import gem.enum.Instrument
 
 import doobie.imports._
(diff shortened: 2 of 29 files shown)


modules/sql/src/main/resources/db/migration/V012__Static_Config.sql
```diff
@@ -0,0 +1,54 @@
...
+-- mos-preimaging flag.  We have the same (static_id, instrument) foreign key
+-- into the static_config table and furthermore guarantee that the instrument
+-- is always Flamingos2.
+CREATE TABLE static_f2 (
+    static_id       integer,
+    instrument      identifier NOT NULL,
+    mos_preimaging  boolean    NOT NULL,
+    PRIMARY KEY (static_id, instrument),
+    FOREIGN KEY (static_id, instrument) REFERENCES static_config ON DELETE CASCADE,
+    CONSTRAINT is_f2 CHECK (instrument = 'Flamingos2')
```
cquiroz: This is what we discussed face to face. It seems fine to me to have this check there

modules/core/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -14,29 +14,46 @@ sealed abstract class DynamicConfig extends Product with Serializable {
...
+      case f2: F2DynamicConfig => Some(f2.key)
+      case _                   => None
     }
 
 }
 object DynamicConfig {
   type Aux[I0] = DynamicConfig { type I = I0 }
+
+  sealed abstract class Impl[I0 <: Instrument with Singleton](val instrument: I0) extends DynamicConfig {
+    type I = I0
```
swalker2m: Yes, I think it is path dependent.  I asked about`Singleton` too.  It's hard to find any documentation about.   [This entry](http://www.scala-archive.org/scala-Singleton-td1940630.html) is the most I could find.  Here I'm just aping what Rob did with the `StaticConfig` and creating instances for all the instruments so that the code compiles without warnings.  The goal though is to allow us to create an observation and know that its static and dynamic configs are consistent.   From #49:

```scala
type RealizedObservation =
  Observation[StaticConfig.Aux[A], Step[DynamicConfig.Aux[A]]] forSome { type A }
```

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -11,23 +11,39 @@ import scalaz._, Scalaz._
 object ObservationDao {
 
   def insert(o: Observation[StaticConfig, _]): ConnectionIO[Int] =
-    Statements.insert(o).run
+    for {
+      id <- StaticConfigDao.insert(o.staticConfig)
```
swalker2m: It will be run in a transaction yes.  In the `Importer` for example all the work associated with importing a single program or observation is done in a transaction by calling `transact` on the `ConnectionIO` we end up with.

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -1,39 +1,90 @@
...
-    }).asInstanceOf[StaticConfig.Aux[i.type]] // Scala isn't smart enough to know this
+  def insert(s: StaticConfig): ConnectionIO[Int] =
+    for {
+      id <- Statements.insertBaseSlice(s.instrument).withUniqueGeneratedKeys[Int]("static_id")
+      _  <- insertConfigSlice(id, s)
+    } yield id
+
+  private def insertConfigSlice(id: Int, s: StaticConfig): ConnectionIO[Int] =
+    s match {
+      case _: AcqCamStaticConfig   => 0.point[ConnectionIO]
```
swalker2m: It's a no-op.  I need a `ConnectionIO[Int]` to return but I don't really have anything to write to the database so I just put the value into a `ConnectionIO`.  I added cases for all the instruments so that I can get any observation to import as long as it has an instrument. The only one with an implementation at this point is F2.

modules/ocs2/src/main/scala/gem/ocs2/Legacy.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.reflect.runtime.universe.TypeTag
+
+import scalaz._
+import Scalaz._
+
+
+/** Legacy system (telescope, instrument, observe, calibration) key and parser
+  * definitions.
+  */
+object Legacy {
```
swalker2m: `Legacy` defines the mappings from strings that you find in OCS2 XML to their definitions in the new model.  I extracted it from where it was because it needs to be shared now by both the `SequenceDecoder` and the new `StaticDecoder`.  The static config is found in the OCS2 sequence's first step so it was convenient to just read it from the same place as the `SequenceDecoder`.

modules/sql/src/main/resources/db/migration/V012__Static_Config.sql
tpolecat: Yep this is what I had in mind. Looks good.

tpolecat:
This looks great so far.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -10,10 +10,13 @@ import scalaz._, Scalaz._
 
 object ObservationDao {
 
-  def insert(o: Observation[StaticConfig, _]): ConnectionIO[Int] =
+  def insert(o: Observation[StaticConfig, Step[DynamicConfig]]): ConnectionIO[Int] =
     for {
       id <- StaticConfigDao.insert(o.staticConfig)
       _  <- Statements.insert(o, id).run
+      _  <- o.steps.zipWithIndex.traverseU { case (s, i) =>
+              StepDao.insert(o.id, Location.unsafeMiddle((i + 1) * 100), s)
```
swalker2m: Might want to look at this change.  It seemed surprising to me that insert would add the static config but not the steps, so I added that.

modules/json/src/main/scala/gem/json.scala
```diff
@@ -50,4 +53,7 @@ package object json {
   implicit def oneAndCodec[F[_], A: CodecJson](implicit ev: CodecJson[F[A]]): CodecJson[OneAnd[F, A]] =
     CodecJson.derived[(A, F[A])].xmap { case (a, fa) => OneAnd(a, fa) } { oa => (oa.head, oa.tail) }
 
+  // Codec for GcalArcs
+  implicit def gcalArcsCodec: CodecJson[GcalArcs] =
+    CodecJson.derived[OneAnd[ISet, GcalArc]].xmap(oa => GcalArcs(oa.head, oa.tail.toList))(_.arcs)
```
swalker2m: It seems this cannot be derived automatically.  I suppose because it is a `sealed abstract case class`.

swalker2m:
I think I'm done with this and would like to move back to adding GMOS, if anybody wants to give it another look-over.  Sorry about all the emails for new commits!

modules/db/src/test/scala/gem/dao/DaoTest.scala
```diff
@@ -0,0 +1,26 @@
...
+    "jdbc:postgresql:gem",
+    "postgres",
+    ""
+  )
+
+  def withProgram[A](test: ConnectionIO[A]): A =
+    (for {
+      _ <- ProgramDao.insert(Program(pid, "Test Prog", List.empty[Step[Nothing]]))
+      a <- test
+      _ <- sql"""DELETE FROM program WHERE program_id = $pid""".update.run
```
tpolecat: Would it be easier to just have the transactor always roll back?

modules/db/src/test/scala/gem/dao/DaoTest.scala
```diff
@@ -10,17 +10,19 @@ import scalaz.effect.IO
...
-    "postgres",
-    ""
+  private val xa = Transactor.after.set(
+    DriverManagerTransactor[IO](
+      "org.postgresql.Driver",
+      "jdbc:postgresql:gem",
+      "postgres",
+      ""
+    ),
+    HC.rollback
```
swalker2m: @tpolecat does this look correct?  It seems to work.

tpolecat: 👍

PR #53, opened 2017-06-19
update doobie version for build build, switch to coursier for lib resolution

tpolecat:
This does two things
- Updates the build build to use doobie 0.4.2-SNAPSHOT just so we're consistent with the main build.
- Updates the main build and the build build to use coursier for library management, which seems to speed things up. First time building it will have to re-download the internet but thereafter it's much faster.

### DIFF
--- a/project/build.sbt
+++ b/project/build.sbt
@@ -1,8 +1,12 @@
+resolvers in ThisBuild +=
+  "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"
+
+lazy val doobieVersion = "0.4.2-SNAPSHOT"
 
 // The build build, for codegen
 libraryDependencies ++= Seq(
-  "org.tpolecat" %% "doobie-core"               % "0.3.0",
-  "org.tpolecat" %% "doobie-contrib-postgresql" % "0.3.0"
+  "org.tpolecat" %% "doobie-core"     % doobieVersion,
+  "org.tpolecat" %% "doobie-postgres" % doobieVersion
 )
 
 addCompilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -4,6 +4,7 @@ addSbtPlugin("org.flywaydb"     % "flyway-sbt"            % "4.0.3")
 addSbtPlugin("org.scalastyle"  %% "scalastyle-sbt-plugin" % "0.8.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager"   % "1.1.4")
 addSbtPlugin("com.typesafe.sbt" % "sbt-git"               % "0.8.5")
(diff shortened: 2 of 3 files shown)


jdnavarro:
Some time ago I tried `coursier` in `ocs3` - as part of my `ensime` setup - but couldn't get it to work at the the first attempt so I didn't bother trying to fix it, but do you think it's worth switching?

cquiroz:
i'm not too familiar with coursier, it claims to be faster to download and I have seen it work with IntelliJ. Ensime doesn't work form me regardless 😄

jdnavarro:
I just tried it now for `ocs3`, just sbt, no ensime, but I'm getting lots of HTTP 500 errors. It looks as if it we needed to switch to some other artifact repository that coursier supports.

cquiroz:
interesting, perhaps our maven repo is not compatible though a 500 is a bit strange in that case

tpolecat:
Coursier is much much faster than Ivy.

@jdnavarro maybe try again? Bintray was having some issues yesterday.

jdnavarro:
> maybe try again?

Now I'm not getting the 500s but I'm getting a type mismatch error. It's weird because if I deactivate `courier` it compiles fine.

Can `courier` in some way change something about the compiler?

 I'm doing `sbt clean` always before trying.

tpolecat:
Hm, don't know. This is for ocs3?

jdnavarro:
Yeah, it's for ocs3. These are the errors:

```
[error] /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala:304: value toMillimetersOfMercury is not a member of squants.motion.Pressure
[error]           val x = gwsReader.getAirPressure.map(_.map(_.toMillimetersOfMercury))
[error]                                                        ^
[info] Full optimizing /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/target/scala-2.11/resource_managed/main/seqexec-cli-opt.js
[info] Fast optimizing /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/target/scala-2.11/resource_managed/main/seqexec-cli.js
[error] /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala:463: type mismatch;
[error]  found   : scalaz.EitherT[scalaz.concurrent.Task,edu.gemini.seqexec.server.SeqexecFailure,Nothing]
[error]  required: edu.gemini.seqexec.server.SeqAction[Unit]
[error]     (which expands to)  scalaz.EitherT[scalaz.concurrent.Task,edu.gemini.seqexec.server.SeqexecFailure,Unit]
[error]     gwsKeywords
[error]     ^
[error] two errors found
```

tpolecat:
Hm, could be differences in eviction policies. You may be getting older/newer versions of squants and scalaz than what you expect. I'm not sure how to diagnose. Maybe ask in sbt/sbt or in coursier/coursier.

PR #54, opened 2017-06-19
add scalastyle

tpolecat:
This adds [Scalastyle](http://www.scalastyle.org/) which gives some lightweight checking for stylistic things like return type annotations (I had to add a few). It's very fast so I put it first in the CI build. There's a lot of room to improve/extend the default configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_script:
   - psql -c 'create database gem' -U postgres
 
 script:
-  - sbt sql/flywayMigrate compile test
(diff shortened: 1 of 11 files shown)


cquiroz:
Where does that scalastyle definition come from?

jdnavarro:
> Where does that scalastyle definition come from?

I think it's the default coming witch scalastyle. I was using it locally for `ocs3` for the code I used to write in the beginning.

cquiroz:
Ok, I thought there would be some typlevel suggested one

PR #56, opened 2017-06-19
add licence and automatic header generation

tpolecat:
This adds the BSD-3 license (same as in OCS) and the `sbt-header` plugin that will ensure all source files have the right header material. For new files the header will be added when you first compile. Only the first two commits are relevent; the third is just 28746382476 license insertions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,11 @@ cache:
   - $HOME/.sbt/launchers
   - $HOME/.ivy2/cache
   - $HOME/.nvm
+  - $HOME/.coursier
 
 before_cache:
   - du -h -d 1 $HOME/.ivy2/cache
   - du -h -d 2 $HOME/.sbt/
+  - du -h -d 2 $HOME/.coursier/
   - find $HOME/.sbt -name "*.lock" -type f -delete
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_script:
   - psql -c 'create database gem' -U postgres
 
 script:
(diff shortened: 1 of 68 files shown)


tpolecat:
I would be open to a shorter custom header, this is just the default one.

tpolecat:
Hm bintray is crapping out. I'll restart the build again in a bit.

cquiroz:
hmm, would this change the files and add the license to each file? I'm not sure I want that, sounds very Repeat yourself

tpolecat:
@cquiroz I agree but it's pretty standard to have a header at the top of every source file.

cquiroz:
Personally I don't like it, I see little benefit over having it at the top level besides making it longer to find the meat 😄 . If it is a policy requirement, we'll have to do it, but otherwise I'd rather avoid it

tpolecat:
I guess that's up to @arturog8m … it does seem like we need *something* to identify source files but it could just be something like

    // Copyright (c) 2017 Gemini Observator, for licensing see LICENSE.txt

tpolecat:
Yeah Sam Halliday is a license nerd and he just does what I have above, more or less. I'll change it.

tpolecat:
Ok how does the 2-line header look?

cquiroz:
sorry for the hassle, that looks to much better. BTW I think we should apply more or less the same policies across gem and ocs3, including this, scalastyle, etc

cquiroz:
I'm ok too. As a general question: Will the tool add it to the files without it automatically?

tpolecat:
@cquiroz yes, when you compile it adds the header if it's not there (or if it's not the expected header ... we can update it in the build and it will change all the files for us).

PR #57, opened 2017-06-20
WIP: add wartremover

tpolecat:
This adds [WartRemover](http://www.wartremover.org/) and fixes a bunch of minor things that can be the source of problems. Mostly:

- making all case classes final
- only using `toString` on types that have overridden it explicitly
- removing all overloading and default arguments
- adding type annotations to all public members
- adding `@SuppressWarnings` for cases that couldn't be fixed easily

Happy to argue about these.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,6 +53,16 @@ lazy val testLibs = Seq(
   "org.scalacheck" %% "scalacheck" % scalaCheckVersion % "test"
 )
 
+lazy val gemWarts =
+  Warts.allBut(
+    Wart.Any,                // false positives
+    Wart.Nothing,            // false positives
+    Wart.Product,            // false positives
+    Wart.Serializable,       // false positives
+    Wart.Recursion,          // false positives
+    Wart.ImplicitConversion  // we know what we're doing
+  )
+
 lazy val commonSettings = Seq(
 
   // These sbt-header settings can't be set in ThisBuild for some reason
@@ -62,6 +72,10 @@ lazy val commonSettings = Seq(
        |For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
        |""".stripMargin
   )),
+
+  // Wartremover in compile and test (not in Console)
+  wartremoverErrors in (Compile, compile) := gemWarts,
+  wartremoverErrors in (Test,    compile) := gemWarts,
 
   scalaOrganization := "org.typelevel",
   scalaVersion := "2.12.2-bin-typelevel-4",
--- a/modules/core/src/main/scala/gem/Location.scala
+++ b/modules/core/src/main/scala/gem/Location.scala
@@ -26,6 +26,14 @@ sealed trait Location extends Product with Serializable {
     * if any.  Beginning and End Locations have no fixed elements.
     */
   protected def minPrefixLength: Int
+
+  final override def toString: String =
+    this match {
+      case Location.Beginning  => "{α}"
+      case Location.Middle(ps) => "{" + ps.map(_.toString).intercalate(",") + "}"
+      case Location.End        => "{ω}"
+    }
+
 }
 
 object Location {
@@ -43,8 +51,6 @@ object Location {
     protected def minPrefixLength: Int =
       0
 
-    override def toString: String =
-      "{α}"
   }
 
   /** A Location that falls somewhere between the Beginning and End.
@@ -62,8 +68,6 @@ object Location {
     protected def minPrefixLength: Int =
       posList.size
 
-    override def toString =
-      toList.mkString("{", ",", "}")
   }
 
   /** A marker for the last Location.  No other Location comes after the End.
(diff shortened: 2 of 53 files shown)


build.sbt
```diff
@@ -48,7 +48,19 @@ lazy val testLibs = Seq(
   "org.scalacheck" %% "scalacheck" % scalaCheckVersion % "test"
 )
 
+lazy val gemWarts =
+  Warts.allBut(
+    Wart.Any,                // false positives
+    Wart.Nothing,            // false positives
+    Wart.Product,            // false positives
+    Wart.Serializable,       // false positives
+    Wart.Recursion,          // false positives
```
jdnavarro: Are there many recursion false positives ? Can `@tailrec` be used in most cases?

cquiroz:
I think I'm ok with these. Is there one to forbid unused imports?

modules/ctl/src/main/scala/free/interp.scala
```diff
@@ -60,6 +60,7 @@ object interpreter {
    * Construct a program to log a message to the console at the given log level and indentation.
    * This is where all the colorizing happens.
    */
+  @SuppressWarnings(Array("org.wartremover.warts.ToString"))
```
cquiroz: What's this?

cquiroz:
@swalker2m I have to agree with you. I don't see much benefit of overriding `toString` on case classes

swalker2m:
> Is there one to forbid unused imports?

We have a compile option for this (`-Ywarn-unused:imports`), though it seems to be turned off at the moment?

```scala
  scalacOptions in (Compile, console) ~= (_.filterNot(Set(
    "-Xfatal-warnings",
    "-Ywarn-unused:imports"
  ))),
```
I kind of like it on because otherwise the import list grows and grows over time.  I try and remove warnings every time I see them too so I would be inclined to turn on `-Xfatal-warnings` as well.  You end up with the `ocs` project and it's 10,000,000 ignored warnings if you don't keep on top of them!

cquiroz:
We have `fatal-warnings` enabled on ocs3. I think I'd add the unused imports too 😄

build.sbt
tpolecat: The problem is that monadic recursion (which is stacksafe via Free or whatever) can't be tail-recursive so it sets off this warning.

modules/ctl/src/main/scala/free/interp.scala
tpolecat: Wartremover triggers a compile error if you use `.toString` on any type that doesn't override the default implementation. So we need to change `ProgramId` to override `toString` rather than overriding it in all of the case classes. This is kind of irritating but I think the wart is reasonable.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -23,6 +23,14 @@ sealed trait Location extends Product with Serializable {
     * if any.  Beginning and End Locations have no fixed elements.
     */
   protected def minPrefixLength: Int
+
+  final override def toString: String =
+    this match {
+      case Location.Beginning  => "{α}"
+      case Location.Middle(ps) => "{" + ps.map(_.toString).intercalate(",") + "}"
+      case Location.End        => "{ω}"
+    }
```
tpolecat: This fixes the `ToString` wart because now `Location` implements `.toString` directly. In principle it should be smart enough to say well it's sealed and every subclass implements `toString` but It can't do that. This seems fine to me though.

modules/ctl/src/main/scala/free/interp.scala
```diff
@@ -104,7 +105,7 @@ object interpreter {
   private def doShell(cmd: String \/ List[String], verbose: Boolean, state: IORef[InterpreterState]): EitherT[IO, Int, Output] = {
 
     def handler(s: String): IO[Unit] =
-      if (verbose) doLog(Level.Shell, s, state).run.map(_.toOption.get) // shh
+      if (verbose) doLog(Level.Shell, s, state).run.void
```
tpolecat: It was basically an assertion that `doLog` can never fail. I could express this in the type but I think it's ok as-is.

cquiroz: Could you explain why the type tells that `doLog` can never fail?

tpolecat:
>  it seems to be turned off at the moment?

It's turned off in `console` but it's on in `compile` ... you get warnings for unused imports.

I don't have `-Xfatal-warnings` on in either place because we have some partial matches that would need a catch-all that throws an exception … I kind of prefer to have the warning but I could go either way.

modules/ctl/src/main/scala/free/interp.scala
tpolecat: > Could you explain why the type tells that doLog can never fail?

It doesn't. I'll fix it.

tpolecat:
@cquiroz I fixed the `doLog` thing in `interp.scala` by factoring out the `IO[Unit]` portion.

swalker2m:
> I don't have -Xfatal-warnings on in either place because we have some partial matches that would need a catch-all that throws an exception 

I believe I removed those.  It comes for me without warnings.

PR #59, opened 2017-06-23
WIP: exact angles

tpolecat:
This introduces `Angle` for exact angles, and the subclass `HourAngle` for exact hour angles (see below).
- Angles are represented as a single `Int` representing milliarcseconds modulo 360°, and can thus be added, subtracted, negated, flipped, and compared exactly.
- Hour angles are the subset of angles whose milliarcsecond values are evenly divisible by 15; that is, everywhere a milliarcsecond value corresponds with a whole number millisecond. 
- `Angle` forms an Abelian group; `HourAngle` forms a subgroup thereof. Normally I don't like subtyping but this case seems to fit perfectly.

More work to do but I'd like to offer this for discussion.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/src/main/scala/gem/math/Angle.scala
@@ -1,0 +1,262 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.math
+
+import scalaz.Monoid
+
+/**
+ * Exact angles represented as integral milliarcseconds. These values form an Abelian group over
+ * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
+ * integral milliarcseconds correspond with clock milliseconds (i.e., where they are evenly
+ * divisible by 15 milliarcseconds) is represented by the HourAgle subtype.
+ * @param toMilliarcseconds This angle in milliarcseconds. Exact.
+ */
+sealed class Angle (val toMilliarcseconds: Int) {
+
+  // Sanity checks … should be correct via the companion constructor.
+  assert(toMilliarcseconds >= 0, s"Invariant violated. $toMilliarcseconds is negative.")
+  assert(toMilliarcseconds < 360 * 60 * 60 * 1000, s"Invariant violated. $toMilliarcseconds is >= 360°.")
+
+  /** Flip this angle 180°. Exact, invertible. */
+  def flip: Angle =
+    this + Angle.Angle180
+
+  /** Additive inverse of this angle (by mirroring around the 0-180 axis). Exact, invertible. */
+  def unary_- : Angle =
+    Angle.fromMilliarcseconds(-toMilliarcseconds.toLong)
+
+  /** This angle in decimal degrees. Approximate, non-invertible */
+  def toDoubleDegrees: Double =
+    toMilliarcseconds.toDouble / (60.0 * 60.0 * 1000.0)
+
+  /** This angle in decimal radisns. Approximate, non-invertible */
+  def toDoubleRadians: Double =
+    toDoubleDegrees.toRadians
+
+  /**
+   * Convert to the closest hour angle by roundinng down to the closest 15 milliarcseconds.
+   * Exact, non-invertible.
+   */
+  def toHourAngle: HourAngle =
+    HourAngle.fromMilliseconds(toMilliarcseconds.toLong / 15L)
+
+  /**
+   * Convert to the closest hour angle iff its magnitide is an even multiple of 15 milliarcseconds.
+   * Exact and invertible where defined.
+   */
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
+  def toHourAngleExact: Option[HourAngle] =
+    if (toMilliarcseconds % 15 == 0) Some(toHourAngle) else None
+
+  /**
+   * Destructure this value into a sum of degrees, arcminutes, arcseconds, and milliarcseconds.
+   * Exact, invertible via `Angle.fromDMS`.
+   */
+  def toDMS: Angle.DMS =
+    Angle.DMS(this)
+
+  /** Sum of this angle and `a`. Exact, commutative, invertible. */
+  def +(a: Angle): Angle =
+    Angle.fromMilliarcseconds(toMilliarcseconds.toLong + a.toMilliarcseconds.toLong)
+
+  /** Difference of this angle and `a`. Exact, invertible. */
+  def -(a: Angle): Angle =
+    Angle.fromMilliarcseconds(toMilliarcseconds.toLong - a.toMilliarcseconds.toLong)
+
+  /** String representation of this Angle, for debugging purposes only. */
+  override def toString =
+    f"Angle($toDMS, $toDoubleDegrees°)"
+
+  /** Angles are equal if their magnitudes are equal. Exact. */
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
+  override final def equals(a: Any) =
+    a match {
+      case a: Angle => a.toMilliarcseconds == toMilliarcseconds
+      case _        => false
+    }
+
+  override final def hashCode =
+    toMilliarcseconds
+
+}
+
+object Angle {
+
+  val Angle0:   Angle = fromDegrees(0L)
+  val Angle90:  Angle = fromDegrees(90L)
+  val Angle180: Angle = fromDegrees(180L)
+
+  /** Construct a new Angle of the given magnitide in integral milliarcseconds, modulo 360°. Exact. */
+  def fromMilliarcseconds(mas: Long): Angle = {
+    val masPer360 = 360 * 60 * 60 * 1000
+    val masʹ = (((mas % masPer360) + masPer360) % masPer360)
+    new Angle(masʹ.toInt)
+  }
+
+  /** Construct a new Angle of the given magnitide in integral arcseconds, modulo 360. Exact. */
+  def fromArcseconds(as: Long): Angle = fromMilliarcseconds(as * 1000)
+
+  /** Construct a new Angle of the given magnitide in integral arcminutes, modulo 360. Exact. */
+  def fromArcminutes(ms: Long): Angle = fromArcseconds(ms * 60)
+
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * Convert to the closest hour angle by rounding down to the closest 15 milliarcseconds.
+   * Exact, non-invertible.
+   */
+  def toHourAngle: HourAngle =
+    HourAngle.fromMilliseconds(toMilliarcseconds.toLong / 15L)
+
+  /**
+   * Convert to the closest hour angle iff its magnitide is an even multiple of 15 milliarcseconds.
```
cquiroz: s/iff/if s/magnitide/magnitude/

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+   * Exact and invertible where defined.
+   */
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
```
cquiroz: Could you remind me why are we suppressing here?

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+ * Exact angles represented as integral milliarcseconds. These values form an Abelian group over
+ * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
+ * integral milliarcseconds correspond with clock milliseconds (i.e., where they are evenly
+ * divisible by 15 milliarcseconds) is represented by the HourAgle subtype.
+ * @param toMilliarcseconds This angle in milliarcseconds. Exact.
+ */
+sealed class Angle (val toMilliarcseconds: Int) {
+
+  // Sanity checks … should be correct via the companion constructor.
+  assert(toMilliarcseconds >= 0, s"Invariant violated. $toMilliarcseconds is negative.")
```
cquiroz: Should these be instead done on the `fromXXX` method?

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+    val d  = (millis / (1000 * 60 * 60))
+    (d, m, s, ms)
+  }
+
+  /**
+   * Integral angle represented as a sum of degrees, arcminutes, arcseconds, and milliarcseconds.
+   * This type is exact and isomorphic to Angle.
+   */
+  final case class DMS(toAngle: Angle) {
+    val (
```
cquiroz: nice

cquiroz:
Why do we need a new `Angle` class if we had one on the new target model in ocs?

modules/core/src/main/scala/gem/math/Angle.scala
sraaphorst: iff = if and only if. Implication in both directions.

sraaphorst: But magnitude 👍

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+    Monoid.instance(_ + _, Angle.Angle0)
+
+  // This works for both DMS and HMS so let's just do it once.
+  protected[math] def toMillisexigesimal(millis: Int): (Int, Int, Int, Int) = {
+    val ms =  millis                     % 1000
+    val s  = (millis / (1000))           % 60
+    val m  = (millis / (1000 * 60))      % 60
+    val d  = (millis / (1000 * 60 * 60))
+    (d, m, s, ms)
+  }
```
sraaphorst: I know that `Duration`s and `Angle`s are related but distinct, but do we want to use constants like this?

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: I'll switch it to use `===`.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: The `fromXXX` methods guarantee these but I wanted a double-check while developing since I have been changing things so much. When we're sure it's right we can turn these off.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: What are you suggesting? We could factor out the constants but I'm not sure it would help.

tpolecat:
> Why do we need a new Angle class if we had one on the new target model in ocs?

The intent would be to build this out so that it would be a drop-in replacement for the current `Angle` class, more or less, to provide for exact comparisons (and exact computations where possible) to try to avoid some of the "close enough" nonsense we had to deal with in OCS.

cquiroz:
Sounds good. There is no normalizint to 360 it seems?

tpolecat:
> There is no normalizint to 360 it seems?

There is. See `Angle.fromMilliarcseconds` which is the only way to construct an `Angle`. It normalized to 360° in mas. Similarly `HourAngle` normalizes to 24h in ms.

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+  def fromDoubleDegrees(ds: Double): Angle =
+    fromMilliarcseconds((ds * 60 * 60 * 1000).toLong)
+
+  /** Construct a new Angle of the given magnitide in radians, modulo 2π. Approximate. */
+  def fromDoubleRadians(rad: Double): Angle =
+    fromDoubleDegrees(rad.toDegrees)
+
+  /** Angle is an Abelian group, but monoid is the best we can do for now. */
+  implicit val AngleMonoid: Monoid[Angle] =
+    Monoid.instance(_ + _, Angle.Angle0)
```
jdnavarro: Aren't there other ways an `Angle` can be a monoid? What about multiplication? I'm not sure what's the usual way to represent different type class instances for the same type in Scala.

sraaphorst: How would you multiply two angles? It seems you are possibly thinking of them as a module over the real numbers?

jdnavarro: I guess they can't be multiplied but I'm asking because I never worked with angles before and don't know what is usually done with them. Is summing angles the only monoidal operation you would do with 2 angles?

cquiroz: I guess this is more general than angles, but in a Monoid of Int what happens if you go over the Int range? Like
```
Int.MaxInt |+| Int.MaxInt
```

It probably doesn't apply to this case since we are restricting the range of the angles

jdnavarro: I don't know what happens in Scala, but in Haskell `Int` is not a Monoid instance by itself. You would have to wrap it with a `Sum` or `Product` to use it as a Monoid, which would be depending on the wrapper.

jdnavarro: Never mind about my question. I just read a bit on angles as monoids and the sum operation for `mappend` and the 0 angle for `mempty` are the most obvious choices for the monoidal instance of `Angle`.

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -0,0 +1,262 @@
...
+import scalaz.Monoid
+
+/**
+ * Exact angles represented as integral milliarcseconds. These values form an Abelian group over
+ * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
+ * integral milliarcseconds correspond with clock milliseconds (i.e., where they are evenly
+ * divisible by 15 milliarcseconds) is represented by the HourAgle subtype.
+ * @param toMilliarcseconds This angle in milliarcseconds. Exact.
+ */
+sealed class Angle (val toMilliarcseconds: Int) {
```
swalker2m: Shouldn't the constructor be private?

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: @cquiroz modular addition forms a monoid (an Abelian group in fact) so it's lawful for `Int` and `Long`, and for these types as well.

@jdnavarro multiplication by a scalar is an operation we need to support, and it can be exact for integral factors. I don't know what it would mean to multiply two angles together because degrees^2 doesn't make immediate sense to me. Regarding canonical monoids in cats/scalaz we usually pick addition with the natural zero when we can identify one, but as you say it's arbitrary. Boolean has no canonical monoid because there's no obvious way to pick one.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: Yes, thank you.

modules/core/src/main/scala/gem/math/Angle.scala
cquiroz: Could you expand a bit about the `Abelian` group. I sort of remember those from the advanced algebra course but not that well.

Wouldn't overflow break it?

tpolecat:
Ok I added some more tests and cleaned up a bit.

modules/core/src/main/scala/gem/math/Angle.scala
sraaphorst: Yes, an `Abelian` group is a commutative group, i.e. `a + b = b + a` for all `a, b` in the group. Integers / reals / rationals (which are rings) acting on an abelian group make it into a module (generalization of a vector space) over the ring.

The operation of multiplying two degrees together makes no sense, so `Angle` cannot be made a ring.

tpolecat: And backing up one step, a group is a monoid where every element has an inverse … for any `a` you can find a `b` such that `a |+| b = zero`. Then if `|+|` is commutative it's Abelian, as Sebastian explained. This [table](https://en.wikipedia.org/wiki/Template:Group-like_structures) is incomplete but might be helpful.

Overflow doesn't break things like `Monoid[Int]`. It's still legal modular arithmetic ... the wraparound point doesn't have to be at zero. In the case of angles the values are constrained enough that they can't overflow.

PR #60, opened 2017-06-28
Add GMOS, Part B

swalker2m:
This is the second of a multi-part project to add GMOS (north and south) to the model.  The two GMOSes are very similar and yet frustratingly different.  This leads to a bit of duplication, at least as implemented right now.  Suggestions on a reasonable way to de-duplicate are certainly welcome.

At the moment the majority of GMOS properties are included but it is still missing some things like nod and shuffle and ROIs.  Those will be added in subsequent PRs.

I'm proposing some changes to the current GMOS model.  In particular instead of having explicit "None" values for disperser, filter, etc., the model makes greater use of `Option`.   If we decide that this is a better way to go then I will need to go back and make Flamingos2 consistent.


modules/core/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -67,3 +67,39 @@ final case class F2DynamicConfig(
...
+final case class GmosNorthDynamicConfig(
+  common:  GmosCommonDynamicConfig,
+  grating: Option[GmosGrating[GmosNorthDisperser]],
+  filter:  Option[GmosNorthFilter],
+  fpu:     Option[GmosCustomMask \/ GmosNorthFpu],
+) extends DynamicConfig {
+
+  type I = Instrument.GmosN.type
+  def instrument: I = valueOf[I]
+}
```
tpolecat: Does this not work as `extends DynamicConfig.Impl(Instrument.GmosN)`?

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  final case class GmosNodAndShuffle(
+    /*placeholder for now*/
+  )
+
+  /** Shared static configuration for both GMOS-N and GMOS-S.
+    */
+  final case class GmosCommonStaticConfig(
+    detector:      GmosDetector,
+    mosPreImaging: Boolean,
```
tpolecat: Can we make an ADT for this? `Boolean` makes me nervous.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,95 @@
...
+        false,
+        None
+      )
+  }
+
+  /** Parameters that determine GMOS CCD readout.
+    */
+  final case class GmosCcdReadout(
+    xBinning:    GmosBinning,
+    yBinning:    GmosBinning,
```
tpolecat: Would there be any benefit to distinguishing X and Y binning?

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,95 @@
...
+        Duration.ofSeconds(300)
+      )
+  }
+
+  /** Custom mask definition, which is available as an alternative to using a
+    * builtin FPU.  Either both these parameters are set or neither are set in a
+    * GMOS observation
+    */
+  final case class GmosCustomMask(
+    maskDefinitionFilename: String,
```
tpolecat: It bothers me that this is a free text field. But I guess that's the way it actually works.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,95 @@
...
+    slitWidth:              GmosCustomSlitWidth
+  )
+
+  /** GMOS grating central wavelength.  For now, just a value class wrapper
+    * around a Double.  This should be switched to Fixed / squants?
+    */
+  final case class GmosCentralWavelength(val nm: BigDecimal) extends AnyVal
```
tpolecat: Comment is incorrect. `BigDecimal` seems good. We should look at [Libra](https://to-ithaca.github.io/libra/) for dimensions … it may end up being better than Squants.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  /** GMOS grating configuration, parameterized on the disperser type.  These
+    * are grouped because they only apply when using a grating.  That is, all
+    * are defined or none or defined in the dynamic config.
+    */
+  final case class GmosGrating[D](
```
tpolecat: I just learned you can say `* @tparam D comment for type param D` for scaladoc comments, so we might want to start adding those to doc comments.

modules/core/src/main/scala/gem/config/StaticConfig.scala
```diff
@@ -27,15 +27,48 @@ final case class NiciStaticConfig()       extends StaticConfig.Impl(Instrument.N
...
-final case class GmosNStaticConfig()      extends StaticConfig.Impl(Instrument.GmosN)
 final case class BhrosStaticConfig()      extends StaticConfig.Impl(Instrument.Bhros)
 final case class VisitorStaticConfig()    extends StaticConfig.Impl(Instrument.Visitor)
 
 final case class F2StaticConfig(mosPreImaging: Boolean) extends StaticConfig {
   type I = Instrument.Flamingos2.type
 
-  def instrument: I = Instrument.Flamingos2
+  def instrument: I = valueOf[I]
+}
```
tpolecat: Same comment as above re `.Impl(const)`, and for classes below.

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -74,6 +76,32 @@ object StaticConfigDao {
...
+        (s: GmosCommonSC)            => (s.detector, s.mosPreImaging)
+      )
+
+    def selectGmosNorth(sid: Int): Query0[GmosNorthStaticConfig] =
+      sql"""
+        SELECT detector,
+               mos_preimaging,
+               stage_mode
+          FROM static_gmos_north
+         WHERE static_id = $sid AND instrument = ${Instrument.GmosN: Instrument}
```
tpolecat: The widening ascriptions here are irritating, sorry.

tpolecat:
Scalastyle doesn't like something. Maybe whitespace around parens or brackets. I ran into that somewhere else. I'm fine turning Scalastyle off if it's not helping. The 2.12.2 warnings plus wartremover are probably sufficient.

swalker2m:
> `The command "sbt headerCheck test:headerCheck scalastyle sql/flywayMigrate compile test" exited with 1.`

Oops, I didn't realize I needed to do the style check explicitly.  I just saw it compiled without warnings or errors and thought it was good.

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -215,4 +215,277 @@ object Parsers {
...
+
+    val builtinRoi: PioParse[Option[GmosBuiltinRoi]] = enum(
+      "Custom ROI"         -> Option.empty[GmosBuiltinRoi],
+      "Full Frame Readout" -> Some(FullFrame      ),
+      "CCD 2"              -> Some(Ccd2           ),
+      "Central Spectrum"   -> Some(CentralSpectrum),
+      "Central Stamp"      -> Some(CentralStamp   ),
+      "Top Spectrum"       -> Some(TopSpectrum    ),
+      "Bottom Spectrum"    -> Some(BottomSpectrum ),
+    )
```
swalker2m: Interesting, this seems like it shouldn't compile but it does.  It does trigger the scalastyle error though.

swalker2m:
I think I addressed all the feedback, except the mask MDF file name. I think there might be some suggested structure to those names (?), but we allow anything in that field today.  We might revisit this at some point but I'm not sure what else to do about it at the moment.

The scalastyle checks actually caught things I would have expected to be compile problems (though they weren't) so it is probably worth keeping.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -0,0 +1,98 @@
...
+    */
+  final case class GmosCustomMask(
+    maskDefinitionFilename: String,
+    slitWidth:              GmosCustomSlitWidth
+  )
+
+  /** GMOS grating central wavelength.  For now, just a value class wrapper
+    * around a BigDecimal.  This should be switched to Fixed / squants?
+    */
+  final case class GmosCentralWavelength(val nm: BigDecimal) extends AnyVal
```
cquiroz: `Wavelength` is available in `squants` and we used it for this very purpose on ocs3

modules/core/src/test/scala/gem/config/Arbitraries.scala
```diff
@@ -40,57 +40,77 @@ trait Arbitraries extends gem.enum.Arbitraries {
...
-
-  implicit val arbF2Static       =
-    Arbitrary(arbitrary(arbBool).map(F2StaticConfig(_)))
+  implicit val arbMosPreImaging: Arbitrary[MosPreImaging] =
+    Arbitrary(
+      Gen.oneOf(MosPreImaging.IsMosPreImaging,
+                MosPreImaging.IsNotMosPreImaging)
+    )
+
+  implicit val arbAcqCamStatic    = const(AcqCamStaticConfig()         )
```
cquiroz: Why do we need an `Arbitrary` `const`? Doesn't it defeat the purpose. or is it to combine it with other Arbitraries?

modules/core/src/test/scala/gem/enum/Arbitraries.scala
```diff
@@ -23,6 +23,29 @@ trait Arbitraries {
   implicit val arbF2WindowCover = pick(F2WindowCover.all)
 
 
+  // Gmos
+
+  implicit val arbGmosAdc             = pick(GmosAdc.all            )
```
cquiroz: What's `pick`?

modules/core/src/test/scala/gem/config/Arbitraries.scala
tpolecat: Placeholders. These data types will get built out.

swalker2m: Right, these are placeholders.  Having an arbitrary `AcqCamStaticConfig` instance available allows us to fill out the match in `genStaticConfigOf(i: Instrument)` and avoid a catch-all case or a missing case warning:

```scala
  def genStaticConfigOf(i: Instrument): Gen[StaticConfig] =
    i match {
      case AcqCam     => arbitrary[AcqCamStaticConfig   ].widen[StaticConfig]
      ...
    }
```

modules/core/src/test/scala/gem/enum/Arbitraries.scala
swalker2m: I just wanted a shortcut to avoid repeating `Arbitrary(oneOf(...))` multiple times.  Maybe there's a better way though.

```scala
  private def pick[A](as: List[A]): Arbitrary[A] =
    Arbitrary(oneOf(as))
```

swalker2m:
Rebased.

modules/core/src/test/scala/gem/config/Arbitraries.scala
cquiroz: Makes sense

cquiroz: Why do we need `widen` given than `AcqCamStaticConfig` is a `StaticConfig`? Is it due to limitations on the type inference?

swalker2m: Thanks, I must have thought `Gen` was invariant or something.  It turns out it isn't needed though.

swalker2m:
I think I'm ready to merge this now and then come back later to add nod and shuffle, ROIs, and smart gcal in separate smaller PRs.

PR #61, opened 2017-06-30
Remove step foreign key from gcal table

swalker2m:
This PR is half a request for ideas and half proposal. I noticed that after importing programs via `ocs2/run` suddenly the Smart Gcal tests would stop working. The issue is that the `smart_f2` table is truncated by the cleanup performed by the `FileImporter` before reading programs. 

```scala
val clean: ConnectionIO[Unit] =
  for {
    _ <- sql"TRUNCATE program CASCADE".update.run
    ...
  } yield ()
```

In particular `TRUNCATE program CASCADE` ultimately truncates the `smart_f2` table needed by F2 Smart Gcal.  This is because `smart_f2` has a foreign key to the `gcal` table, which in turn has a foreign key to `step` which references `observation` which references `program`. 

One way to fix this would be to delete all rows from program with `DELETE FROM program` instead of using `TRUNCATE`.  That would take advantage of normal delete cascading which would cull associated `gcal` table rows but not remove those required by `smart_f2`.  Unfortunately it is very slow, even for a database with just a single semester of queue programs loaded.  Maybe that's okay though because we can always use `flywayClean` before importing?

Another option is to remove the foreign key reference from `gcal` to `step` and thus stop the cascading truncates.  The foreign key reference was only added to make automatic cleanup of `gcal` rows happen when the corresponding `step` row is deleted.  I've taken that approach in this PR.  Now a trigger is required to clean `gcal` rows when the associated step is removed.

Also, I noticed that `static_config`, which has no foreign key references to observation is not touched by the `TRUNCATE program CASCADE` cleanup, so I explicitly listed it in `FileImporter.clean`

```scala
  val clean: ConnectionIO[Unit] =
    for {
      _ <- sql"DELETE FROM gcal USING step_gcal WHERE gcal.gcal_id = step_gcal.gcal_id".update.run
      _ <- sql"TRUNCATE program, static_config CASCADE".update.run
      _ <- sql"TRUNCATE log".update.run
      _ <- sql"DELETE FROM semester".update.run
    } yield ()
```

### DIFF
--- a/modules/sql/src/main/resources/db/migration/V015__Remove_Step_FK_From_Gcal.sql
+++ b/modules/sql/src/main/resources/db/migration/V015__Remove_Step_FK_From_Gcal.sql
@@ -1,0 +1,21 @@
+--
+-- Remove step_id from gcal.  step_gcal is used to tie gcal to step.  The
+-- step_id was added to make deletion of step_gcal cascade through to the
+-- gcal table.  When truncating the program table though, the truncate
+-- unfortunately trickles down to gcal which also contains gcal configs
+-- for smart_gcal.
+--
+
+ALTER TABLE gcal
+  DROP COLUMN step_id CASCADE;
+
+-- Delete the corresponding gcal row when a step_gcal row is removed.
+CREATE FUNCTION delete_gcal() RETURNS trigger AS $delete_gcal$
+  BEGIN
+    DELETE FROM gcal WHERE gcal_id = OLD.gcal_id;
+    RETURN OLD;
+  END;
+$delete_gcal$ LANGUAGE plpgsql;
+
+CREATE TRIGGER delete_gcal AFTER DELETE ON step_gcal
+  FOR EACH ROW EXECUTE PROCEDURE delete_gcal();
--- a/modules/db/src/main/scala/gem/dao/GcalDao.scala
+++ b/modules/db/src/main/scala/gem/dao/GcalDao.scala
@@ -13,8 +13,8 @@ import java.time.Duration
 
 object GcalDao {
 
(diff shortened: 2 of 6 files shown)


modules/sql/src/main/resources/db/migration/V015__Remove_Step_FK_From_Gcal.sql
```diff
@@ -0,0 +1,21 @@
...
+
+-- Delete the corresponding gcal row when a step_gcal row is removed.
+CREATE FUNCTION delete_gcal() RETURNS trigger AS $delete_gcal$
+  BEGIN
+    DELETE FROM gcal WHERE gcal_id = OLD.gcal_id;
+    RETURN OLD;
+  END;
+$delete_gcal$ LANGUAGE plpgsql;
+
+CREATE TRIGGER delete_gcal AFTER DELETE ON step_gcal
```
cquiroz: I haven't seen a TRIGGER in a long time

tpolecat:
I think it might make more sense now to have the file importer delegate to Flyway programmatically to drop and re-create the database, rather than trying to figure out a manual way to get the database back to its initial state. The telnet server ensures the database is up to date when it starts up (see [here](https://github.com/gemini-hlsw/gem/blob/master/modules/telnetd/src/main/scala/main.scala#L31-L37)) … the file importer could do the same thing, but actually wipe the database out first. The point of the tool as I see it is a way to fill the db with test data so this seems ok.

swalker2m:
> I think it might make more sense now to have the file importer delegate to Flyway programmatically to drop and re-create the database

Makes sense to me.  I'll give that a go.

PR #62, opened 2017-07-03
Use Flyway clean/migrate before import

swalker2m:
Following the suggestion from @tpolecat in #61, this PR switches to use Flyway for database cleaning before a file imports.

### DIFF
--- a/modules/ocs2/src/main/scala/gem/ocs2/DoobieClient.scala
+++ b/modules/ocs2/src/main/scala/gem/ocs2/DoobieClient.scala
@@ -16,18 +16,16 @@ import scalaz.effect.IO
 /** Shared support for import applications using Doobie. */
 trait DoobieClient {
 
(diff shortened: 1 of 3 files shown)


modules/ocs2/src/main/scala/gem/ocs2/DoobieClient.scala
```diff
@@ -16,18 +16,16 @@ import scalaz.effect.IO
 /** Shared support for import applications using Doobie. */
 trait DoobieClient {
 
+  val Url  = "jdbc:postgresql:gem"
+  val User = "postgres"
+  val Pass = ""
```
cquiroz: no password?

tpolecat: We should change everything to use a `gem` user but I don't think it's important right now.

PR #65, opened 2017-07-03
move ctl module to cats

tpolecat:
This is the first PR in a series to move gem to cats. Since the `ctl` module is a standalone thing it can be done independently. Notable changes:

- Replace `optparse-applicative` with `decline`, which uses cats and is also much simpler.
- `whenM` and `unlessM` don't exist in cats 0.9.0 but are merged and will be in the next version as `whenA` and `unlessA`. For now they're rewritten in terms of `if`.
- `IO.putStrLn`, etc., aren't provided so these are done as `IO(Console.println(...))` etc.
- `IORef` isn't provided so a simple version is included.
- `a.point[F]` is now `a.pure[F]` everywhere.
- All uses of `\/` are now `Either`, which is right-biased now.

Nothing major.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,6 +13,10 @@ lazy val http4sVersion            = "0.15.2a"
 lazy val scalaXmlVerson           = "1.0.6"
 lazy val scalaParsersVersion      = "1.0.4"
 lazy val tucoVersion              = "0.1.1"
+lazy val catsVersion              = "0.9.0"
+lazy val catsEffectVersion        = "0.3"
+lazy val declineVersion           = "0.2.2"
+lazy val mouseVersion             = "0.9"
 
 enablePlugins(GitVersioning)
 
@@ -255,11 +259,12 @@ lazy val ctl = project
   .enablePlugins(AutomateHeaderPlugin)
   .settings(commonSettings)
   .settings (
-    resolvers += "bmjames Bintray Repo" at "https://dl.bintray.com/bmjames/maven",
+    resolvers += Resolver.bintrayRepo("bkirwi", "maven"),
     libraryDependencies ++= Seq(
-      "org.scalaz"  %% "scalaz-core"   % scalazVersion,
-      "org.scalaz"  %% "scalaz-effect" % scalazVersion,
-      "net.bmjames" %% "scala-optparse-applicative" % "0.5"
+      "org.typelevel"           %% "cats"        % catsVersion,
+      "org.typelevel"           %% "cats-effect" % catsEffectVersion,
+      "com.monovore"            %% "decline"     % declineVersion,
+      "com.github.benhutchison" %% "mouse"       % mouseVersion
     ),
     addCommandAlias("gemctl", "ctl/runMain gem.ctl.main")
   )
--- a/modules/ctl/src/main/scala/command.scala
+++ b/modules/ctl/src/main/scala/command.scala
@@ -37,7 +37,7 @@ object Command {
   def parse[A](progName: String, args: List[String]): IO[Option[Command]] =
     mainParser(progName).parse(args) match {
       case Right(c) => c.some.pure[IO]
-      case Left(h) => IO(Console.println(Console.BLUE + h.toString + Console.RESET)).as(none[Command])
+      case Left(h) => IO(Console.println(Console.BLUE + h.toString + Console.RESET)).as(none[Command]) // scalastyle:ignore
     }
 
   // Opts implementation below
--- a/modules/ctl/src/main/scala/command.scala
+++ b/modules/ctl/src/main/scala/command.scala
@@ -39,21 +39,6 @@ object Command {
       case Right(c) => c.some.pure[IO]
       case Left(h) => IO(Console.println(Console.BLUE + h.toString + Console.RESET)).as(none[Command])
     }
-    // execParserPure(prefs(idm[PrefsMod]), mainParser, args) match {
-    //   case Success(c) => IO(Some(c))
-    //   case Failure(f) =>
-    //     import Predef._
-    //     val (msg, exit) = renderFailure(f, progName)
-    //     IO.putStr(Console.BLUE) *>
-    //     IO.putStrLn(msg) *>
-    //     IO.putStrLn("")  *>
-    //     IO.putStrLn(s"""
-    //       |Hints:
-    //       |  $progName --help            To see available commands.
-    //       |  $progNage COMMAND --help    To see help on a specific command.
-    //      """.trim.stripMargin) *>
-    //      IO.putStr(Console.RESET) as None
-    // }
 
   // Opts implementation below
 
--- a/modules/ctl/src/main/scala/command.scala
+++ b/modules/ctl/src/main/scala/command.scala
@@ -3,11 +3,9 @@
 
 package gem.ctl
 
-import net.bmjames.opts._
-import net.bmjames.opts.types.{ Success, Failure }
-import net.bmjames.opts.builder.internal.{ Mod, CommandFields }
+import com.monovore.decline.{ Command => Cmd, _ }
 
-import scalaz._, Scalaz._, scalaz.effect._
+import cats.implicits._, cats.effect.IO
 
 import gem.ctl.free.ctl.{ Server, Host }
 import gem.ctl.free.interpreter.Config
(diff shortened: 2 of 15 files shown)


cquiroz:
This is a very big change, from scalaz to cats. I hope we can discuss it a bit before committing, especially with the upcoming merge of gem and ocs3

tpolecat:
What are your concerns? I don't think we have any choice since all the infrastructure libs are committing, and it's never going to get any easier. Is there a good reason to wait until after ocs3 and gem are merged?

tpolecat:
Got core ported also. Works.

modules/ctl/src/main/scala/command.scala
```diff
@@ -37,116 +35,119 @@ object Command {
...
+  private lazy val host: Opts[Option[String]] =
+    Opts.option[String](
+      short   = "H",
+      long    = "host",
+      metavar = "HOST",
+      help    = "Use the specified docker host (machine when given with -m)."
+    ).orNone
+
+  private lazy val server: Opts[Server] =
+    (machine |@| host |@| user) map {
```
jdnavarro: Maybe not the place to ask because here the applicative syntax fits nicely, but now that I see this, do you know if there is anything to make for-comprehesions desugar to applicatives when possible?

swalker2m:
This looks good to me, pending @cquiroz  input.  We’re trying to get the two projects (gem and ocs3) together and ultimately switch the seqexec over to work with the new sequence model.  Necessarily it’s a bit of a moving target since we’re also trying to make progress on the new model and have to experiment, etc.  After cats I know there is a doobie update and some work on improving the queries to remove duplication but I believe we should also prioritize merging the two projects.  That may mean pausing on some larger updates until ocs3 can catch up a bit.  After discussing it with @jluhrs and @jdnavarro, I think they will be scheduling the update to cats soon.

modules/ctl/src/main/scala/hi/deploy.scala
```diff
@@ -9,8 +9,9 @@ import gem.ctl.low.git._
 import gem.ctl.low.docker._
 import gem.ctl.hi.common._
 
-import scalaz._, Scalaz._
+import cats.implicits._
 import scala.util.matching.Regex
+import mouse.all._
```
jdnavarro: What is this?

modules/ctl/src/main/scala/hi/deploy.scala
```diff
@@ -37,7 +42,7 @@ object deploy {
         c <- info(s"Using $rev.") *> commitForRevision(rev)
         _ <- info(s"Commit is ${c.hash}")
         u <- uncommittedChanges.map(_ && rev === "HEAD")
-        _ <- u.whenM(warn("There are uncommitted changes. This is a UNCOMMITTED deployment."))
+        _ <- if (u) warn("There are uncommitted changes. This is a UNCOMMITTED deployment.") else ().pure[CtlIO]
```
jdnavarro: What about `whenA`?

modules/ctl/src/main/scala/ioref.scala
```diff
@@ -0,0 +1,17 @@
...
+// N.B. this isn't in cats-effect yet
+@SuppressWarnings(Array("org.wartremover.warts.Var"))
+final class IORef[A] private (var value: A) {
+  def mod(f: A => A): IO[Unit] = IO(value = f(value))
+  def read: IO[A] = IO(value)
+}
+object IORef {
+  def apply[A](a: A): IO[IORef[A]] =
+    IO(new IORef(a))
+}
```
jdnavarro: Are there any plans to have something that fulfills the role of `IORef` in `cats-effect`? This can come handy for the seqexec, too, but I was looking  into using `fs2` [`Ref`](https://oss.sonatype.org/service/local/repositories/releases/archive/co/fs2/fs2-core_2.12/0.9.6/fs2-core_2.12-0.9.6-javadoc.jar/!/fs2/util/Async$$Ref.html).

modules/ctl/src/main/scala/hi/deploy.scala
jdnavarro: Just re-read the top level comment and realized it's coming the next version, so never mind.

modules/ctl/src/main/scala/command.scala
tpolecat: @jdnavarro no. In principle you could write a compiler plugin for this but it would be very complicated.

modules/ctl/src/main/scala/hi/deploy.scala
tpolecat: Some minor syntax enhancements like `"123".parseInt` that are standard in scalaz are not provided by cats, so that's what mouse does. I'm pushing to merge it with cats for 1.0 though. It's kind of dumb.

modules/ctl/src/main/scala/ioref.scala
tpolecat: I don't know. The code here is all synchronous so a simple mutable cell is fine.

modules/ctl/src/main/scala/command.scala
jdnavarro: For GHC there is already a language extension for just this so that's why I was wondering if there was anything in the works for Scala. It was for 2 years in development before being available in a stable GHC release, so it must be tricky.

cquiroz:
My concerns are not as much on the technical merits of this PR or `cats`, but on how we are organizing (or not) development:

* Effect on the schedule: We wanted to merge `ocs3` with `gem`, but `gem` keeps moving the goal post making it difficult to stay abreast and at the same time progress on taking the `seqexec` to production. A change from `scalaz` to `cats` will possibly produce an important delay. I'd expect to have an assessment or discussion on the cost/effort before embarking on sweeping changes.

* Lack of process. It bothers me the lack of a process or discussion on how do we decide on using libs X or Y, or advanced techniques. In the past we had taken the approach of discussing it on the team and perhaps do some experimentation, as a result in many cases we have rejected certain libraries for going too far, having poor documentation, not being OSGi friendly, etc. At the moment there doesn't seem to be such process. That maybe ok in certain niches but when we are talking about changing something at the core like `cats` from `scalaz`, the impact is broad and affects the work of everybody on the team.

cquiroz:
Note that I'm totally tainted by the terrible experience of how `scalaz` was introduced to the team and its effects on schedule. Sorry for any unfairness.

swalker2m:
> Effect on the schedule: We wanted to merge ocs3 with gem, but gem keeps moving the goal post 

This is my fault as I was keen to do this perhaps before we're really ready.  What this is telling me is that perhaps this goal is premature.  We're exploring how best to do things still and the `gem` project will need to continue to evolve.  We might revisit this goal further down the road after the initial seqexec release.

> Lack of process.

To a large extent, the pull request discussion serves as the process.  It gives everybody a chance to see proposed changes and comment before they are committed.  A PR can always be opened with a proposal before doing any work as well.  I'm open to scheduling a meeting though whenever there are larger concerns than can be handled by a discussion here.  For the specific case of scalaz, I don't think there is much of an option unless we want to freeze everything since it is basically deprecated.

modules/ctl/src/main/scala/ioref.scala
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.ctl
+
+import cats.effect.IO
+
+// N.B. this isn't in cats-effect yet
+@SuppressWarnings(Array("org.wartremover.warts.Var"))
+final class IORef[A] private (var value: A) {
```
cquiroz: Should access to `value` be synchronized?

build.sbt
```diff
@@ -255,11 +259,12 @@ lazy val ctl = project
...
-    resolvers += "bmjames Bintray Repo" at "https://dl.bintray.com/bmjames/maven",
+    resolvers += Resolver.bintrayRepo("bkirwi", "maven"),
     libraryDependencies ++= Seq(
-      "org.scalaz"  %% "scalaz-core"   % scalazVersion,
-      "org.scalaz"  %% "scalaz-effect" % scalazVersion,
-      "net.bmjames" %% "scala-optparse-applicative" % "0.5"
+      "org.typelevel"           %% "cats"        % catsVersion,
+      "org.typelevel"           %% "cats-effect" % catsEffectVersion,
+      "com.monovore"            %% "decline"     % declineVersion,
+      "com.github.benhutchison" %% "mouse"       % mouseVersion
```
cquiroz: How much can `mouse` supplant the variety of methods we use in scalaz. e.g. `option` or the `? a | b` type of expressions?

Is it expected that all of them will be made available via `mouse`?

modules/ctl/src/main/scala/ioref.scala
tpolecat: This code is single-threaded but yeah that's a good idea. I'll fix it.

build.sbt
tpolecat: Yeah all of those should be in mouse. If they're not we can PR them.

tpolecat:
Closing this … I'm working off this branch on the grander upgrade.

PR #66, opened 2017-07-05
Increase precision of Angle/HourAngle

tpolecat:
This increases the precision of `Angle` and `HourAngle` by three orders of magnitude, allowing exact representations down to 1µas for angles, and 1ms / 15µas for hour angles. This was needed to support the Hamamatsu CCD, whose pixel size is known to 100µas precision.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/src/main/scala/gem/math/Angle.scala
@@ -4,7 +4,7 @@
 package gem.math
 
 import scalaz.{ Monoid, Show }
-import scalaz.std.anyVal.intInstance
+import scalaz.std.anyVal.longInstance
 import scalaz.syntax.equal._
 
 /**
@@ -14,7 +14,7 @@ import scalaz.syntax.equal._
  * divisible by 15 milliarcseconds) is represented by the HourAgle subtype.
  * @param toMilliarcseconds This angle in milliarcseconds. Exact.
  */
-sealed class Angle protected (val toMilliarcseconds: Int) {
+sealed class Angle protected (val toMilliarcseconds: Long) {
 
   // Sanity checks … should be correct via the companion constructor.
   assert(toMilliarcseconds >= 0, s"Invariant violated. $toMilliarcseconds is negative.")
@@ -77,7 +77,7 @@ sealed class Angle protected (val toMilliarcseconds: Int) {
     }
 
   override final def hashCode =
-    toMilliarcseconds
+    toMilliarcseconds.toInt
 
 }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -4,33 +4,33 @@
...
-    Angle.fromMilliarcseconds(-toMilliarcseconds.toLong)
+    Angle.fromMicroarcseconds(-toMicroarcseconds.toLong)
 
   /** This angle in decimal degrees. Approximate, non-invertible */
   def toDoubleDegrees: Double =
-    toMilliarcseconds.toDouble / (60.0 * 60.0 * 1000.0)
+    toMicroarcseconds.toDouble / (60.0 * 60.0 * 1000.0 * 1000.0)
 
   /** This angle in decimal radisns. Approximate, non-invertible */
   def toDoubleRadians: Double =
```
cquiroz: Would it makes sense to indicate the approx nature of these methods on the name?

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -119,54 +127,63 @@ object Angle {
...
  * addition, where the inverse is reflection around the 0-12h axis. This is a subgroup of the
- * integral Angles where milliarcseconds are evenly divisible by 15.
+ * integral Angles where microarcseconds are evenly divisible by 15.
  */
-final class HourAngle private (mas: Int) extends Angle(mas) {
+final class HourAngle private (µas: Long) extends Angle(µas) {
 
   // Sanity checks … should be correct via the companion constructor.
-  assert(toMilliarcseconds %  15 === 0, s"Invariant violated. $mas isn't divisible by 15.")
+  assert(toMicroarcseconds %  15 === 0, s"Invariant violated. $µas isn't divisible by 15.")
```
cquiroz: Will you use the assert or rather make the `apply` return an `Option`?

tpolecat: `apply` normalizes the angle to [0-360°) so it never fails. This assertion is just another check that `apply` is doing what it's supposed to do.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: I guess that's how I read `: Double` … all bets are off once you go to floating point. I'm not opposed to renaming but I wonder what the convention should be in the codebase generally since there's a lot of floating point.

cquiroz: I'm not sure either. I'm thinking on the concept of `unsafeXXX` methods, maybe `approximalDoubleRadians`. It doesn't look that nice though

swalker2m: I think the fact that the type is `Double` is probably enough, as @tpolecat says.

PR #67, opened 2017-07-06
Port ProgramId

tpolecat:
Work in progress:

- Adds a `half` enum+ADT in the database for `A`/`B`, with decimal representations 0 and 1.
- Adds existing timezone info to `Site` enum+ADT.
- Adds `Semester` data type with conversions to/from `java.time` types.
- Adds `ProgramId` data type.
- Adds parsers for all of the above.

This is a lot of code but it's pretty evenly split between `src/main` and `src/test`.


tpolecat:
```scala
scala> val s = Semester(Year.of(2010), Half.A)
s: gem.Semester = Semester(2010A)
 
scala> s.start.localDate
res0: java.time.LocalDate = 2010-02-01
 
scala> s.start.localDateTime
res1: java.time.LocalDateTime = 2010-01-31T14:00
 

scala> s.start.atSite(Site.GN)
res2: java.time.ZonedDateTime = 2010-01-31T14:00-10:00[Pacific/Honolulu]
 
scala> s.start.atSite(Site.GN).toInstant
res3: java.time.Instant = 2010-02-01T00:00:00Z
 
scala> Semester.fromSiteAndInstant(Site.GS, Instant.now)
res4: gem.Semester = Semester(2017A)
 
scala>
```

modules/core/src/main/scala/gem/Semester.scala
```diff
@@ -0,0 +1,99 @@
...
+  /** The semester immediately preceding this one. */
+  def prev: Semester =
+    plusSemesters(-1)
+
+  /** Module of various representations of the semester start. */
+  object start {
+    lazy val yearMonth: YearMonth = YearMonth.of(year.getValue, half.startMonth)
+    lazy val localDate: LocalDate = LocalDate.of(year.getValue, half.startMonth, 1)
+    lazy val localDateTime: LocalDateTime = LocalDateTime.of(localDate, LocalTime.MIDNIGHT).minusHours(10) // 2pm the day before
+    def zonedDateTime(zoneId: ZoneId): ZonedDateTime = ZonedDateTime.of(localDateTime, zoneId)
```
swalker2m: I probably would have just provided the corresponding `Instant` but this is easily convertible.

cquiroz:
Can you give some examples on the painful parts of the java time API

modules/core/src/main/scala/gem/enum/package.scala
```diff
@@ -30,4 +34,38 @@ package object enum {
         case SmartGcalType.NightBaseline => baseline(GcalBaselineType.Night)
       }
   }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  implicit class HalfCompanionOps(companion: Half.type) {
+
+    def unsafeFromInt(n: Int): Half =
```
cquiroz: Could you make it total by doing `n % 2`

modules/core/src/main/scala/gem/Semester.scala
```diff
@@ -0,0 +1,99 @@
...
+
+/** A (Year, Half) pair. */
+final case class Semester(year: Year, half: Half) {
+
+  /** This Semester plus the given number of years. */
+  def plusYears(n: Int): Semester =
+    copy(year = year.plusYears(n.toLong))
+
+  /** This Semester plus the given number of half-years. */
+  def plusSemesters(n: Int): Semester = {
```
cquiroz: Date math is hard. This may benefit from some tests

modules/core/src/main/scala/gem/enum/package.scala
tpolecat: It would, but that's not the semantics I want. We want `unsafeFromInt(n).toInt == n` which wouldn't be the vase if we normalized `n`. It makes sense for angles but I don't think it makes sense for enumerated types.

modules/core/src/main/scala/gem/Semester.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  /** The semester immediately preceding this one. */
+  def prev: Semester =
+    plusSemesters(-1)
+
+  /** Module of various representations of the semester start. */
+  object start {
+    lazy val yearMonth: YearMonth = YearMonth.of(year.getValue, half.startMonth)
+    lazy val localDate: LocalDate = LocalDate.of(year.getValue, half.startMonth, 1)
+    lazy val localDateTime: LocalDateTime = LocalDateTime.of(localDate, LocalTime.MIDNIGHT).minusHours(10) // 2pm the day before
```
cquiroz: If I read this correctly, the semester A starts january 31st at 2pm. Is that correct?

modules/core/src/main/scala/gem/Semester.scala
tpolecat: Yeah it needs tests. I played with it in the REPL and it seems to be right.

modules/core/src/main/scala/gem/Semester.scala
tpolecat: Correct.

project/gen2.scala
```diff
@@ -32,13 +32,17 @@ object gen2 {
   implicit val AngleMeta: Meta[Angle] =
     Meta[Double].xmap(Angle.fromArcsecs, _.toArcsecs)
 
+  implicit val ZoneIdMeta: Meta[ZoneId] =
+    Meta[String].xmap(ZoneId.of, _.toString)
```
cquiroz: `ZoneId.of` can throw an exception for unknown zone ids.Would that affect the code reading from the db?

cquiroz:
Looking good 👍

project/gen2.scala
tpolecat: This is only used at build time so we will know when we try to compile, but it's also possible that the runtime JVM won't have the timezones, in which case gem will just crash. It can't run without this information.

modules/core/src/main/scala/gem/Semester.scala
tpolecat: Yeah I don't know what we'll end up wanting. This can be trimmed down.

modules/core/src/main/scala/gem/enum/package.scala
cquiroz: This made me realize that I don't see the definition of `Half`. Why do you need a `toInt` method? for serializing to the db?

modules/core/src/main/scala/gem/enum/package.scala
```diff
@@ -30,4 +34,38 @@ package object enum {
         case SmartGcalType.NightBaseline => baseline(GcalBaselineType.Night)
       }
   }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  implicit class HalfCompanionOps(companion: Half.type) {
```
cquiroz: Could this be an `AnyVal` class?

tpolecat: Yes, it could be. I'll fix that.

tpolecat:
So, this last commit added parsers for parsing things, rather than using regexes and string manipulation. They're all in a single module now but are referenced by the `fromString` methods on various companions. This has some benefits:

- Parsers compose so we can re-use bits and pieces rather than repeating things.
- They're safer than regexes and easier to maintain.
- We can get better error messages, if desired.
- These parsers can tell you when partial input is ok "so far" so the web UI can show red/yellow/green input fields.
- I maintain this lib so we can get things changed/fixed quickly if needed.

The concern of course is in adding another library, but it's small and I think it's worthwhile.

cquiroz:
I'm ok with `atto` though I wish that we could unify its usage, avoiding having other parser options. I think this is the case so far across gem and ocs3

tpolecat:
Hopefully just having atto on the classpath will prevent people from doing stringy stuff, since there's basically zero barrier to entry. You don't have to extend anything, etc., like with stdlib parsers.

tpolecat:
The last commit rejiggered the tests a little bit. I put all the `Arbitrary` instances in their own files in an `arb` package, which seems more maintainable than stuffing them into big files. Each has `trait ArbWhatever` for `Arbitrary[WhateverAndCloselyRelatedThings]` and then `object ArbWhatever extends ArbWhatever` which lets you get the instances either by extending the trait or by importing members from the object. I prefer the latter approach, but eventually we might want to roll things up into bigger imports and this gives us a mechanism to do it. I didn't re-org existing ones but I can do a follow-on PR for that, if there are no objections to this strategy.

modules/core/src/main/scala/gem/enum/package.scala
tpolecat: `Half` is generated from the db along with all the other enumerated types. Look in `modules/core/target/scala-2.12/src_managed/main/gem`.

The `.toInt` method gives a numeric meaning to the ordering, which lets us do math with semesters by treating `(year, half)` as a single value `2y + h`, which turns `plusSemesters/prev/next` into simple integer arithmetic without the weird boundary logic. It could be inlined in `Semester` but it seemed like it might come in handy elsewhere so I pushed it down into `Half`.

tpolecat:
Ok assuming the tests pass I think this is ready for a real review. Please comment on the full change set rather than on individual commits because they're easy to miss otherwise.

tpolecat:
Note that this ports `ProgramId` but does not (yet) update the gem program model to use it. That's next.

tpolecat:
Hm.

tpolecat:
Bah, incremental compilation.

modules/core/src/main/scala/gem/ProgramId.scala
```diff
@@ -0,0 +1,211 @@
...
+      Science.unsafeApply(site, semester, programType, index)
+
+    override def format =
+      s"${site.shortName}-${semester.format}-${programType.shortName}-$index"
+
+  }
+  object Science {
+
+    /** Construct a `Science` program id, if `index` is positive. */
+    def apply(site: Site, semester: Semester, programType: ProgramType, index: Int): Option[Science] =
```
swalker2m: Many times I've wished for an unsigned `Int` type.

modules/core/src/test/scala/gem/enum/Arbitraries.scala
```diff
@@ -1,69 +0,0 @@
...
-
-trait Arbitraries {
-
-  private def pick[A](as: List[A]): Arbitrary[A] =
-    Arbitrary(oneOf(as))
-
-
-  // Flamingos2
-
-  implicit val arbF2Disperser   = pick(F2Disperser.all  )
```
swalker2m: Do we still get arbitraries for the instrument enums somewhere?  I'm not seeing them.

modules/sql/src/main/resources/db/migration/V015__half.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TYPE half AS ENUM (
+    'A', 'B'
+);
```
swalker2m: Should the program types be defined in the database as well?

modules/core/src/main/scala/gem/ProgramId.scala
cquiroz: Yeah it would be useful. Perhaps something like

```
case class UnsignedNum[A](val a: A)(implicit n: Numeric[A])
```

You'd still get an option on the `apply` though

swalker2m: Spire has an interesting [UInt](https://github.com/non/spire/blob/master/core/shared/src/main/scala/spire/math/UInt.scala). It stores an ordinary signed `Int` but treats the negative part as numbers beyond `MaxValue`.

tpolecat: Yeah, a better way to do this is with https://github.com/fthomas/refined but I didn't want to introduce a new foundational library. We still need to figure out what we're doing about units/dimensions/etc. so I kind of wanted to defer that question a bit longer.

modules/core/src/main/scala/gem/ProgramId.scala
```diff
@@ -0,0 +1,211 @@
...
+    /** Parse a `Science` program id from a string, if possible. */
+    def fromString(s: String): Option[ProgramId] =
+      Parsers.parseExact(Parsers.programId.science)(s)
+
+    /** `Science` program ids are ordered pairwise by their data members. */
+    implicit val ScienceOrder: Order[Science] =
+      Order[Site]       .contramap[Science](_.site)        |+|
+      Order[Semester]   .contramap[Science](_.semester)    |+|
+      Order[ProgramType].contramap[Science](_.programType) |+|
+      Order[Int]        .contramap[Science](_.index)
```
tpolecat: You can also do `orderBy(a => (a.site, a.semester, ...` and do the ordering by constructing tuples and comparing those since the instance is available for free, but this should be a little faster since it's just selecting members and not constructing anything.

modules/core/src/main/scala/gem/ProgramId.scala
cquiroz: Refined sounds like a nice lib but it is a very strong dependency at the very core of how you define the classes. It uses a compiler plugin right?

modules/core/src/main/scala/gem/ProgramId.scala
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Programs are ordered lexically by prodict prefix (Daily, Nonstandard, then Science) and then
+   * by the defined orderings for individual cases when constructors match.
+   */
+  implicit val ProgramIdOrder: Order[ProgramId] =
+    Order.order {
+      case (a: Science,     b: Science)     => a cmp b
+      case (a: Daily,       b: Daily)       => a cmp b
+      case (a: Nonstandard, b: Nonstandard) => a cmp b
+      case (a,              b)              => a.productPrefix cmp b.productPrefix
```
tpolecat: Yeah, I'm not really happy with that either. I'll write out the cases. Should we order by Science, Daily, then Nonstandard?

modules/core/src/main/scala/gem/Semester.scala
```diff
@@ -0,0 +1,124 @@
...
+   * This Semester plus the given number of years.
+   * @throws DateTimeException for final year out of range -999999999 - 999999999
+   */
+  def plusYears(n: Int): Semester =
+    copy(year = year.plusYears(n.toLong))
+
+  /** This Semester plus the given number of half-years. */
+  def plusSemesters(n: Int): Semester = {
+    val yy = year.getValue
+    val hs = yy * 2 + half.toInt * yy.signum + n
```
tpolecat: `Semester` takes an arbitrary `Year`, which can be negative. It's just to satisfy the tests really.

modules/core/src/test/scala/gem/enum/Arbitraries.scala
tpolecat: Yes, every generated enum has an `Enumerated` instance so we can abstract over them. The `Arbitrary` instances arise from [`arb/Enumerated`](https://github.com/gemini-hlsw/gem/pull/67/files#diff-f104c02e609c7657aab942d113cb487e).

modules/sql/src/main/resources/db/migration/V015__half.sql
tpolecat: You mean an enumeration for `Science`, `Daily`, `Nonstandard`? Yes I think so and this tag can be used as the basis for the ordering question above.

modules/core/src/main/scala/gem/ProgramId.scala
tpolecat: Opened https://github.com/gemini-hlsw/gem/issues/68 to track this.

modules/core/src/test/scala/gem/enum/Arbitraries.scala
swalker2m: Oh, sure that's much better.

cquiroz:
Considering that things like `ProgramId`, `Semester`, etc is something we want to use on the front end too. It would be good to compile these modules to `scala.js` as well just to catch potential problems earlier

tpolecat:
@cquiroz that's a good idea re: scala-js. the intent is that at least all of core should cross-compile.

cquiroz:
yeah, let's plan to do that in the near future. It may involve splitting some modules due to dependencies, e.g. doobie is not scala.js compatible I suppose

tpolecat:
Core only depends on scalaz, shapeless, and atto so it should work.

cquiroz:
nice, then it should be no troubles, let's see when we get typelevel compiler and scala.js to play along too

PR #70, opened 2017-07-11
remove spmodel core

tpolecat:
This adds an Offset class and removes the last uses of the spModel.core. We are now safely unmoored. I will build out `Offset` a bit in a followup PR.


modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -111,10 +118,14 @@ object Angle {
   def fromDegrees(ms: Int): Angle =
     fromArcminutes(ms * 60)
 
-  /** Construct a new Angle of the given magnitide in integral degrees, modulo 360°. Approximate. */
+  /** Construct a new Angle of the given magnitide in double degrees, modulo 360°. Approximate. */
```
cquiroz: magnitude?

modules/json/src/main/scala/gem/json.scala
```diff
@@ -20,7 +20,10 @@ package object json {
 
   // Angle mapping to signed arcseconds. NOT implicit.
   val AngleMetaAsSignedArcseconds: CodecJson[Angle] =
-    CodecJson.derived[Double].xmap(Angle.fromArcsecs)(_.toSignedDegrees * 3600)
+    CodecJson.derived[BigDecimal]
```
cquiroz: Does this mean that on the wire the value will have the decimal point shifted?

tpolecat: Yeah the wire value will be in signed arcseconds. This is for offsets so the values are like `-0.25` or whatever. I think probably it makes sense to pass them as full-precision angles but we're not there yet. This is mostly just a sanity check that we're not doing anything to preclude JSON encoding.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: Can you think of a better word?

cquiroz: Is magnitide a word? I didn't know

modules/json/src/main/scala/gem/json.scala
cquiroz: Makes more sense now thinking that is a different unit 👍

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: Oh, typo. Thanks.

tpolecat:
Close and re-opened to see if I can get my new commits to show up. Strange.

tpolecat:
Oh, heh, wrong branch.

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: Fixed in the next PR.

modules/core/src/main/scala/gem/math/Offset.scala
```diff
@@ -0,0 +1,25 @@
...
+  object P {
+    val Zero: P =
+      P(Angle.Angle0)
+  }
+
+  final case class Q(toAngle: Angle)
+  object Q {
+    val Zero: Q =
+      Q(Angle.Angle0)
+  }
```
swalker2m: Will we define `Order`, etc. or the equivalent in cats once we switch?  I guess you did say you would build it out more in a subsequent PR.

modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -75,19 +75,23 @@ package object dao extends MoreTupleOps with ToUserProgramRoleOps {
       EitherConnectionIO.right[A, T](c)
   }
 
-  // Angle mapping to signed arcseconds via NUMERIC. NOT implicit.
+  // Angle mapping to signed arcseconds via NUMERIC. NOT implicit. We're mapping a type that
+  // is six orders of magnitude more precise than the database column, so we will shift
+  // the decimal point back and forth.
```
swalker2m: Maybe the database columns should have more precision?  For offsets that would be overkill, but does this mean we are limited in how precisely an `Angle` can be stored in the database? If there is just one precision for all angles it should probably be the most precise value supported by `Angle`?

project/gen2.scala
```diff
@@ -19,6 +17,12 @@ import shapeless.record._
 
 object gen2 {
 
+  final case class Angle(toArcsecs: Double)
+  object Angle {
+    def fromArcsecs(d: Double) =
+      Angle(d)
+  }
```
swalker2m: Why do we need a second `Angle` here?

modules/core/src/main/scala/gem/math/Offset.scala
tpolecat: Yeah I have another branch waiting. We can discuss there.

modules/db/src/main/scala/gem/dao/package.scala
tpolecat: Yeah this isn't implicit so it's not the canonical mapping for angles. It's used explicitly by offsets so a human looking at the database will see a signed arcsecond value like `-1.25`. Same with the JSON mapping. I think as I build out the target model we'll get a better handle on how to map angles in a way that's precise but also human-readable, so I may revisit this.

project/gen2.scala
tpolecat: In the build build we don't have access to OCS2's `Angle` anymore, nor gem's `Angle` since it hasn't been built yet. This is just enough structure to read an angle from the database and insert its value into the generated Scala code.

PR #73, opened 2017-07-12
Instances for Offset, Angle, Semester

tpolecat:
This adds instances and tests for `Offset` and unrelatedly fixes a typo Carlos found.

Shane, you mentioned wanting `Order` instances. Do we need to sort offsets by size?

### DIFF
--- a/modules/core/src/main/scala/gem/math/Offset.scala
+++ b/modules/core/src/main/scala/gem/math/Offset.scala
@@ -4,7 +4,8 @@
 package gem
 package math
 
-import scalaz.Monoid
+import scalaz.{ Equal, Monoid, Show }
+import scalaz.std.tuple._
 
 /** Angular offset with P and Q components. */
 final case class Offset(p: Offset.P, q: Offset.Q) {
@@ -13,7 +14,7 @@ final case class Offset(p: Offset.P, q: Offset.Q) {
   def unary_- : Offset =
     Offset(-p, -q)
 
-  /** Componentwise sum of this offet and `o`. Exact. */
+  /** Componentwise sum of this offset and `o`. Exact. */
   def +(o: Offset): Offset =
     Offset(p + o.p, q + o.q)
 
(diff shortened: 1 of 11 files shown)


tpolecat:
This is mostly boilerplate and it's repetitive but I could't figure out a useful way of abstracting it out without inventing an `AbGroup` typeclass, which we will get for free when we switch to cats and introduce Spire.

modules/core/src/main/scala/gem/math/Offset.scala
```diff
@@ -4,22 +4,77 @@
...
+import scalaz.Monoid
+
+/** Angular offset with P and Q components. */
+final case class Offset(p: Offset.P, q: Offset.Q) {
+
+  /** This offset, with both components reflected around the 0 .. 180° axis. Exact, invertable. */
+  def unary_- : Offset =
+    Offset(-p, -q)
+
+  /** Componentwise sum of this offet and `o`. Exact. */
```
cquiroz: typo: s/offet/offset

modules/core/src/test/scala/gem/math/OffsetPSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  // Compilation test
+  protected val a0 = implicitly[Monoid[Offset.P]]
+
+  "Conversion to angle" must "be invertable" in {
+    forAll { (p: Offset.P) =>
+      Offset.P(p.toAngle) shouldEqual p
+    }
+  }
+
+  "Offset.P forms an Abelian Group over addition. It" must "be associative" in {
```
cquiroz: Unrelated to the PR, but github doesn't know how to color code this type of code, confusing...

tpolecat:
I'll do `Equal` for now and we can deal with ordering if it comes up.

tpolecat:
I added `Equal` and `Show` a few places where they were missing, plus tests for naturality (meaning they're consistent with `==` and `.toString`).

PR #74, opened 2017-07-13
Clean up Dataset

tpolecat:
This cleans up `Dataset`a bit and adds some comments, instances, arbitraries, and minimal tests

### DIFF
--- a/modules/core/src/main/scala/gem/Dataset.scala
+++ b/modules/core/src/main/scala/gem/Dataset.scala
@@ -4,21 +4,24 @@
 package gem
 
 import java.time.Instant
-
 import scalaz._, Scalaz._
 
+/** A labeled, timestamped file. */
 final case class Dataset(
-  label: Dataset.Label,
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/gem/Dataset.scala
```diff
@@ -31,6 +34,27 @@ object Dataset {
...
+      Order[Observation.Id].contramap[Label](_.observationId) |+|
+      Order[Int]           .contramap[Label](_.index)
+
+    implicit val LabelShow: Show[Label] =
+      Show.showA
+
   }
 
+  /**
+   * Labels are ordered by their ids, which are normally unique. For completeness they are further
```
swalker2m: "_Datasets_ are ordered by their _labels_, which are normally unique."

tpolecat: I'll fix this in the next pr.

PR #77, opened 2017-07-14
enumerated ordering

tpolecat:
This fixes the ordering of enumerated types to respect the order of the `all` collection. In practice what this means is that the sort order for generated enumerated types will now match the order in which they members are selected from the database by the code generator, and coincidentally the order in which they are declared in the generated code.

### DIFF
--- a/modules/core/src/main/scala/gem/Enumerated.scala
+++ b/modules/core/src/main/scala/gem/Enumerated.scala
@@ -21,7 +21,7 @@ trait Enumerated[A] extends Order[A] {
   /** Select the member of this enumeration with the given tag, throwing if absent. */
   def unsafeFromTag(tag: String): A = fromTag(tag).getOrElse(sys.error("Invalid tag: " + tag))
 
-  /** Relative order of `a` and `b` in `all`;. */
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/gem/Enumerated.scala
```diff
@@ -6,20 +6,33 @@ package gem
...
+
+  // Hashed index lookup, for efficient use as an `Order`.
+  private lazy val indexOfTag: Map[String, Int] =
+    all.zipWithIndex.map { case (a, n) => (tag(a), n) } (collection.breakOut)
+
 }
 
 object Enumerated {
-  def apply[A](implicit ev: Enumerated[A]): Enumerated[A] = ev
+  def apply[A](implicit ev: Enumerated[A]): ev.type = ev
```
tpolecat: Returning the singleton type here is a better way to implement this pattern because it preserves refinements. Not relevant here but a good habit.

swalker2m: Maybe an example of a refinement and how you would make use of it would help me here.

PR #78, opened 2017-07-14
re-org, doc

tpolecat:
Ok, before you freak out. There is no new code here. It's all re-organization and comments, for the purpose of having a fighting chance at good developer doc.

- All sealed trait constructors are in their companion objects; `F2DynamicConfig` is now `DynamicConfig.F2` for instance.
- A good hunk of `core` has reasonable Scaladoc now, and browsing around the API makes some amount of sense now because stuff is grouped and nested reasonably.

To see the effect, do `core/doc` in sbt on your local branch and look at the generated doc (it will helpfully print out the target dir). It's kind of crappy. Then compare with [this](http://tpolecat.github.io/doc/api/gem/index.html) which is how it looks after this PR.


modules/core/src/main/scala/gem/enum/package.scala
```diff
@@ -10,21 +10,31 @@ import scalaz.Scalaz._
 // The members of this package are generated from database tables, which are the source of truth.
 // See project/gen2.scala for details. Associations with other model types, as needed, are provided
 // here as implicit classes wrapping val the generated companion object extends AnyVals.
+
+/**
+ * Enumerated types (normally generated from database tables) and related syntactic snrichments.
```
swalker2m: enrichments

tpolecat:
Ok fixed. Needs re-approval.

tpolecat:
Ok rebased, fixed those typos.

tpolecat:
Hm I updated the doobie snapshot and something has broken. Something weird in the bytecode.

tpolecat:
Ok, so it looks like coursier doesn't deal well with moving snapshots so I turned it off. This exposed a breakage that I fixed. @swalker2m this may also explain your weird issue.

PR #82, opened 2017-07-19
Add GMOS SmartGcal WIP

swalker2m:
This PR is a work-in-process update to Smart Gcal.  The purpose is to support GMOS Smart Gcal, but that necessitated a few fundamental changes:

1. The GMOS `.csv` files from OCS2 are hyper-complex.  To simplify parsing, I created an `exportSmartGcal` ODB shell command which exports to an easier to read format.  I've updated the Smart Gcal import to use that.  Since the GMOS `.csv` file is rather large with all the wildcards and regexes expanded, it was necessary to use `scalaz.stream` and doobie bulk importing features in order to read them within a reasonable(ish) time. Even so, GMOS-N requires almost 3 minutes so there is likely room for improvement.

2. GMOS supports defining wavelength ranges in the Smart Gcal mapping.  A GMOS step has a central wavelength value but the GCal config to which it corresponds is usable for a range of wavelengths.  This required splitting the existing `SmartGcalKey` into `SmartGcalSearchKey` and `SmartGcalDefinitionKey`. In the GMOS example, the search key would include the exact wavelength of the GMOS step while the definition key includes the wavelength range.

3. Previously GCal configurations for manual gcal steps were kept together with Smart Gcal configs.  You could differentiate the two based on whether the `gcal` `step_id` was `null`.  For manual gcal steps the `step_id` would be set but for smart gcal mappings it would be `null`.   The existing `step_gcal` table has now been expanded with duplicate `gcal` configuration columns, simplifying the queries and avoiding a join.  This also makes it trivial and fast to clean the Smart Gcal mappings before importing.

So far, only GMOS-N is supported and some additional cleanup and testing is needed.  Since this PR is already so big though, I wanted a checkpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,16 +79,12 @@ There are no other instrument-specific slices for science steps yet.
 
 #### Importing Smart Gcal Configuration
 
-Smart Gcal configuration is stored in database tables like everything else instead of in `.csv` files downloaded from SVN as in OCS2.  The `.csv` files can be imported though much like old program `.xml` files.  First, make a symlink to a directory containing the `.csv` files. 
+Smart Gcal configuration is stored in database tables instead of in `.csv` files downloaded from SVN as in OCS2.  The `.csv` files can be imported, though much like old program `.xml` files, we use a modified format.  To obtain compatible Smart Gcal `.csv` files from OCS2, start the ODB and use the `exportSmartGcal` shell command providing the name of a directory into which to write the files.
+
+Next, make a symlink to the directory containing the `.csv` files that were exported:
 
 ```
 ln -s /path/to/old/smart/gcal/csv smartgcal
-```
-
-If you've ever used the old OT or ODB, you will likely have downloaded the `.csv` files for Smart Gcal. For example, for the production OT you can find them in your home directory:
-
-```
-~/.ocs15/Gemini\ OT\ 2017A.1.1.2_mac/data/jsky.app.ot/smartgcal
 ```
 
 Having created the symlink, you can then import from the `sbt` prompt:
--- a/modules/core/src/main/scala/gem/config/Gmos.scala
+++ b/modules/core/src/main/scala/gem/config/Gmos.scala
@@ -6,6 +6,8 @@ package gem.config
 import gem.enum._
 import java.time.Duration
 
+import scalaz._, Scalaz._
+
 /**
  * Additional type hierarchy over the low-level GMOS enums.
  * @group Instrument-Specific Models
@@ -81,9 +83,16 @@ object Gmos {
   )
 
   /** GMOS grating central wavelength.  For now, just a value class wrapper
-    * around a BigDecimal.  This should be switched to Fixed / squants?
+    * around an integer.
+    * TODO: wavelength. This class needs to be converted to a generic Wavelength
+    * class and built out.
     */
-  final case class GmosCentralWavelength(val nm: BigDecimal) extends AnyVal
+  final case class GmosCentralWavelength(val toAngstroms: Int) extends AnyVal
+
+  object GmosCentralWavelength {
+    implicit val OrderGmosCentralWavelength: Order[GmosCentralWavelength] =
+      Order.orderBy(_.toAngstroms)
+  }
 
   /** GMOS grating configuration, parameterized on the disperser type.  These
     * are grouped because they only apply when using a grating.  That is, all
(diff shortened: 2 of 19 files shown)


tpolecat:
This looks great so far. GMOS is kind of complicated. 👍

README.md
```diff
@@ -79,16 +79,12 @@ There are no other instrument-specific slices for science steps yet.
...
 
-Smart Gcal configuration is stored in database tables like everything else instead of in `.csv` files downloaded from SVN as in OCS2.  The `.csv` files can be imported though much like old program `.xml` files.  First, make a symlink to a directory containing the `.csv` files. 
+Smart Gcal configuration is stored in database tables instead of in `.csv` files downloaded from SVN as in OCS2.  The `.csv` files can be imported, though much like old program `.xml` files, we use a modified format.  To obtain compatible Smart Gcal `.csv` files from OCS2, start the ODB and use the `exportSmartGcal` shell command providing the name of a directory into which to write the files.
 
-```
-ln -s /path/to/old/smart/gcal/csv smartgcal
-```
-
-If you've ever used the old OT or ODB, you will likely have downloaded the `.csv` files for Smart Gcal. For example, for the production OT you can find them in your home directory:
+Next, make a symlink to the directory containing the `.csv` files that were exported:
```
cquiroz: Do we need this procedure to start developing on gem?

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -81,9 +83,16 @@ object Gmos {
   )
 
   /** GMOS grating central wavelength.  For now, just a value class wrapper
-    * around a BigDecimal.  This should be switched to Fixed / squants?
+    * around an integer.
+    * TODO: wavelength. This class needs to be converted to a generic Wavelength
+    * class and built out.
     */
-  final case class GmosCentralWavelength(val nm: BigDecimal) extends AnyVal
+  final case class GmosCentralWavelength(val toAngstroms: Int) extends AnyVal
```
cquiroz: Is using `Int` the reason not to use squant's `Wavelength`?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -6,34 +6,58 @@ package gem.dao
...
+  def bulkInsertGmosNorth(entries: Vector[(GcalLampType, GcalBaselineType, SmartGcalKey.GmosNorthDefinition, GcalConfig)]): scalaz.stream.Process[ConnectionIO, Int] =
+    bulkInsert[SmartGcalKey.GmosNorthDefinition](Statements.bulkInsertGmosNorth, entries)
+
+  private def bulkInsert[K](
+                update:  Update[((GcalLampType, GcalBaselineType, K), Int)],
+                entries: Vector[(GcalLampType, GcalBaselineType, K, GcalConfig)]): scalaz.stream.Process[ConnectionIO, Int] =
+
+    Process.emitAll(entries.map(t => (t._1, t._2, t._3)))        // Process[ConnectionIO, (Lamp, Baseline, Key)]
+      .zip(GcalDao.bulkInsertSmartGcal(entries.map(_._4)))       // Process[ConnectionIO, ((Lamp, Baseline, Key), Id)]
+      .chunk(4096)                                               // Process[ConnectionIO, Vector[((Lamp, Baseline, Key), Id)]]
```
cquiroz: is this 4096 bytes or 4096 entries?

modules/core/src/main/scala/gem/config/Gmos.scala
tpolecat: Yeah ... Squants only likes `Double` right now, right?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
tpolecat: Elements. It groups every 4096 `((Lamp, Baseline, Key), Id)` valus into a `Vector`, then we insert each chunk.

README.md
swalker2m: You only need it if you have to work with Smart Gcal.  I suppose I could check in a snapshot of the updated `.csv` files, so you would have something to use without going through `exportSmartGcal` from an ODB shell.

swalker2m:
I guess what I'd like to do is merge this part, if that seems ok, and continue with GMOS-S etc. in a separate PR.

tpolecat:
(You may have conflicts so I would do it locally instead of hitting the merge button here.)

swalker2m:
Rebased.  Unfortunately needs a fresh review.

PR #84, opened 2017-07-19
wavelength

tpolecat:
This adds a `Wavelength` class that just wraps a positive `Int` (for integral angstroms) and doesn't support operations other than comparison. It's minimal but I'm not sure what else we need immediately so I'd prefer to leave it simple and wait until more use cases arise.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Wavelength.scala
+++ b/modules/core/src/main/scala/gem/math/Wavelength.scala
@@ -40,7 +40,7 @@ object Wavelength {
 
   /** Construct a wavelength from integral angstroms, if non-negative. */
   def fromAngstroms(angstroms: Int): Option[Wavelength] =
-    Some(angstroms).filter(_ > 0).map(new Wavelength(_))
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+  override final def hashCode =
+    toAngstroms
+
+}
+
+object Wavelength {
+
+  final lazy val ZeroAngstroms = unsafeFromAngstroms(0)
```
cquiroz: no monoid?

cquiroz: is it because it must be positive?

tpolecat: No, addition isn't closed … it might overflow.

tpolecat:
Hey Scalacheck found a bug! I wasn't allowing zero.

cquiroz:
Great use of scalacheck!

PR #85, opened 2017-07-20
Switch to Wavelength

swalker2m:
Small update to take advantage of `gem.math.Wavelength` in GMOS.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/package.scala
+++ b/modules/db/src/main/scala/gem/dao/package.scala
@@ -7,7 +7,7 @@ import doobie.postgres.imports._
 import doobie.imports._
 import doobie.enum.jdbctype.{ Distinct => JdbcDistinct, Array => _, _ }
 
-import gem.math.{ Angle, Offset }
+import gem.math.{ Angle, Offset, Wavelength }
 
 import java.sql.Timestamp
 import java.time.{Duration, Instant}
@@ -92,6 +92,10 @@ package object dao extends MoreTupleOps {
   // OffsetQ maps to a signed angle in arcseconds
   implicit val OffsetQMeta: Meta[Offset.Q] =
     AngleMetaAsSignedArcseconds.xmap(Offset.Q(_), _.toAngle)
+
+  // Wavelength maps to an integer in angstroms
+  implicit val WavelengthMeta: Meta[Wavelength] =
+    Meta[Int].xmap(Wavelength.unsafeFromAngstroms, _.toAngstroms)
 
   // Program.Id as string
   implicit val ProgramIdMeta: Meta[Program.Id] =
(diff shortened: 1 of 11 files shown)


modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -93,6 +93,10 @@ package object dao extends MoreTupleOps {
   implicit val OffsetQMeta: Meta[Offset.Q] =
     AngleMetaAsSignedArcseconds.xmap(Offset.Q(_), _.toAngle)
 
+  // Wavelength maps to an integer in angstroms
+  implicit val WavelengthMeta: Meta[Wavelength] =
+    Meta[Int].xmap(Wavelength.unsafeFromAngstroms, _.toAngstroms)
+
```
tpolecat: So, I think we may want to leave this non-implicit and have specialized wrappers for `GmosWavelength` or something that map to an exact decimal representation in nanometers (as we did with offsets, which are stored as decimal arcseconds) if that's what a human wants to see. I think it's important to be able to look at a database table and see values that make intuitive sense. We need to add comments to the columns in question, now that I think about it.

But if it's the case that every wavelength in the science model (as far as we know) makes sense in angstroms then this mapping seems fine to me.

modules/json/src/main/scala/gem/json.scala
```diff
@@ -25,6 +25,10 @@ package object json {
         b => Angle.fromMicroarcseconds(b.underlying.movePointLeft(6).longValue))(
         a => BigDecimal(new java.math.BigDecimal(a.toMicroarcseconds).movePointRight(6)))
 
+  // Wavelength mapping to integral Angstroms.
+  implicit val WavelengthCodec: CodecJson[Wavelength] =
+    CodecJson.derived[Int].xmap(Wavelength.unsafeFromAngstroms)(_.toAngstroms)
+
```
tpolecat: Same comment as above, assuming we want JSON to be human-readable, which I think is a good starting position.

modules/db/src/main/scala/gem/dao/package.scala
swalker2m: I thought about that but it seemed to me that the current state of representing the central wavelength in um in some cases and nm in others was too confusing and thought it best to guarantee that when you see a wavelength column you always know it is angstroms.

`Angle` seemed different because there are various usages and no universally applicable natural unit.  Maybe I'm wrong though.

tpolecat: Ok that makes sense. If wavelengths always show up as 123 Å then 👍

swalker2m:
After discussion with @tpolecat, I think that in fact we probably will need to represent wavelengths in different units in the database.  I'm merging now anyway, knowing that we'll come back to this, since this is an improvement over what is there now.

PR #87, opened 2017-07-24
User cleanup

tpolecat:
This cleans up some messiness in the user DAO. It's still unused other than `selectRootUser` since there's no way to create another user.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/UserDao.scala
+++ b/modules/db/src/main/scala/gem/dao/UserDao.scala
@@ -16,7 +16,7 @@ object UserDao {
    * raises an exception if the root user is missing.
    * @group Queries
    */
-  val selectRoot: ConnectionIO[User[ProgramRole]] =
+  val selectRootUser: ConnectionIO[User[ProgramRole]] =
     OptionT(selectUser(User.Id.Root)).getOrElse(sys.error("No root user. Cannot continue."))
 
   // Helper method to add roles to a selected user, if any.
@@ -28,14 +28,14 @@ object UserDao {
   } .run
 
   /**
-   * Select the given user, with roles.
+   * Select the given user, if any, with roles.
    * @group Queries
    */
   def selectUser(uid: User.Id): ConnectionIO[Option[User[ProgramRole]]] =
     selectUserImpl(Statements.selectUser(uid))
 
   /**
-   * Select the given user, with roles, if the id and password match a known user.
+   * Select the given user, if any, with roles, if the id and password match a known user.
    * @group Queries
    */
   def selectUserʹ(uid: User.Id, password: String): ConnectionIO[Option[User[ProgramRole]]] =
@@ -47,13 +47,6 @@ object UserDao {
    */
   def selectRoles(id: User.Id): ConnectionIO[Map[Program.Id, Set[ProgramRole]]] =
     Statements.selectRoles(id).list.map(_.foldMap { case (k, v) => Map((k -> Set(v))) })
-
-  /**
-   * Select the given user, with roles. Raises an exception if the user is not found.
-   * @group Queries
-   */
-  def selectUserWithRoles(id: User.Id): ConnectionIO[User[ProgramRole]] =
-    (Statements.selectUser(id).unique |@| selectRoles(id))((u, r) => u.copy(roles = r))
 
   /**
    * Attempts to change the specified user's password, yielding success. A cause of failure (user
--- a/modules/db/src/main/scala/gem/dao/UserDao.scala
+++ b/modules/db/src/main/scala/gem/dao/UserDao.scala
@@ -12,26 +12,34 @@ import doobie.imports._
 object UserDao {
 
   /**
-   * Select the super-user, for system processes.
+   * Select the root user, for system processes. Always succceds if the system is consistent;
+   * raises an exception if the root user is missing.
    * @group Queries
    */
-  val selectRoot: ConnectionIO[User[Nothing]] =
-    selectUser("root")
+  val selectRoot: ConnectionIO[User[ProgramRole]] =
+    OptionT(selectUser(User.Id.Root)).getOrElse(sys.error("No root user. Cannot continue."))
+
+  // Helper method to add roles to a selected user, if any.
+  private def selectUserImpl(q: Query0[User[Nothing]]): ConnectionIO[Option[User[ProgramRole]]] = {
+    for {
+      u <- OptionT(q.option)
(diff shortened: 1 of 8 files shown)


modules/db/src/main/scala/gem/dao/UserDao.scala
```diff
@@ -11,55 +11,74 @@ import doobie.imports._
...
-         }
+       """ ++ and).query[(String, String, String, String, Boolean)].map {
+        case (i, f, l, e, s) => User(i, f, l, e, s, Map.empty)
+      }
+
+    def selectUser(id: String): Query0[User[Nothing]] =
+      selectUserImpl(id, Fragment.empty)
+
+    def selectUserʹ(id: String, password: String): Query0[User[Nothing]] =
+      selectUserImpl(id, fr"AND md5 = md5($password)")
```
cquiroz: MD5 is no longer considered safe enough. Even SHA1 is considered weak, the current recommendations is along the lines of HMAC256 or HMAC512

cquiroz:
This looks very good in general, I'd be happy though, to delegate user handling to ITS on the form of their Active Directory DB. Nevertheless we should probably have the option to fallback to our own user db

tpolecat:
I don't know if it's practical to rely on ITS for authentication because they would need to create accounts for all the PIs and I doubt they'll want to do that.

cquiroz:
I think that has been discussed at some moment but I have no idea of the results. Still we'd probably want to combine, perhaps having our own db for PIs but still delegating to ITS for internal users

modules/db/src/main/scala/gem/dao/UserDao.scala
swalker2m: I think this is still in proof-of-concept stage.

tpolecat: Good idea. `pgcrypto` is a provided extension we can enable. It provides `hmac()`. Opened #89 to track.

cquiroz: nice, I didn't know postgres would support crypto directly. sounds good

tpolecat:
Opened #90 to track ITS user management.

cquiroz:
As we briefly discussed over chat we could even consider doing user management with other platforms, e.g. github, google, etc. But that should be discussed with science if it make sense for them

tpolecat:
I read a bit on OpenAuth and it's really a system for authorizing an application to do work on a user's behalf. It's not an authentication system, although it is sometimes used that way.

cquiroz:
At least OpenAuth2 lets you do stuff like authorizing a user which is essentially what you want. As a good standard every implementation is slightly different though 😄

cquiroz:
But again, maybe science prefers our own DB + Gemini rather than anything else

tpolecat:
If we use the Twitter we can tweet on the user's behalf! "My program was just demoted to Band 4!"

modules/db/src/main/scala/gem/dao/UserDao.scala
jdnavarro: I think in general any standard hashing algorithm is not recommended for storing passwords without several safety measures like salting, authenticating, etc.

I'd use one of the specialized libraries to store passwords `bcrypt` `scrypt` or `PKDF2` and forget about the details. I'm sure there'll be plenty of implementations in Java/Scala.

If we don't want to bother evaluating crypto algorithms every time we need to pick one, I'd use all the defaults in the [libsodium](https://download.libsodium.org/doc/) and forget about it. I never had the chance to use it because of legacy constraints but here I don't we have that problem.

jdnavarro: Just read `pgcryto` provides de `S2K` algorithm taken from PGP which takes care of the salting and it's purposely made slow. So I guess that should be good enough for us.

tpolecat: Let's keep in mind that we're not doing banking here. We can use something better than MD5 but honestly I don't think this is a huge problem. If you gain access to the database there's no point trying to crack the passwords since you have the data already. If you try to do it through the REST API it will take forever. We can also have an incorrect password lock-out if we want. Let's not over-dramatize.

cquiroz: I beg to disagree. Many users reuse their passwords, even if the ODB isn't perhaps that critical, they could reuse their passwords in other more valuable settings

jdnavarro: We are not a bank but most users reuse passwords so we have the responsibility of not leaking them out.

But there is no drama in those password storing algorithms, they are straightfoward to use, just like hashing, take care of everything for you, and you can have peace of mind without having to think about it.

tpolecat: I guess that's a reasonable point. Are we happy with S2K?

cquiroz: I think that maybe enough. We can do it later though, just let's not forget it 😄 
The password business is quite messy that's why I'd be happy to delegate it to ITS/github/google or whoever wants it

tpolecat: Actually S2K isn't a hashing algorithm. `crypt-bf/10` takes a few hundred ms on my machine and we can up the count arbitrarily as machines get faster so how does that sound?

cquiroz: From what I knew `hmacs` are recommended but the documentation on postgress suggest to use something like `crypt-bf/10` so I'd follow that advice. `crypt-bf/8` says it would take 3927y to brute force an 8 char password. That sounds more than enough

jdnavarro: If I were working on a bank I guess I would still keep nitpicking about the algorithm but I will restrain myself 😄 , so 👍 for `crypt-bf/10`.

I'd say any algorithm designed purposely designed to store passwords with the proper reputation should be good enough for us. So S2K it's perfectly fine since it's the recommended by PostgreSQL.

In any case all these functions to store passwords and its parameters should be quite easy to swap in and out at any moment so we don't need to get it perfectly right now.

Can we assume all users stored in the database will have a Google account? All Gemini personnel will have one. If that's the case, we could delegate all the authentication to Google, as @cquiroz suggested, and then avoid storing any passwords. Nowadays it's very easy to use all those authentication services, but if we need to support users without Google accounts (is there anyone in the planet without a Google account? 😄 ) then it may not be worth implementing both ways of authenticating users.

PR #91, opened 2017-07-24
Add GMOS-S Smart Gcal

swalker2m:
Adds GMOS-S Smart Gcal support, including import from OCS2.  It's kind of copy/pasty, unfortunately.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
+++ b/modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
@@ -8,6 +8,7 @@ import gem.SmartGcal._
 import gem.config._
 import gem.config.DynamicConfig.{ SmartGcalKey, SmartGcalSearchKey }
 import gem.enum._
+import gem.math.Wavelength
 import doobie.imports._
 
 import scalaz._
@@ -21,6 +22,7 @@ object SmartGcalDao {
       ids <- k match {
                case f2: SmartGcalKey.F2              => selectF2(f2, t)
                case gn: SmartGcalKey.GmosNorthSearch => selectGmosNorth(gn, t)
+               case gs: SmartGcalKey.GmosSouthSearch => selectGmosSouth(gs, t)
              }
       gcs <- ids.traverseU { GcalDao.selectSmartGcal }.map(_.collect { case Some(a) => a })
     } yield gcs
@@ -30,6 +32,9 @@ object SmartGcalDao {
 
   def selectGmosNorth(k: SmartGcalKey.GmosNorthSearch, t: SmartGcalType): ConnectionIO[List[Int]] =
     t.fold(Statements.selectGmosNorthByLamp(k), Statements.selectGmosNorthByBaseline(k)).list
+
+  def selectGmosSouth(k: SmartGcalKey.GmosSouthSearch, t: SmartGcalType): ConnectionIO[List[Int]] =
+    t.fold(Statements.selectGmosSouthByLamp(k), Statements.selectGmosSouthByBaseline(k)).list
 
   val createIndexF2: ConnectionIO[Int] =
     Statements.createIndexF2.run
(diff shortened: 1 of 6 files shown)


modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -31,6 +33,9 @@ object SmartGcalDao {
   def selectGmosNorth(k: SmartGcalKey.GmosNorthSearch, t: SmartGcalType): ConnectionIO[List[Int]] =
     t.fold(Statements.selectGmosNorthByLamp(k), Statements.selectGmosNorthByBaseline(k)).list
 
+  def selectGmosSouth(k: SmartGcalKey.GmosSouthSearch, t: SmartGcalType): ConnectionIO[List[Int]] =
+    t.fold(Statements.selectGmosSouthByLamp(k), Statements.selectGmosSouthByBaseline(k)).list
```
cquiroz: Got a bit lost here. Why can you call `fold` on `SmartGcalType`?

modules/ocs2/src/main/scala/gem/ocs2/SmartGcalImporter.scala
```diff
@@ -80,12 +81,33 @@ object SmartGcalImporter extends TaskApp with DoobieClient {
...
 
-    (SmartGcalKey.GmosNorthDefinition(c, (wmin, wmax)), gcal)
+    (SmartGcalKey.GmosDefinition(c, (wmin, wmax)), gcal)
+  }
+
+  def parseGmosSouth(input: List[String]): (SmartGcalKey.GmosSouthDefinition, List[String]) = {
+    import Parsers.Gmos._
+    import Parsers.GmosSouth._
+
+    val disperserS :: filterS :: fpuS :: xBinS :: yBinS :: _ :: ampGainS :: wavelengthMinS :: wavelengthMaxS :: gcal = input
```
cquiroz: What is this? destructuring a list? What if the length doesn't match?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
swalker2m: There are two basic kinds of SmartGcal queries.  One looking for just a matching Arc or Flat specifically (by lamp) and one looking for day vs. night calibration sets.  The fold just picks the family of query that we will do.

modules/ocs2/src/main/scala/gem/ocs2/SmartGcalImporter.scala
swalker2m: Yeah it's definitely fragile, but this is only used for the initial import of string `.csv` GCAL data.  Once the data is in tables we won't need the importer anymore, I think.  The data is not edited by users but rather written out to a known format by the `exportSmartGcal` OSGi shell command I added to the ODB.  If anything goes wrong the import blows up, but that's fine I claim because this code won't be used by end-users.

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
cquiroz: How do you do that? is explicitly defined or is it defined as an Either? other?

swalker2m: I think I misread your question.  The `fold` method comes in via an enrichment method on the generated enum:

```scala
  /**
   * Enrichment methods for [[SmartGcalType]].
   * @group Enrichments
   */
  implicit class SmartGcalTypeOps(val value: SmartGcalType) extends AnyVal {
    def fold[X](lamp: GcalLampType => X, baseline: GcalBaselineType => X): X =
      value match {
        case SmartGcalType.Arc           => lamp(GcalLampType.Arc)
        case SmartGcalType.Flat          => lamp(GcalLampType.Flat)
        case SmartGcalType.DayBaseline   => baseline(GcalBaselineType.Day)
        case SmartGcalType.NightBaseline => baseline(GcalBaselineType.Night)
      }
  }
```

Issue #71, opened 2017-07-11
intermittent test failure with Daily program ids


tpolecat:
Something isn't quite right.

```
[info] Daily
[info] - should reparse
[info] - should never reparse into a Nonstandard, even if we try
[info] - should find the correct observing day given a site and instant
[info] - should be consistent re: .start and .fromSiteAndInstant
[info] - should be consistent re: .end and .fromSiteAndInstant *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: Daily(GS,ENG,2016-08-15) did not equal Daily(GS,ENG,2016-08-14)
[info]     Location: (ProgramIdSpec.scala:60)
[info]     Occurred when passed generated values (
[info]       arg0 = Daily(GS,ENG,2016-08-14)
[info]     )
```

swalker2m:
Fixed by PR #93.

PR #93, opened 2017-07-24
Fix daily program id bug

swalker2m:
There is an [issue](https://github.com/gemini-hlsw/gem/issues/71) in the way that `Daily` program ids start and end times are calculated.  It doesn't account for daylight savings.  For example, at GS daylight savings starts and ends on a random, different day each year.  In the case of GS-ENG20160814, the old method for calculating the end date would take the last nanosecond of 8/13

```
2016-08-13T23:59:59.999999999-04:00[America/Santiago]
```

and add 14 hours, which usually produces the nanosecond before 2 PM on 8/14.  Because of the daylight savings event though we instead get:

```
2016-08-14T14:59:59.999999999-03:00[America/Santiago]
```

This update just makes the start and end times more explicit.


cquiroz:
very nice

PR #94, opened 2017-07-25
Add semester to daily program id

swalker2m:
I noticed that the `Daily` program id was missing its `Semester`.  I imagine there is a reason but I wasn't sure why so I thought I'd open a PR to find out.  At any rate, we associate semesters with daily program ids in the old core `ProgramId` and make use of it when filtering programs in the OT open dialog.

### DIFF
--- a/modules/core/src/main/scala/gem/ProgramId.scala
+++ b/modules/core/src/main/scala/gem/ProgramId.scala
@@ -90,7 +90,7 @@ object ProgramId {
     localDate:        LocalDate
   ) extends ProgramId(
     Some(site),
-    None,
+    Some(Semester.fromLocalDate(localDate)),
     Some(dailyProgramType.toProgramType)
   ) {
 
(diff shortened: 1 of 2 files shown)


tpolecat:
Ah, I didn't know that was necessary. Sounds good. Can we turn it into a data member? We will ned to do the `sealed abstract case class` trick to ensure that the semester and date are consistent. Also need to note in the comments that the semester is inferred.

tpolecat:
And I suppose add a test to confirm that the date+site is always consistent with the semester?

swalker2m:
I'm inclined to just calculate it on demand instead of storing it or trying to make sure it is consistent.  Does that make sense:

```scala
    /** Semester inferred from the `localDate` associated with this id. */
    def semester: Semester =
      Semester.fromLocalDate(localDate)
```

swalker2m:
Or maybe we want to pattern match on it?

tpolecat:
We could turn the `*Option` members into abstract methods and require the subclasses to implement them, which would let us compute the value on demand without needing to store the extra member and do the consistency nonsense. I think that might be the ethical thing to do here. I'm not super worried about pattern-matching.

swalker2m:
> We could turn the `*Option` members into abstract methods ...

Makes sense to me.  I took a stab at that.  I don't think there's really anything here to add to the test cases?

tpolecat:
> I don't think there's really anything here to add to the test cases?

Maybe a sanity check that the semester and date are consistent, but otherwise no.

PR #96, opened 2017-07-25
update user password hashing, add tests

tpolecat:
This PR updates password hashing `bcrypt/10` and builds out the `UserDao` enough to test.

Resolves #96

### DIFF
--- a/modules/db/src/main/scala/gem/dao/UserDao.scala
+++ b/modules/db/src/main/scala/gem/dao/UserDao.scala
@@ -49,6 +49,32 @@ object UserDao {
     Statements.selectRoles(id).list.map(_.foldMap { case (k, v) => Map((k -> Set(v))) })
 
   /**
+   * Insert or update a program role.
+   * @group Updates
+   */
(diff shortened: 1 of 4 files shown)


modules/db/src/main/scala/gem/dao/UserDao.scala
```diff
@@ -81,11 +110,39 @@ object UserDao {
       """.query
 
     def changePassword(uid: User.Id, oldPassword: String, newPassword: String): Update0 =
-      sql"""
+      (fr"""
         UPDATE gem_user
-        SET    md5 = md5($newPassword)
-        WHERE  id  = ${uid}
-        AND    md5 = md5($oldPassword)
+        SET    hash = crypt($newPassword, gen_salt('bf', 10))
```
jdnavarro: Nice, I didn't realize `pgcrypto` already included the `bcrypt` algorithm. I wonder why they don't highlight it in the documentation.

cquiroz:
Looks great and a nice collection of tests 👍

PR #97, opened 2017-07-27
WIP - gem webserver

tpolecat:
This upgrades http4s to 0.16 and makes a few changes to the `Service` interface, but is mostly a new sub-project that adds a web server. All you can do now is log in and do a text search for programs. Copied many ideas from ocs3, thanks @cquiroz for hints.


modules/json/src/main/scala/gem/json.scala
```diff
@@ -66,4 +66,17 @@ package object json {
   // Codec for GcalArcs
   implicit def gcalArcsCodec: CodecJson[GcalArcs] =
     CodecJson.derived[OneAnd[ISet, GcalArc]].xmap(oa => GcalArcs(oa.head, oa.tail.toList))(_.arcs)
+
+  // Codec for role maps
+  implicit def programIdKeyedMapCodec[A: CodecJson]: CodecJson[Map[Program.Id, A]] =
+    CodecJson.derived[Map[String, A]].xmap(
```
cquiroz: neat

modules/web/src/main/scala/Application.scala
```diff
@@ -0,0 +1,49 @@
...
+  def globToSql(s: String): String =
+    s.replaceAll("\\*", "%")
+     .replaceAll("\\.", "?")
+
+  /** Gem application endpoints. */
+  def service: AuthedService[GemService[Task]] =
+    AuthedService {
+
+      // Select matching program ids and titles.
+      case GET -> Root / "api" / "query" / "program" :? Query(q) +& Limit(n) as gs =>
```
cquiroz: What's the role of `Limit`?

modules/web/src/main/scala/Gatekeeper.scala
```diff
@@ -0,0 +1,112 @@
...
+     */
+    def encode(env: Environment, user: User[_]): Task[String] =
+      create(user, env.config.jwt.ttlSeconds).map(env.encodeJwt(_))
+
+    /**
+     * Like `encode`, but embeds the encoded JWT in a cooke with environment-specified
+     * name.
+     */
+    def cookie(env: Environment, user: User[_]): Task[Cookie] =
+      encode(env, user).map(Cookie(env.config.jwt.cookieName, _))
```
cquiroz: You may want to mark the cookie `httpOnly` and `secure` when doing `https`

modules/web/src/main/scala/package.scala
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+  implicit class QueryParamDecoderOps[A](fa: QueryParamDecoder[A]) {
+    def matcher(key: String): QueryParamDecoderMatcher[A] =
+      new QueryParamDecoderMatcher[A](key)(fa) {}
+    def optMatcher(key: String): OptionalQueryParamDecoderMatcher[A] =
+      new OptionalQueryParamDecoderMatcher[A](key)(fa) {}
+  }
+
+  implicit class RequestOps(req: Request) {
```
cquiroz: `AnyVal`

modules/web/src/main/scala/Gatekeeper.scala
```diff
@@ -0,0 +1,112 @@
...
+          }
+        }
+
+      // curl -i -b gem.jwt=... localhost:8080/something/else
+      case req =>
+        GemToken.decodeFromCookie(env, req).flatMap {
+          case Left(msg)  => Forbidden(msg)
+          case Right(jwt) =>
+            env.service(jwt.uid).flatMap {
+              case None      => Forbidden(s"JWT is valid but user ${jwt.uid} was not found.")
```
cquiroz: If you want to be picky, I suppose this leaks some information. OTOH hand we constructed the cookie with the private key so an attacker should not be able to make one

modules/web/src/main/scala/Application.scala
tpolecat: It says "only show me the first `n` matches" … the UI should limit the search because there's no point showing thousands of matches.

modules/web/src/main/scala/Gatekeeper.scala
tpolecat: 👍  ok I'll circle back to that when we set up https.

modules/web/src/main/scala/package.scala
tpolecat: 👍

PR #98, opened 2017-07-31
Port of core to support scala.js

cquiroz:
After several attempts I think I managed to get `gem/core` to compile to js using the typelevel compiler. 

There is one test case failing on js and I'm ignoring it but should be reviewed

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,7 @@ lazy val http4sVersion            = "0.16.0a-M3"
 lazy val scalaXmlVerson           = "1.0.6"
 lazy val scalaParsersVersion      = "1.0.4"
 lazy val tucoVersion              = "0.1.1"
-lazy val attoVersion              = "0.5.2"
+lazy val attoVersion              = "0.5.3"
 lazy val slf4jVersion             = "1.7.25"
 lazy val jwtVersion               = "0.14.0"
 
@@ -51,10 +51,13 @@ organizationName in ThisBuild := "Association of Universities for Research in As
 startYear        in ThisBuild := Some(2017)
 licenses         in ThisBuild += ("BSD-3-Clause", new URL("https://opensource.org/licenses/BSD-3-Clause"))
 
-lazy val testLibs = Seq(
-  "org.scalatest"  %% "scalatest"  % scalaTestVersion  % "test",
-  "org.scalacheck" %% "scalacheck" % scalaCheckVersion % "test"
-)
+// Make JS tests run fine on travis
+parallelExecution in (ThisBuild, Test) := false
+
+lazy val testLibs = Def.setting(Seq(
+  "org.scalatest"  %%% "scalatest"  % scalaTestVersion  % "test",
+  "org.scalacheck" %%% "scalacheck" % scalaCheckVersion % "test"
+))
 
 lazy val gemWarts =
   Warts.allBut(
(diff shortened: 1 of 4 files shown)


modules/core/src/test/scala/gem/LocationSpec.scala
```diff
@@ -70,7 +70,7 @@ class LocationSpec extends FlatSpec with Matchers with PropertyChecks with Arbit
     }
   }
 
-  it should "produce a sorted list of Location.Middle" in {
+  ignore should "produce a sorted list of Location.Middle" in {
```
cquiroz: This is failing with a `Maximum call stack size exceeded` error on JS. The stack size is smaller on JS so it produces this error. Any ideas to change `find` to make it stack safe? Otherwise we could make `count` smaller

swalker2m: The recursive function in `find` is `@tailrec` but I take it that doesn't help JavaScript?

cquiroz: TBH I don't know

swalker2m: Maybe it's the `foldRight` that is used in `toBase10`.

tpolecat: The tail call rewrite is independent of the runtime … it happens way before you know whether you're targeting JVM or JS. So this is suspicious.

swalker2m: I suspect the `foldRight` in `toBase10`.  Switching to `reverse` first and `foldLeft` may work, I guess.  Even if it does work, there may be better solutions.

```scala
loc.positions.take(len).reverse.foldLeft((Zero, One)) { case ((acc, pow), i) =>
```

build.sbt
```diff
@@ -164,27 +167,41 @@ lazy val flywaySettings = Seq(
...
+      "org.scalaz"   %%% "scalaz-core"          % scalazVersion,
+      "com.chuusai"  %%% "shapeless"            % shapelessVersion,
+      "org.tpolecat" %%% "atto-core"            % attoVersion,
+      "org.tpolecat" %%% "atto-compat-scalaz72" % attoVersion
     ),
     sourceGenerators in Compile +=
       Def.task { gen2((sourceManaged in Compile).value / "gem").unsafePerformIO }.taskValue
   )
+  .jsSettings(
+    // Skip using the typelevel compiler until interoperability is improved
```
cquiroz: I should remove that line 😄

modules/core/src/test/scala/gem/LocationSpec.scala
cquiroz: The change from `foldLeft` to `foldRight` doesn't fix the test

swalker2m: Looking at it closely, there are a number of `IList` methods that aren't stack safe.  If you want to merge this with the test case ignored I can work on making it work with JS.

cquiroz:
I'm considering removing a`coreJS` from the aggregate so you can develop without the increased compilation times but keep it on travis to alert you if you break the js side

modules/core/src/test/scala/gem/LocationSpec.scala
tpolecat: Hm, every method on `IList` should be stacksafe. Still suspicious.

swalker2m: If we're otherwise okay on this PR, I'll work on it after it is merged.

PR #100, opened 2017-08-03
Add Support for GMOS Nod and Shuffle

swalker2m:
GMOS nod and shuffle support.  I ended up creating a separate `gmos_nod_and_shuffle` table because it seemed offensive to add all the N&S columns to both `static_gmos_north` and `static_gmos_south`.


modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -67,77 +64,172 @@ object StaticConfigDao {
...
+          SELECT a_offset_p,
+                 a_offset_q,
+                 b_offset_p,
+                 b_offset_q,
+                 e_offset,
+                 offset_rows,
+                 cycles
+            FROM gmos_nod_and_shuffle
+           WHERE static_id = $sid AND instrument = $i
+        """.query[GmosNodAndShuffle]
```
swalker2m: I could join `gmos_nod_and_shuffle` when selecting the static config instead.  This seemed simple though.  Opinions?

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -67,77 +64,172 @@ object StaticConfigDao {
...
+
+    import gem.config.Gmos.GmosNodAndShuffle
+    import gem.enum.Instrument.{ GmosN, GmosS }
+    import StaticConfig.{ GmosNorth, GmosSouth }
+
+    def insertNorth(sid: Int, gn: GmosNorth): ConnectionIO[Int] =
+      for {
+        i <- insertNodAndShuffle(sid, GmosN, gn.common.nodAndShuffle)
+        j <- Statements.Gmos.insertNorth(sid, gn).run
+      } yield i + j
```
swalker2m: Is this result correct?  The number of rows inserted?

tpolecat:
I will have a look in a bit, sorry, tied up.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -4,19 +4,108 @@
...
+  }
+
+  // TODO: there are many ways to misconfigure Nod And Shuffle.  Some of these
+  // ways can be caught in a companion object constructor, and some cannot, or
+  // at least cannot easily.  In the first category, we should definitly check
+  // whether posA and posB are close enough together (< 2 arcsecs) to allow
+  // e-offsetting.  In the second category, shuffle offset in detector rows has
+  // to be a multiple of y-binning.  Worse, y-binning is part of the dynamic
+  // configuration so it isn't clear what happens if the shuffle offset isn't
+  // always a multiple of y-binning.
```
tpolecat: This seems like the kind of consistency check we will need to run after the fact, and report it as user error.

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
tpolecat: It's not really clear what we should return from updates that affect many tables. The value doesn't really mean anything so it might be better as `Unit`.

swalker2m: Ok, that makes sense.  I'll fix it in the next PR.

PR #101, opened 2017-08-03
DAO Cleanup

swalker2m:
A few minor adjustments:

* In `StepDao`, partition the instrument-specific statements into their own modules to match `StaticConfigDao`.

* In `StepDao` and `StaticConfigDao`, void the result of the `insert` statements since the integer returned isn't useful.

* Remove some unused code in `StepDao`.


modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -535,6 +305,207 @@ object StepDao {
...
+            w <- wavelength.flatMap(Wavelength.fromAngstroms)
+          } yield GmosGrating(d, o, w)
+      }
+
+      final case class GmosFpuBuilder[U](
+                         mdfFileName: Option[String],
+                         customSlitWidth: Option[GmosCustomSlitWidth],
+                         builtin: Option[U]) {
+
+        val toFpu: Option[GmosCustomMask \/ U] = {
```
cquiroz: Since we are on 2.12 shouldn't we use `Either`?

tpolecat: `EitherT` uses `\/` so we're arguably better off with `\/` until we switch to cats. It will all get resolved soonish, I hope.

PR #102, opened 2017-08-03
ra/dec

tpolecat:
This adds data types for RA and Dec.

### DIFF
--- a/modules/core/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/src/main/scala/gem/config/DynamicConfig.scala
@@ -23,9 +23,6 @@ sealed abstract class DynamicConfig extends Product with Serializable {
   /** Obtains the smart gcal search key that corresponds to the instrument
     * configuration, if any. This key can be used to find the matching gcal
     * configuration.
-    *
-    * @see [[gem.dao.SmartGcalDao]]
-    *
     * @return corresponding smart gcal search key, if any
     */
   def smartGcalKey: Option[DynamicConfig.SmartGcalSearchKey] =
--- a/modules/core/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/src/main/scala/gem/math/Angle.scala
@@ -31,8 +31,10 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
     Angle.fromMicroarcseconds(-toMicroarcseconds.toLong)
 
   /**
-   * Mirror this agle around the axis defined by `a`. This operation is specified completely by the
-   * identity `b - a = (a mirrorBy b) - b`.
+   * Mirror image of this angle, when the mirror stands at angle `a`; or picture picking up the
+   * circle and flipping it over, around a line drawn from the center going off in direction `a`.
+   * So `(88° mirrorBy 90°) = 92°` for instance, as is `88° mirrorBy 270°` since it's the same
+   * line. This operation is specified completely by the identity `b - a = (a mirrorBy b) - b`.
    */
   def mirrorBy(a: Angle): Angle = {
     val Δ = a.toMicroarcseconds - toMicroarcseconds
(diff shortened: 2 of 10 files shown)


modules/core/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -23,9 +23,6 @@ sealed abstract class DynamicConfig extends Product with Serializable {
   /** Obtains the smart gcal search key that corresponds to the instrument
     * configuration, if any. This key can be used to find the matching gcal
     * configuration.
-    *
-    * @see [[gem.dao.SmartGcalDao]]
-    *
```
tpolecat: This caused a Scaladoc error because this type is in another module.

modules/core/src/main/scala/gem/math/Declination.scala
```diff
@@ -0,0 +1,101 @@
...
+   */
+  def unsafeFromAngle(a: Angle): Declination =
+    fromAngle(a).getOrElse(sys.error(s"Declination out of range: $a"))
+
+  /**
+   * Declinations are ordered from north to south.
+   * @group Typeclass Instances
+   */
+  implicit val DeclinationOrder: Order[Declination] =
+    Order.orderBy(_.toAngle.toSignedMicroarcseconds)
```
swalker2m: It seems counter-intuitive that 90 comes before 0 before -90.  Perhaps we should flip it?

modules/core/src/main/scala/gem/math/Declination.scala
```diff
@@ -0,0 +1,101 @@
...
+    val dms = toAngle.toDMS; import dms._
+    val signedDegrees = if (degrees > 180) degrees - 360 else degrees
+    f"Dec($signedDegrees:$arcminutes%02d:$arcseconds%02d.$milliarcseconds%03d$microarcseconds%03d)"
+  }
+
+  final override def equals(a: Any) =
+    a match {
+      case dec: Declination => dec.toAngle === this.toAngle
+      case _                => false
+    }
```
swalker2m: Why wouldn't the generated `equals` and `hashCode` work?

modules/core/src/main/scala/gem/math/Declination.scala
tpolecat: Oh I thought this would order them from -90 .. 90. Did I get it backwards?

modules/core/src/main/scala/gem/math/Declination.scala
tpolecat: Yes, good catch.

swalker2m:
We'll need JSON codecs for these, I assume, and `Meta` instances if they are stored in the database.

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -20,14 +20,25 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
...
     Angle.fromMicroarcseconds(-toMicroarcseconds.toLong)
 
+  /**
+   * Mirror this agle around the axis defined by `a`. This operation is specified completely by the
+   * identity `b - a = (a mirrorBy b) - b`.
+   */
+  def mirrorBy(a: Angle): Angle = {
+    val Δ = a.toMicroarcseconds - toMicroarcseconds
+    Angle.fromMicroarcseconds(toMicroarcseconds + Δ * 2L)
+  }
```
tpolecat: I'll embiggen the comment.

modules/core/src/main/scala/gem/math/Declination.scala
swalker2m: Sorry, I think you're right.  I missed the `Signed` part of `toSignedMicroarcseconds`.  The comment seems wrong then?  `Declinations are ordered from north to south.`

tpolecat:
> We'll need JSON codecs for these, I assume, and Meta instances if they are stored in the database.

I have been doing these on-demand … we can add as needed. I'm not sure what the wire format and the database representation should look like yet.

modules/core/src/main/scala/gem/math/Declination.scala
tpolecat: Doh!

tpolecat:
Ok I made the changes suggested above.

PR #104, opened 2017-08-04
Improve enum generation, move into top-level project

tpolecat:
This moves the enum generation into the `sql` project and cleans it up a bit. It also randomly updates sbt, Scala, argonaut, and argonaut-shapeless since I happened to notice they needed updating. This is the content of the first commit.

The second commit is huge and just includes the generated files. I only separated them to make it easier to review.


tpolecat:
Well that's a weird failure. Hm.

tpolecat:
Aha. Flyway was only working because we used to have doobie in the build, which was pulling in the Postgres driver. Adding it manually fixed it. Also updated Flyway because why not?

cquiroz:
I like how often random and sbt go together

PR #105, opened 2017-08-07
Coordinates

tpolecat:
This adds `Coordinates` and some operations that require dropping into floating point.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Coordinates.scala
+++ b/modules/core/src/main/scala/gem/math/Coordinates.scala
@@ -10,7 +10,7 @@ import scalaz._, Scalaz._
 final case class Coordinates(ra: RightAscension, dec: Declination) {
 
   /**
-   * Offset this `Coordinates` by the given deltas, and indicate whether the declination crossed
+   * Offset these `Coordinates` by the given deltas, and indicate whether the declination crossed
    * a pole; if so the right ascension will have been flipped 180°.
    * @group Operations
    */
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/gem/math/Coordinates.scala
```diff
@@ -0,0 +1,99 @@
...
+   */
+  def interpolate(that: Coordinates, f: Double): Coordinates = {
+    val δ = angularDistance(that).toDoubleRadians
+    if (δ === 0) this
+    else {
+      val φ1 = this.dec.toAngle.toDoubleRadians
+      val φ2 = that.dec.toAngle.toDoubleRadians
+      val λ1 = this.ra.toAngle.toDoubleRadians
+      val λ2 = that.ra.toAngle.toDoubleRadians
+      val a  = sin((1 - f) * δ) / sin(δ) // n.b. this line is wrong on the web page
```
cquiroz: i suppose this will be undefined at delta pi. Should this be handled here  or is the angular distance never 180 degrees?

tpolecat: Well, same at 0° … the math works out since `Double` never overflows and `atan2` understands infinity I guess. I haven't really thought about how this works. 

```scala
scala> import gem.math.Coordinates._
import gem.math.Coordinates._
 
scala> Zero.angularDistance(Zero).toDoubleRadians
res2: Double = 0.0
 
scala> Zero.interpolate(Zero, 0)
res3: gem.math.Coordinates = Coordinates(RA(0:00:00.000000),Dec(0:00:00.000000))
 
scala> SouthPole.angularDistance(NorthPole).toDoubleRadians
res4: Double = 3.141592653589793
 
scala> SouthPole.interpolate(NorthPole, 0.5)
res5: gem.math.Coordinates = Coordinates(RA(12:00:00.000000),Dec(0:00:00.000000))
```

PR #106, opened 2017-08-07
Add GMOS ROI

swalker2m:
Adds GMOS region of interest support.  Only one custom ROI definition can be used with an observation, but individual steps may choose to use the custom ROI or one of the builtin ROI options.  Accordingly support for ROI is distributed across the static and dynamic configurations:

* The static configuration contains a `List[GmosCustomRoiEntry]` that defines the regions of interest for the steps in which a `Custom` `GmosRoi` is selected.

* The dynamic configuration contains a new `GmosRoi` entry which is one of the builtin options or else `Custom`.

Obviously this presents an opportunity to configure a custom ROI in the static configuration and never use it in any step, or worse specify a `Custom` ROI in the dynamic steps and never define it in the static configuration.  This will have to be caught and flagged by an observation check just as we do today in OCS2.

I defined logic for detecting overlapping ROI entries but am not using it yet.  When we add support for checking nod and shuffle e-offsetting distances I think we may want to guard construction of `GmosCommonStaticConfig` with these types of checks.


modules/core/src/main/scala/gem/enum/GmosRoi.scala
```diff
@@ -0,0 +1,51 @@
...
+package enum
+
+import scalaz.syntax.equal._
+import scalaz.std.string._
+
+/**
+ * Enumerated type for GMOS ROI (region of interest).
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GmosRoi(
```
swalker2m: This is just renamed from `GmosBuiltinRoi` since it now includes a `Custom` option.  I removed the specific ROI descriptions because we don't use them in OCS2.

modules/core/src/test/scala/gem/arb/Offset.scala
```diff
@@ -9,13 +9,15 @@ import org.scalacheck._
 import org.scalacheck.Arbitrary._
 
 trait ArbOffset {
-  import ArbAngle._
-
   implicit val arbOffsetP: Arbitrary[Offset.P] =
-    Arbitrary(arbitrary[Angle].map(Offset.P(_)))
+    Arbitrary(
+      Gen.chooseNum(0, 10000).map(mas => Offset.P(Angle.fromMilliarcseconds(mas)))
+    )
```
swalker2m: This is unfortunate, but necessary since super large offsets don't fit in the `NUMERIC(9,3)` field we use for offsets.  We may want to revisit this because we can't test with negative offsets as it stands?

modules/ocs2/src/main/scala/gem/ocs2/SequenceDecoder.scala
```diff
@@ -19,106 +19,12 @@ import Scalaz._
...
-      f <- Filter.parse(cm)
-      u <- Fpu.cparse(cm).map(_.flatten)
-      m <- gmosCustomMask(cm)
-      fpu = u.map(_.right[Gmos.GmosCustomMask]) orElse m.map(_.left[GmosSouthFpu])
-    } yield DynamicConfig.GmosSouth(c, g, f, fpu)
-  }
+  def decode(n: Node): PioError \/ List[Step[DynamicConfig]] =
+    (n \ "step").toList.scanLeft(EmptyConfigMap) { (m, stepNode) =>
+      stepNode.addStepConfig(m)
+    }.drop(1).traverseU(parseStep)
```
swalker2m: Here I moved some code around and grouped GMOS-related functions together.  Sorry, the change looks more massive than it actually was.

modules/core/src/test/scala/gem/arb/Offset.scala
tpolecat: Offsets are stored in the database as signed arcseconds, so small negatives angles should also work.

modules/core/src/main/scala/gem/config/Gmos.scala
```diff
@@ -104,26 +112,93 @@ object Gmos {
...
+
+  /** GMOS custom ROI entry definition. */
+  sealed abstract case class GmosCustomRoiEntry(
+    xMin:   Short,
+    yMin:   Short,
+    xRange: Short,
+    yRange: Short
+  ) {
+
+    // Enforced by fromDescription constructor
```
tpolecat: I think these sanity checks are probably worthwhile, even though they're ugly.

tpolecat:
Something we might consider is an ROI type that's a bag of rectangles that coalesce when a rectangle is overlapped completely. This data type would be a bounded semilattice, which is a set-like monoid where `|+|` is commutative and idempotent.

PR #107, opened 2017-08-08
Fix GMOS filter wavelength type

swalker2m:
This PR addresses an egregious mistake in defining the GMOS filter wavelength type.  In particular I declared these as `Angle`s originally when of course they are `Wavelength` values.  I will update F2 in a future PR since it still uses raw `Double` wavelength values.

I had thought I would need to differentiate between μm and nm but it turns out that both GMOS and F2 specify filter wavelengths in μm, despite GMOS's central wavelength in nm.  Nevertheless it seems useful to be able to capture the conversion from either μm or nm values so I left it there.


swalker2m:
For reference, here is a snippet of the `e_gmos_north_filter` definition:

```
CREATE TABLE e_gmos_north_filter (
    id         identifier            PRIMARY KEY,
    short_name character varying(10) NOT NULL,
    long_name  character varying(30) NOT NULL,
    wavelength numeric(4,3)          NOT NULL,
    obsolete   boolean               NOT NULL
);

COPY e_gmos_north_filter(id, short_name, long_name, wavelength, obsolete) FROM stdin;
GPrime	g	g_G0301	0.475	f
RPrime	r	r_G0303	0.630	f
IPrime	i	i_G0302	0.780	f
ZPrime	z	z_G0304	0.925	f
Z	Z	Z_G0322	0.876	f
Y	Y	Y_G0323	1.01	f
...
```

This reminds me that we need to go back and add comments describing the units of the columns where it isn't clear.

cquiroz:
I could take a look at libra given my squants history, I'll try to get them to run on scala.js too

tpolecat:
@cquiroz that would be fantastic if you have time

swalker2m:
Keep in mind I'm not proposing a general solution here, just a way to be able to read the wavelength value from the database into a `gem.math.Wavelength` with the proper value.  We still need Libra or something to represent the units in the model.

PR #111, opened 2017-08-09
F2/GMOS consistency

swalker2m:
This PR includes a few minor updates to make F2 and GMOS more consistent.  In particular:

* Instead of using special disperser and FPU `None` values, the configurations are updated to be `Option[F2Disperser]` etc. like we do for GMOS.

* Switched F2 wavelength value type from `Double` to `Wavelength`.

* Renamed the enumerated type `F2FpUnit` to `F2Fpu` to match `GmosNorthFpu` and `GmosSouthFpu`.

* Renamed `gem.config.Gmos` to `gem.config.GmosConfig` to match `gem.config.GcalConfig` and the new `gem.config.F2Config`.  Having the additional types for an instrument be named the same as the instrument causes a number of import clash issues.


modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -31,6 +31,9 @@ object EnumDef {
 
     implicit def caseOptionAngle [S <: Symbol] = at[(S, Option[Angle] ) ] { case (s, _) => "  val " + s.name + ": Option[gem.math.Angle]" }
     implicit def caseOptionDouble[S <: Symbol] = at[(S, Option[Double]) ] { case (s, _) => "  val " + s.name + ": Option[Double]" }
+
+    implicit def caesOptionWavelengthNm[S <: Symbol] = at[(S, Option[Wavelength.Nm])] { case (s, _) => s"  val ${s.name}: Option[gem.math.Wavelength]" }
```
cquiroz: case?

swalker2m: oops

PR #112, opened 2017-08-09
Update Gem to cats

tpolecat:
This updates gem to use Cats instead of scalaz. It's mostly mechanical. A few changes you will notice:
- Many of the library dependencies now have `-Mx` suffixes. They're all in a holding pattern awaiting cats 1.0.0 final and will go final when cats does.
- Everything uses `cats.effect.IO` … no more `Task`
- The old `|@|` syntax in scalaz is more general in cats, and is bedazzled onto tuple types, so instead of `(a |@| b)(f)` you can say `(a, b).mapN(f)`.
- scalaz collection types are now stdlib types, at least until we get a stable release of [dogs](https://github.com/stew/dogs), which may happens before this is merged. I don't think it's an urgent issue yet. In any case:

  - `ISet` is now `Set`. The element types in question all have well-behaved universal hashcode and equality.

  - `IList` is now `List`. This was only used because it was the backing store for scalaz's `NonEmptyList`, but cats uses `List` anyway so overall it's a simplification.

  - `==>>` is now `TreeSet`. The only use is with `Loc` keys, which have a reasonable stdlib `Ordering`, and the data type defines all the operations we need.

  - There is a tiny `gem.util.Lens` class which will hold us over until Monocle is available for cats (soon).


tpolecat:
Ok this is done I think. It's up to date and everything compiles and the tests pass. @swalker2m it would be helpful if you could pull this branch and try out the ocs2 importer and smart gcal stuff because there aren't any tests. GitHub tells me you can do it thus:

```
git checkout -b tpolecat-ctl-cats master
git pull https://github.com/tpolecat/gem.git ctl-cats
```

Right now we're on circe 0.8 which will be unreliable running against cats 0.10 but I think it's ok for now. I'll update when Travis is ready (he said a day or two). We're also on a very unstable http4s 0.18 snapshot. Hopefully there will be a milestone next week. In both cases it doesn't really matter because the web stuff is more a proof of concept than anything else right now.

tpolecat:
I will be off-grid from mid-day tomorrow until Sunday afternoon so there's no immediate urgency here, but it would be nice to not have to rebase again if I can avoid it. This is a lot of edits.

swalker2m:
I'll be away tomorrow anyway but nevertheless I'll hold off on making any changes until this is committed. I've been skimming through the commits and it looks great. I'll check out the importers on Monday though.

tpolecat:
Ok, sounds good. Enjoy your weekend.

swalker2m:
Anything we should worry about?

```
[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn] 
[warn] 	* co.fs2:fs2-core_2.12:0.10.0-M5 is selected over 0.10.0-M2
[warn] 	    +- co.fs2:fs2-scodec_2.12:0.10.0-M5                   (depends on 0.10.0-M5)
[warn] 	    +- org.tpolecat:doobie-free_2.12:0.5.0-M2             (depends on 0.10.0-M5)
[warn] 	    +- co.fs2:fs2-io_2.12:0.10.0-M5                       (depends on 0.10.0-M5)
[warn] 	    +- org.http4s:jawn-fs2_2.12:0.11.0-M1                 (depends on 0.10.0-M2)
[warn] 
[warn] 	* org.typelevel:cats-core_2.12:1.0.0-MF is selected over 0.9.0
[warn] 	    +- com.github.benhutchison:mouse_2.12:0.10-MF         (depends on 1.0.0-MF)
[warn] 	    +- org.tpolecat:atto-core_2.12:0.6.1-M1               (depends on 1.0.0-MF)
[warn] 	    +- org.typelevel:cats-free_2.12:1.0.0-MF ()           (depends on 1.0.0-MF)
[warn] 	    +- org.tpolecat:doobie-free_2.12:0.5.0-M2             (depends on 1.0.0-MF)
[warn] 	    +- org.typelevel:cats-effect_2.12:0.4                 (depends on 1.0.0-MF)
[warn] 	    +- gem-core:gem-core_2.12:20170814T124651             (depends on 1.0.0-MF)
[warn] 	    +- io.circe:circe-core_2.12:0.8.0 ()                  (depends on 0.9.0)
```

modules/ocs2/src/main/scala/gem/ocs2/SmartGcalImporter.scala
```diff
@@ -175,35 +174,40 @@ object SmartGcalImporter extends TaskApp with DoobieClient {
...
 
-    def lines(l: GcalLampType): Process[Task, SmartGcalLine[K]] =
-      scalaz.stream.io
-          .linesR(new File(dir, fileName(instFilePrefix, l)).getPath)
+    def lines(l: GcalLampType): Stream[IO, SmartGcalLine[K]] =
+      io.file.readAll[IO](new File(dir, fileName(instFilePrefix, l)).toPath, 4096)
+          .through(text.utf8Decode)
+          .through(text.lines)
           .map(_.split(',').map(_.trim).toList)
           .map(parseLine(_, l, parser))
```
swalker2m: For some reason an empty line is being inserted, which blows up the parser.  The parser doesn't try to handle anything other than exactly what is produced by the smart-gcal exporter.  An ad-hoc solution would be to filter them, I suppose, but I do wonder where the blank line comes from.  I assume it has to do with `text.lines`.

```
...
List(R3000, K_SHORT, LONGSLIT_8, 1, NIR balance, IR, Ar arc, Closed, 80000, 1, NIGHT)
List(R3000, J_LOW, LONGSLIT_4, 1, NIR balance, IR, Ar arc, Closed, 30000, 1, NIGHT)
List(R1200HK, HK, LONGSLIT_3, 1, NIR balance, IR, Ar arc, Closed, 90000, 1, NIGHT)
List(R3000, J, LONGSLIT_3, 1, NIR balance, IR, Ar arc, Closed, 30000, 1, NIGHT)
List()
[error] (run-main-2) scala.MatchError: List() (of class scala.collection.immutable.$colon$colon)
scala.MatchError: List() (of class scala.collection.immutable.$colon$colon)
	at gem.ocs2.SmartGcalImporter$.parseF2(SmartGcalImporter.scala:63)
...
```

tpolecat:
> Anything we should worry about?

The second eviction warnins is due to circe not being on 1.0.0-MF yet, which Travis assures me will be fixed soon. The first warning is due to an adapter layer for http4s, looks like. I'll ass Ross about that.

modules/ocs2/src/main/scala/gem/ocs2/SmartGcalImporter.scala
tpolecat: Ok I'll follow up with the fs2 people.

modules/core/src/main/scala/gem/Dataset.scala
```diff
@@ -35,7 +36,7 @@ object Dataset {
         case -1 => None
         case  n =>
           val (a, b) = s.splitAt(n)
-          b.drop(1).parseInt.toOption.flatMap { n =>
+          MonadError[Either[Throwable, ?], Throwable].catchNonFatal(b.drop(1).toInt).toOption.flatMap { n =>
             Observation.Id.fromString(a).map(oid => Dataset.Label(oid, n))
```
tpolecat: Yeah I want to PR those into mouse, which is where these things are ending up. I did this for expediency but we need to fix it.

modules/core/src/main/scala/gem/Dataset.scala
```diff
@@ -49,12 +50,12 @@ object Dataset {
...
      */
     implicit val LabelOrder: Order[Label] =
-      Order[Observation.Id].contramap[Label](_.observationId) |+|
+      Order[Observation.Id].contramap[Label](_.observationId) whenEqual
       Order[Int]           .contramap[Label](_.index)
 
     /** @group Typeclass Instances */
     implicit val LabelShow: Show[Label] =
-      Show.showA
+      Show.fromToString
```
tpolecat: Yeah the semigroup for `Eq` isn't implicit for some reason. I talked to Kai and we can't see a reason for this so it may be changed. But yeah in any case `.whenEqual` is fine.

I think the `.fromToString`, etc., are a bit nicer, yeah. Cats is more thoughtful about naming and much more consistent.

modules/core/src/main/scala/gem/Location.scala
```diff
@@ -96,11 +92,13 @@ object Location {
...
-    fi.foldRight(IList.empty[Int]) { (i, lst) =>
-      (lst.isEmpty && i === Int.MinValue) ? lst | i +: lst
-    } match {
-      case ICons(h, t) => new Middle(NonEmptyList.nel(h, t)) {}
-      case _           => Beginning
+    fi.foldRight(Eval.now(List.empty[Int])) { (i, elst) =>
+      elst.map { lst =>
+        if (lst.isEmpty && i === Int.MinValue) lst else (i :: lst)
+      }
+    }.value match {
```
tpolecat: Not sure what you mean, sorry. `Foldable.foldRight` wants an `Eval` as the accumulator. Cats does this rather than use by-name arguments. Can this be simplified?

modules/core/src/main/scala/gem/Observation.scala
```diff
@@ -42,22 +43,26 @@ object Observation {
...
   implicit val ObservationBitraverse: Bitraverse[Observation] =
     new Bitraverse[Observation] {
-      def bitraverseImpl[G[_]: Applicative, A, B, C, D](fab: Observation[A,B])(f: A => G[C], g: B => G[D]): G[Observation[C,D]] =
-        (f(fab.staticConfig) |@| fab.steps.traverse(g))((c, d) => fab.copy(staticConfig = c, steps = d))
+      def bifoldLeft[A, B, C](fab: Observation[A,B], c: C)(f: (C, A) => C, g: (C, B) => C): C =
+        fab.steps.foldLeft(f(c, fab.staticConfig))(g)
+      def bifoldRight[A, B, C](fab: Observation[A,B], c: Eval[C])(f: (A, Eval[C]) => Eval[C], g: (B, Eval[C]) => Eval[C]): Eval[C] =
+        fab.steps.foldRight(f(fab.staticConfig, c))(g)
+      def bitraverse[G[_]: Applicative, A, B, C, D](fab: Observation[A, B])(f: A => G[C], g: B => G[D]): G[Observation[C,D]] =
+        (f(fab.staticConfig), fab.steps.traverse(g)).mapN((c, d) => fab.copy(staticConfig = c, steps = d))
```
tpolecat: Yeah, cats doesn't provide default implementations of typeclass methods when the default is inefficient, as with `foldRight` in terms of `traverse` (which is implemented by traversing with `State[List[A], ?]`). I argued against this but lost. In practice it doesn't come up a lot.

modules/core/src/main/scala/gem/Dataset.scala
tpolecat: I see that they're in mouse already. Could have sworn I tried that earlier. I'll change them.

modules/core/src/main/scala/gem/Location.scala
swalker2m: No, thanks that explains it.  👍

modules/ocs2/src/main/scala/gem/ocs2/SmartGcalImporter.scala
swalker2m: Just to confirm, if I add `.filter(_.trim.nonEmpty)` after `.through(text.lines)` then I get to the next error (which is my fault -- I need to update some constraints on the `smart_f2` table).

tpolecat: Ok, shall I add that line and then call it good for now? It would be nice to add some tests for the importer but we need scrubbed data or some kind of encrypted archive.

swalker2m: 👍 for adding that line.

PR #113, opened 2017-08-14
Fix smart_f2 constraints

swalker2m:
F2 disperser and FPU are now optional, so the `smart_f2` constraints need to be adjusted to match.

(I'll rebase and merge this after the cats update is merged.)


swalker2m:
Rebased, ready to go.

PR #114, opened 2017-08-15
initial ephemeris

tpolecat:
This adds a minimal ephemeris data type, in preparation for the target model build-out.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Ephemeris.scala
+++ b/modules/core/src/main/scala/gem/math/Ephemeris.scala
@@ -1,0 +1,75 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.math
+
+import cats.{ Eq, Foldable, Monoid }
+import cats.implicits._
+import java.time.Instant
+import scala.collection.immutable.TreeMap
+
+/**
+ * Time-parameterized coordinates over a fixed interval, defined pairwise. Coordinates that fall
+ * between known instants are interpolated.
+ */
+sealed abstract case class Ephemeris private (toMap: TreeMap[Instant, Coordinates]) {
+  import Ephemeris.Element
+
+  // N.B. this case class is abstract and has a private ctor because we want to keep construction of
+  // the TreeMap private to ensure that the correct ordering is used and remains consistent.
+
+  /** First element in this ephemeris, if any. */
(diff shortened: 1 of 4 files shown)


tpolecat:
Rather than having the site be part of the ephemeris I will probably have a `Map[Site, Ephemeris]` on `Target.Nonsidereal`, or a distinct `Ephemerides` type that wraps such a map.

modules/core/src/main/scala/gem/math/Ephemeris.scala
```diff
@@ -0,0 +1,75 @@
...
+import cats.{ Eq, Foldable, Monoid }
+import cats.implicits._
+import java.time.Instant
+import scala.collection.immutable.TreeMap
+
+/**
+ * Time-parameterized coordinates over a fixed interval, defined pairwise. Coordinates that fall
+ * between known instants are interpolated.
+ */
+sealed abstract case class Ephemeris private (toMap: TreeMap[Instant, Coordinates]) {
```
cquiroz: Just a slight note. `TreeMap` is not available on scala.js

modules/core/src/main/scala/gem/math/Ephemeris.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  /** The sum of this ephemeris and `e`, taking values from `e` in the case of overlap. */
+  def ++(e: Ephemeris): Ephemeris =
+    new Ephemeris(toMap ++ e.toMap) {}
+
+}
+object Ephemeris {
+
+  /** An ephemeris element. */
+  type Element = (Instant, Coordinates)
```
cquiroz: `Element` is quite a generic name, maybe will get confused with e.g. DOM's `Element` or XML's `Element`

modules/core/src/main/scala/gem/math/Ephemeris.scala
tpolecat: Really? Why not? It's part of the standard library.

modules/core/src/main/scala/gem/math/Ephemeris.scala
tpolecat: Yeah I was thinking people would say `Ephemeris.Element` to be clear but maybe it could be something else.

swalker2m: I like the idea of qualifying the name with the module.

tpolecat:
Rebased, switched tests to `CatsSuite` and added law checking for `Eq` and `Monoid`.

PR #116, opened 2017-08-15
Switch to CatsSuite, add law checking

tpolecat:
This changes the `Angle` tests to use [`CatsSuite`](https://github.com/typelevel/cats/blob/6fefa30bfb85e90d193b53300b8313c4dd6c0201/testkit/src/main/scala/cats/tests/CatsSuite.scala), which provides some useful defaults and machinery, specifically for taking advantage of laws defined for cats typeclasses. The five cases in the Abelian group tests are now just

```scala
checkAll("Angle", GroupLaws[Angle].commutativeGroup)
```

This also makes gem consistent with testing in other Typelevel libs like http4s, which means there's one less thing to think about. I don't have strong feelings about it but it seems ok to me.

If this looks ok I will switch all the tests to use `CatsSuite`.

### DIFF
--- a/modules/core/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/src/main/scala/gem/math/Angle.scala
@@ -9,7 +9,7 @@ import cats.instances.long._
 import cats.syntax.eq._
 
 /**
- * Exact angles represented as integral microarcseconds. These values form an Abelian group over
+ * Exact angles represented as integral microarcseconds. These values form a commutative group over
  * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
  * integral microarcseconds correspond with clock microseconds (i.e., where they are evenly
  * divisible by 15 microarcseconds) is represented by the HourAgle subtype.
@@ -209,8 +209,8 @@ object Angle {
 
 
 /**
- * Exact hour angles represented as integral microseconds. These values form an Abelian group over
- * addition, where the inverse is reflection around the 0-12h axis. This is a subgroup of the
+ * Exact hour angles represented as integral microseconds. These values form a commutative group
+ * over addition, where the inverse is reflection around the 0-12h axis. This is a subgroup of the
  * integral Angles where microarcseconds are evenly divisible by 15.
  * @see The helpful [[https://en.wikipedia.org/wiki/Hour_angle Wikipedia]] article.
  */
--- a/modules/core/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/src/main/scala/gem/math/Angle.scala
@@ -3,7 +3,8 @@
 
 package gem.math
 
-import cats.{ Eq, Monoid, Show }
+import cats.{ Eq, Show }
+import cats.kernel.CommutativeGroup
 import cats.instances.long._
 import cats.syntax.eq._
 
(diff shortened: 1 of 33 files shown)


tpolecat:
Ah well since nobody is here to stop me I'll go ahead and do them all.

modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -143,11 +144,12 @@ object Angle {
   def fromDoubleRadians(rad: Double): Angle =
     fromDoubleDegrees(rad.toDegrees)
 
-  /** Angle is an Abelian group, but monoid is the best we can do for now. */
-  implicit val AngleMonoid: Monoid[Angle] =
-    new Monoid[Angle] {
+  /** Angle forms a commutative group. */
+  implicit val AngleCommutativeGroup: CommutativeGroup[Angle] =
```
cquiroz: Is `CommutativeGroup` a `Monoid`?

modules/core/src/test/scala/gem/Arbitraries.scala
```diff
@@ -32,6 +32,9 @@ trait Arbitraries extends gem.config.Arbitraries  {
       )
     }
 
+  implicit val cogLocation: Cogen[Location] =
```
cquiroz: What's `Cogen`?

modules/core/src/main/scala/gem/math/Angle.scala
tpolecat: Yes. It's a monoid where `|+|` is commutative and every element has an inverse, such that `a |+| -a = zero`. Also called an Abelian group. See the upper-right part of the [infographic](https://github.com/tpolecat/cats-infographic).

modules/core/src/test/scala/gem/Arbitraries.scala
tpolecat: `Cogen` is the dual of `Gen`. It's needed in order to generate "interesting" arbitrary functions. See [Erik's talk](https://www.youtube.com/watch?v=UlZGZh7hfbs) at about 36:00 for a really good explanation.

tpolecat:
> Is there some documentation on CatsSuite, it is not very clear why would or wouldn't you use it.

Not really, I'm mostly using it because it gives you `checkAll` which lets you do law testing. But I also think it's helpful to just use the same style cats is using since a lot of choices are pretty arbitrary.

modules/core/src/test/scala/gem/Arbitraries.scala
swalker2m: I watched the bit about `Cogen`.  It's very nicely done.  So when I see a `Cogen[A]` now it tells me that it will allow us to create a generator of arbitrary functions from `A => B`, given a `Gen[B]` as well.  Correct?  Do we actually make use of this anywhere yet?

tpolecat: > Do we actually make use of this anywhere yet?

Yes, you need it to test the congruence law for `Eq` … `a == b` implies `f(a) == f(b)` so you need to generate arbitrary `f: A => A`. Defined [here](https://github.com/typelevel/cats/blob/6fefa30bfb85e90d193b53300b8313c4dd6c0201/kernel-laws/src/main/scala/cats/kernel/laws/OrderLaws.scala#L36-L38).

PR #117, opened 2017-08-16
WIP Libra

cquiroz:
This is just to discuss the option to use `Libra`, at the moment this cannot be merged since `Libra` doesn't support scala.js yet.

Basically the `Wavelength` case class was replaced by a type `Wavelength` and a new type is created `WavelengthInÅngström` for a wavelength in `Å`

We are not doing operations across quantities so using `Libra` may not give us a lot but we can discuss it based on the code. A nice benefit is that we can write generic encoders.

Since `libra` brings `spire` I started using `UInt` for wavelength to ensure we have positive numbers. Again this is debatabale

TOD (In case we want to go this route)

- [ ] rebase onto cats
- [ ] support scala.js
- [ ] generic `Meta` for doobie


modules/core/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -4,54 +4,41 @@
...
+  }
+
+  /** Construct a WavelengthInÅngström from unsigned integer angstroms. */
+  def fromAngstroms(angstroms: UInt): WavelengthInÅngström =
+    angstroms.Å
 
-  final lazy val ZeroAngstroms = unsafeFromAngstroms(0)
+}
+
+object WavelengthInÅngström {
```
swalker2m: Seems like this might give us grief on Linux.

cquiroz: Yeah, it is the correct way to write bit it is very hard to write and could create problems. I'm open to write it in ASCII

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
   - npm install jsdom
 
 script:
-  - sbt headerCheck test:headerCheck scalastyle sql/flywayMigrate compile test coreJS/test
+  - sbt headerCheck test:headerCheck scalastyle sql/flywayMigrate compile test
```
cquiroz: Removed coreJS to get travis to run

modules/core/src/main/scala/gem/math/Quantities.scala
```diff
@@ -0,0 +1,16 @@
...
+package math
+
+import libra.{ QuantityOf, Unit }
+import spire.math.UInt
+
+/**
+ * Typeclass that can create a Quantity from a UInt
+ * Useful to create generic encoders
+ */
+trait QuantityOfFromUInt[A, B <: Unit[A]] {
```
cquiroz: This is a typeclass to build `QuantityOf`. I'm not sure how sound this is but it is useful to write generic encoders

modules/core/src/test/scala/gem/math/WavelengthSpec.scala
```diff
@@ -4,44 +4,40 @@
...
+    forAll { (a: WavelengthInÅngström, b: WavelengthInÅngström) =>
+      Order[UInt].order(a.value, b.value) shouldEqual
+      Order[WavelengthInÅngström].order(a, b)
     }
   }
 
-  "Show" must "be natural" in {
-    forAll { (a: Wavelength) =>
-      a.toString shouldEqual Show[Wavelength].shows(a)
+  /*"Show" must "be natural" in {
```
cquiroz: Given that `WavelengthInAngstrom` is just a type we can't override `toString`

modules/db/src/main/scala/gem/dao/package.scala
```diff
@@ -94,8 +97,12 @@ package object dao extends MoreTupleOps {
     AngleMetaAsSignedArcseconds.xmap(Offset.Q(_), _.toAngle)
 
   // Wavelength maps to an integer in angstroms
-  implicit val WavelengthMeta: Meta[Wavelength] =
-    Meta[Int].xmap(Wavelength.unsafeFromAngstroms, _.toAngstroms)
+  implicit val WavelengthMeta: Meta[WavelengthInÅngström] =
+    Meta[Int].xmap(i => Wavelength.fromAngstroms(UInt(i)), _.value.toInt)
+
+  // Wanted to do a generic quantity Meta but there is a bug with missing TypeTag
```
cquiroz: I wanted to write a generic `Meta` but this raises some issues with missing `TypeTag`. @tpolecat indicates it is an issue with `shapeless` `HList`

modules/json/src/main/scala/gem/json.scala
```diff
@@ -26,9 +28,9 @@ package object json {
         b => Angle.fromMicroarcseconds(b.underlying.movePointLeft(6).longValue))(
         a => BigDecimal(new java.math.BigDecimal(a.toMicroarcseconds).movePointRight(6)))
 
-  // Wavelength mapping to integral Angstroms.
-  implicit val WavelengthCodec: CodecJson[Wavelength] =
-    CodecJson.derived[Int].xmap(Wavelength.unsafeFromAngstroms)(_.toAngstroms)
+  // Generic QuantityOf codec
+  implicit def QuantityOfCodec[A, B <: Unit[A]](implicit F: QuantityOfFromUInt[A, B]): CodecJson[QuantityOf[UInt, A, B]] =
```
cquiroz: This should encode/decode any `QuantityOf` for UInt

tpolecat:
Are you going to add scala-js to Libra? I'm sure Zainab would take a PR

cquiroz:
First I need singleton-ops to support scala.js see 
https://github.com/fthomas/singleton-ops/pull/34

tpolecat:
@swalker2m in case you haven't seen it, [this](https://www.youtube.com/watch?v=CrwVD0Pco5Q) is a good talk.

cquiroz:
Closing this as we are most likely not using libra

PR #120, opened 2017-08-22
Fix deprecation warning

swalker2m:
Fixes a deprecation warning that stops the build.


tpolecat:
Actually this should already be a functor in the latest http4s. Can you double-check?

tpolecat:
Cool, sorry about the confusion.

PR #121, opened 2017-08-22
add coordinate parsers, re-organize a bit

tpolecat:
This adds parsers for angles, ra, dec, and coordinates. It also breaks the `Parsers.scala` module into a package with smaller modules, in the model of the `arb` package, just to keep things manageable. At the user level we have `format/unformat` pairs which delegate to a `Parser` and we test round-tripping, but in the future these pairs should become `Prism`s, along with other external format conversions such as offset angles as signed decimal arcseconds.

The current format for coordinates is

    17h 57m 48.49803s +04° 41′ 36.2072″ 

which is what wikipedia uses, although it really could be anything at all. I just picked something that's easy to read. The parser also supports

    17 57 48.49803 04 41 36.2072

which is what Simbad uses, and

    17:57:48.49803 4:41:36.2072

which ocs uses I think. In any case it's easy to change but this is as good a starting point as any.


modules/core/src/main/scala/gem/math/Angle.scala
```diff
@@ -144,6 +159,14 @@ object Angle {
   def fromDoubleRadians(rad: Double): Angle =
     fromDoubleDegrees(rad.toDegrees)
 
+  /** Attempt to parse an [[Angle]] from a `.formatDMS`-formatted string. */
+  def unformatDMS(s: String): Option[Angle] =
```
cquiroz: unformat? Why not `parse`?

tpolecat: Yeah `parse` is probably better. I wanted them to be parallel but it's kind of irritating. I'll rename them,

modules/core/src/main/scala/gem/parser/Angle.scala
```diff
@@ -0,0 +1,70 @@
...
+  def genHMS(t1: Parser[_], t2: Parser[_], t3: Parser[_]): Parser[HourAngle] =
+    genAngle(t1, t2, t3).map((HourAngle.fromHMS _).tupled)
+
+  /** 00:00:00.000000 */
+  val hms1: Parser[HourAngle] =
+    genHMS(char(':'), char(':'), ok(()))
+
+  /** 00 00 00.000000 */
+  val hms2: Parser[HourAngle] =
+    genHMS(spaces1, spaces1, ok(()))
```
swalker2m: 👍 This is also the format in horizons output.

modules/core/src/main/scala/gem/parser/Angle.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /** 00h 00m 00.000000s */
+  val hms3: Parser[HourAngle] =
+    genHMS(token(char('h')), token(char('m')), char('s'))
+
+  val hms = hms1 | hms2 | hms3
```
swalker2m: I would be tempted to name it "HMS" or something.  When do you add a name and when is it better not to?

modules/core/src/main/scala/gem/parser/Time.scala
```diff
@@ -0,0 +1,35 @@
...
+  val month2: Parser[Month] =
+    intN(2).flatMap(catchDTE(Month.of)) namedOpaque "Month"
+
+  /** Parser for a `LocalDate` in the form `20151107`. */
+  def yyyymmdd: Parser[LocalDate] =
+    for {
+      y <- year4
+      m <- month2
+      d <- intN(2) namedOpaque "2-digit day of month"
+      a <- catchDTE((d: Int) => LocalDate.of(y.getValue, m, d))(d)
```
swalker2m: The `catchDTE` seems a bit awkward in that we have to fabricate a function that we wouldn't otherwise need.  Should it just be:

```scala
def catchDTE[B](b :=> B): Parser[B] =
  try { ok(b) } catch { case e: DateTimeException => err[B](e.toString) }
```

modules/core/src/main/scala/gem/parser/Angle.scala
tpolecat: So far we're throwing away failure information by turning `ParseResult` into `Option` everywhere so it doesn't matter that much, but if we ever start trying to diagnose parse failures for users then naming will be important. So yes, you're right. I will go back and add names where it makes sense.

modules/core/src/main/scala/gem/parser/Time.scala
tpolecat: The original idea is that `catchDTE(a)` gives you an `A => Parser[B]` so you can pass it to `flatMap` as I'm doing with a similar combinator [here](https://github.com/tpolecat/gem/blob/b04fd18dedad9603e056a499b1b3e814de9dbae1/modules/core/src/main/scala/gem/parser/Misc.scala#L24-L30). But yes, the use case here doesn't benefit from the machinery. I'll simplify.

modules/core/src/main/scala/gem/parser/Angle.scala
```diff
@@ -0,0 +1,70 @@
...
+   * and at most 6 digits following the decimal point, and terminal parsers for each segment.
+   */
+  def genAngle(t1: Parser[_], t2: Parser[_], t3: Parser[_]): Parser[(Int, Int, Int, Int, Int)] =
+    for {
+      h  <- int <~ t1
+      m  <- int <~ t2
+      s  <- int <~ char('.')
+      µs <- frac(6)
+      _  <- t3
+    } yield (h, m, s, µs / 1000, µs % 1000)
```
jdnavarro: Why not use the Applicative instance here?

modules/core/src/main/scala/gem/parser/Misc.scala
```diff
@@ -0,0 +1,65 @@
...
+  val hyphen: Parser[Unit] =
+    char('-').void
+
+  /** Parser for one or more spaces. */
+  val spaces1: Parser[Unit] =
+    skipMany1(char(' '))
+
+  /** Parser for a non-whitespace string. */
+  val nonWhitespace: Parser[String] =
+    stringOf(elem(c => !c.isWhitespace))
```
jdnavarro: Not that it matters to us performance wise, but I'm curious to know if there is an easy way to check equality for whole string buffers instead of going char by char.

That's one of the main advantages of the Haskell's `attoparsec` with respect to the original `Parsec`.

tpolecat: Yeah, actually there's a more efficient combinator I can use here. `takeWhile` will look at input chunk by chunk instead of char by char.

modules/core/src/main/scala/gem/parser/Misc.scala
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.parser
+
+import atto._, Atto._
+import scala.annotation.tailrec
+
+/** General-purpose parsers and combinators that aren't provided by atto. */
```
jdnavarro: Why not upstream them? They seem quite useful for any kind of parsing in general.

modules/core/src/main/scala/gem/syntax/Parser.scala
```diff
@@ -0,0 +1,24 @@
...
+
+package gem.syntax
+
+import atto._, Atto._
+
+final class ParserOps[A](val self: Parser[A]) extends AnyVal {
+
+  /** Parse entire input into an Option. */
+  def parseExact(s: String): Option[A] =
+    (self <~ endOfInput).parseOnly(s).option
```
jdnavarro: Very clever 👍. Does it backtrack if it fails? What this will return? `(parseExact("foo") | parseExact("bar")).parseExact("bar")`

modules/core/src/main/scala/gem/parser/Misc.scala
tpolecat: Yeah, when we're sure they do what we want I'll merge them back.

modules/core/src/main/scala/gem/syntax/Parser.scala
tpolecat: This isn't a combinator really … it's an end of world operation that discharges `Parser` and gives you a normal `Option` back.

But yes, the inner combinator `a <~ endOfInput` does backtrack as expected, and it already exists as `phrase`, so I'll change it! Easy to re-invent things ;-)

modules/core/src/main/scala/gem/parser/Angle.scala
tpolecat: No reason really. Parsing is sequential so there's no difference, it's just a matter of what looks better. I'll try it the other way.

tpolecat: The old `(a |@| b |@| c)(f)` syntax was bothersome because operator precedence of `|@|` wasn't low enough. But the new `(a, b, c).mapN(f)` syntax doesn't have that issue and it works well. So I changed a bunch of `for` comps.

jdnavarro: Well, here I don't think using the applicative matters that much here. The `<*` `*>` `$>` `<$` can also come handy for parsing combinators but if they are such a pain to write I don't think it's worth bothering.

I wonder why the `|@|` doesn't have the lowest associative precedence, when would you ever need to have the `|@|` start associating from the other side? `<rant>`If the applicative syntax wasn't so cumbersome they could be actually used more generally`</rant>`.

I hope some kind of [syntactic sugar](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#applicative-do-notation) for Applicatives becomes more widespread.

modules/core/src/main/scala/gem/parser/Angle.scala
```diff
@@ -16,54 +17,48 @@ trait AngleParsers {
...
-    for {
-      h  <- int <~ t1
-      m  <- int <~ t2
-      s  <- int <~ char('.')
-      µs <- frac(6)
-      _  <- t3
-    } yield (h, m, s, µs / 1000, µs % 1000)
+    (int <~ t1, int <~ t2, int <~ char('.'), frac(6) <~ t3) mapN { (h, m, s, µs) =>
+      (h, m, s, µs / 1000, µs % 1000)
+    } named s"genAngle($t2, $t2, $t3)"
```
swalker2m: It may not be technically necessary, but the `for` comprehension syntax is easier to read, I think.  This is fine too though.

modules/core/src/main/scala/gem/parser/Angle.scala
```diff
@@ -0,0 +1,65 @@
...
+   * and at most 6 digits following the decimal point, and terminal parsers for each segment.
+   */
+  def genAngle(t1: Parser[_], t2: Parser[_], t3: Parser[_]): Parser[(Int, Int, Int, Int, Int)] =
+    (int <~ t1, int <~ t2, int <~ char('.'), frac(6) <~ t3) mapN { (h, m, s, µs) =>
+      (h, m, s, µs / 1000, µs % 1000)
+    } named s"genAngle($t2, $t2, $t3)"
+
+  /** Generic parser for the components of an HourAngle; see `genAngle`. */
+  def genHMS(t1: Parser[_], t2: Parser[_], t3: Parser[_]): Parser[HourAngle] =
+    genAngle(t1, t2, t3).map((HourAngle.fromHMS _).tupled) named s"genHMS($t2, $t2, $t3)"
```
sraaphorst: Why `genHMS($t2, $t2, $t3)` and not `genHMS($t1, $t2, $t3)`?
(Same comment for `genAngle`.)

tpolecat: Thanks, I'll fix.

PR #122, opened 2017-08-24
Convert core to a full cross project

cquiroz:


tpolecat:
What's the effect of this change?

cquiroz:
Nothing at the moment, but you can add jvm only tests to core creating a dir core/jvm/src/test/scala

This could be used by @swalker to track the troubles he's having with test resources

swalker2m:
This is what I expected to see originally -- I wasn't familiar with the `Pure` vs. `Full` option.  It seems fine to me. 👍 @tpolecat any reservations?

tpolecat:
Nope, looks fine.

PR #126, opened 2017-08-25
combinators

tpolecat:
This adds `takeUntilLast` which I think does what @swalker2m was looking for.

```scala
scala> takeUntilLast(_ == 'x').parseOnly("abc")
res0: atto.ParseResult[String] = Fail(abc,List(takeUntilʹ(…)),not satisfied)
 
scala> takeUntilLast(_ == 'x').parseOnly("abcx")
res1: atto.ParseResult[String] = Done(,abc)
 
scala> takeUntilLast(_ == 'x').parseOnly("abcxdef")
res2: atto.ParseResult[String] = Done(def,abc)
 
scala> takeUntilLast(_ == 'x').parseOnly("abcxdefx")
res3: atto.ParseResult[String] = Done(,abcxdef)
 
scala> takeUntilLast(_ == 'x').parseOnly("abcxdefxyz")
res4: atto.ParseResult[String] = Done(yz,abcxdef)
 
scala> takeUntilLast(_ == 'x').parse("ab").feed("cxde").feed("fxyz").done
res5: atto.ParseResult[String] = Done(yz,abcxdef)
```

So you can get the paren bracketing with

```scala
scala> val parens = char('(') ~> takeUntilLast(_ == ')')
parens: atto.Parser[String] = ('(') ~> takeUntilʹ(…)
 
scala> parens.parseOnly("(ab(c)de)fh)hij")
res7: atto.ParseResult[String] = Done(hij,ab(c)de)fh)
```

### DIFF
--- a/modules/core/shared/src/main/scala/gem/parser/Misc.scala
+++ b/modules/core/shared/src/main/scala/gem/parser/Misc.scala
@@ -85,7 +85,7 @@ trait MiscParsers {
         case -1 => err[String]("not satisfied")
         case  n => take(n) <~ advance(1)
       }
-    } named "takeUntilʹ(…)"
+    } named "takeUntilLast(…)"
 
 }
 object MiscParsers extends MiscParsers
--- a/modules/core/shared/src/main/scala/gem/parser/Misc.scala
+++ b/modules/core/shared/src/main/scala/gem/parser/Misc.scala
@@ -65,5 +65,27 @@ trait MiscParsers {
 
   } namedOpaque s"frac($n)"
 
+
+  /** Force and return all remaining input without consuming it. */
+  val force: Parser[String] =
+    get.flatMap { s =>
+      opt(ensure(s.length + 1)) flatMap {
+        case Some(_) => force
+        case None    => ok(s)
+      }
+    }
+
+  /**
+   * Force and examine remaining input and yield the longest prefix up to (but not including) a
+   * char satisfying `p`, which is discarded. Fails is no such char is found.
+   */
+  def takeUntilLast(p: Char => Boolean): Parser[String] =
+    force.flatMap { s =>
+      s.lastIndexWhere(p) match {
+        case -1 => err[String]("not satisfied")
+        case  n => take(n) <~ advance(1)
+      }
+    } named "takeUntilʹ(…)"
+
 }
 object MiscParsers extends MiscParsers
--- a/modules/core/shared/src/main/scala/gem/parser/Misc.scala
+++ b/modules/core/shared/src/main/scala/gem/parser/Misc.scala
@@ -77,7 +77,7 @@ trait MiscParsers {
 
   /**
    * Force and examine remaining input and yield the longest prefix up to (but not including) a
-   * char satisfying `p`, which is discarded. Fails is no such char is found.
+   * char satisfying `p`, which is discarded. Fails if no such char is found.
    */
   def takeUntilLast(p: Char => Boolean): Parser[String] =
     force.flatMap { s =>


modules/core/shared/src/main/scala/gem/parser/Misc.scala
```diff
@@ -65,5 +65,27 @@ trait MiscParsers {
 
   } namedOpaque s"frac($n)"
 
+
+  /** Force and return all remaining input without consuming it. */
+  val force: Parser[String] =
+    get.flatMap { s =>
+      opt(ensure(s.length + 1)) flatMap {
```
swalker2m: How is `opt(ensure(s.length + 1))` not always `None`?

tpolecat: atto does incremental parsing, so it may not see all the input at once. `get` lets you peek at [the remainder of] the current chunk of input. When it runs out you get a `Partial` parse result that says it's ok so far but you need to feed more (or say you're done) before you know for sure whether it succeeds or fails.

If your parser need a specific minimum amount of input you can use `ensure` to force more chunks to be read if there's not enough left in the current chunk. This is a suspended computation so what you "forcing" means is that you end up with is a `Partial` that will continue to be `Partial` until you feed it enough data to satisfy `ensure` or establish that there's really not enough input.

Neither `get` nor `ensure` actually "consume" the input, so once you establish that you have sufficient input that matches whatever you're looking for you can then `take` or `advance` or whatever to move forward.

So tl;dr `force` just repeatedly asks for more input than we have in order to pull everything into the buffer.

modules/core/shared/src/main/scala/gem/parser/Misc.scala
```diff
@@ -65,5 +65,27 @@ trait MiscParsers {
...
+    }
+
+  /**
+   * Force and examine remaining input and yield the longest prefix up to (but not including) a
+   * char satisfying `p`, which is discarded. Fails if no such char is found.
+   */
+  def takeUntilLast(p: Char => Boolean): Parser[String] =
+    force.flatMap { s =>
+      s.lastIndexWhere(p) match {
+        case -1 => err[String]("not satisfied")
```
jdnavarro: What's the type returned by `err[String]`? Is it the same type for all parsing errors in `atto`?

swalker2m: It produces a `Parser[String]`, a "Parser that consumes no data and fails with the specified error message." when applied.

jdnavarro: I've seen in the source and the tutorials that `err` returns a `Failure` which I assume it contains all the context information of the error, like in which line/column or how what's the input remainder (in the case of `takeUntilLast` it doesn't matter). But I can't where `Failure` is defined.

I'm not trying to point at anything wrong here, I'm just trying to learn more about `atto`.

jdnavarro: Never mind about my previous comment, I [found it](https://github.com/tpolecat/atto/blob/de9e21f5b2ade83102a425089ef9e294fb157cb8/modules/core/src/main/scala/atto/Parser.scala#L75). It seems the error has all the context I was mentioning above. Neat 👍.

tpolecat: Error reporting isn't very good. You get an error message (a string), the current stack (which is a list of strings), and the remaining unconsumed input. If you want to know the parse position you can use the `pos` constructor which consumes no input and yields the absolute offset.

PR #127, opened 2017-08-28
WIP on proper motion

tpolecat:
This is on hold pending some test data. I'll move on to something else in the meantime.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
+++ b/modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
@@ -1,0 +1,148 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.math
+
+import cats.implicits._
+import scala.math.{ sin, cos, hypot, atan2 }
+
+/**
+ * Time-parameterized coordinates, based on an observed position at some point in time (called
+ * the `epoch`) and measured velocities in distance (`redShift`; i.e., Doppler shift) and position
+ * (`velocity`) per year.
+ * @param base observed coordinates at `epoch`
+ * @param epoch time of the base observation; typically `Epoch.J2000`
+ * @param pv proper velocity *per year* in [[RightAscension]] and [[Declination]]
+ * @param rv radial velocity in km/y, positive if receding
+ * @param px parallax
+ */
+final case class ProperMotion(
+  base:  Coordinates,
+  epoch: Epoch,
+  pv:    Option[Offset],
+  rv:    Option[Double], // for now
+  px:    Option[Angle]
+) {
+
+  def plusYears(dt: Double): Coordinates =
+    ProperMotion.properMotion(base, epoch, pv.orEmpty, px.orEmpty, rv.getOrElse(0.0), dt)
+
+}
+
+
+object ProperMotion {
+
+  /** Proper motion of Barnard's Star, for reference. */
+  val Barnard: ProperMotion =
+    ProperMotion(
+      Coordinates.parse("17 57 48.49803 +04 41 36.2072").getOrElse(sys.error("oops")),
+      Epoch.J2000,
+      Some(Offset(
+        Offset.P(Angle.fromMicroarcseconds( -798580L)),
+        Offset.Q(Angle.fromMicroarcseconds(10328120L))
+      )),
+      Some(-110.6), // km/s
+      Some(Angle.fromMicroarcseconds(545620L))
+    )
+
+  def main(args: Array[String]): Unit = {
+    (0.0 to 10.0 by 1.0) foreach { y =>
+      val cs = Barnard.plusYears(y)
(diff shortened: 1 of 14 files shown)


modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,182 @@
...
+ * @param baseCoordinates observed coordinates at `epoch`
+ * @param epoch           time of the base observation; typically `Epoch.J2000`
+ * @param properVelocity  proper velocity '''per year''' in [[RightAscension]] and [[Declination]], if any
+ * @param radialVelocity  radial velocity (km/y, positive if receding), if any
+ * @param parallax        parallax, if any
+ */
+final case class ProperMotion(
+  baseCoordinates: Coordinates,
+  epoch:           Epoch,
+  properVelocity:  Option[Offset],
```
tpolecat: I don't know if it's ethical to use an `Offset` here although that's what the velocity per unit time is I guess.

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,182 @@
...
+  radialVelocity:  Option[Double], // for now
```
tpolecat: I don't know if it's worth the trouble newtyping this since it's never used anywhere else. I guess with Libra or whatever we could do this without giving it much thought.

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,182 @@
...
+  parallax:        Option[Angle]
+) {
+
+  /** Coordinates `elapsedYears` fractional epoch-years after `epoch`. */
+  def plusYears(elapsedYears: Double): Coordinates =
```
tpolecat: We also need `atInstant(t: Instant)`.

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,182 @@
...
+  private type Vec3 = (Double, Double, Double)
+
+  // |+| gives us addition for VecN, but we also need scalar multiplication
+  private implicit class Vec3Ops(a: Vec3) {
+    def *(d: Double): Vec3 =
+      (a._1 * d, a._2 * d, a._3 * d)
+  }
+
+  /**
+   * Proper motion correction in base units.
```
tpolecat: This is private ... once we drop down to `Double` it probably makes sense to hide our cards. We'll see if it turns out to be a pain.

tpolecat:
> This is pretty bad-ass.

I don't understand it and fully expect it to be wrong once we look more closely. I'm just enjoying not knowing yet.

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,182 @@
...
+ * @param radialVelocity  radial velocity (km/y, positive if receding), if any
+ * @param parallax        parallax, if any
+ */
+final case class ProperMotion(
+  baseCoordinates: Coordinates,
+  epoch:           Epoch,
+  properVelocity:  Option[Offset],
+  radialVelocity:  Option[Double], // for now
+  parallax:        Option[Angle]
+) {
```
jdnavarro: What will be the normal way to find out the parameters to create a `ProperMotion`? Will they be retrieved from some external service or will we have them stored somewhere locally? Are these parameters constant for all sidereal targets?

swalker2m: The values come from target resolution services like Simbad.  I'm not sure but I imagine our guide star catalog has proper motion information?

http://simbad.u-strasbg.fr/simbad/sim-basic?Ident=vega&submit=SIMBAD+search

tpolecat:
Andy thinks this is good enough since it matches his back-of-the-envelope calculations but I'm not very confident. I couldn't come up with any evidently obvious test cases that actually passed.

But in any case this gives us enough structure to move forward with the target build-out so I guess it's ok. Since the initial PR I added an integral `RadialVelocity` class so we don't have a `Double` in the model.

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -0,0 +1,188 @@
...
+ */
+final case class ProperMotion(
+  baseCoordinates: Coordinates,
+  epoch:           Epoch,
+  properVelocity:  Option[Offset],
+  radialVelocity:  Option[RadialVelocity],
+  parallax:        Option[Angle]
+) {
+
+  /** Coordinates `elapsedYears` fractional epoch-years after `epoch`. */
```
swalker2m: Does this comment need to be updated now that this returns a `ProperMotion`?

tpolecat: Yep, I'll fix it Tuesday (Monday is a holiday here).

PR #128, opened 2017-08-28
Horizons Designation

swalker2m:
Brings the `HorizonsDesignation` over from the old `ocs` project so that it is is available to use in the non-sidereal target model.  Horizons designation is a unique horizons id that can be used to construct non-sidereal target queries in JPL horizons.  This is mostly boilerplate but there were some small decisions that may need to be re-evaluated:

1) I placed the `HorizonsDesignation` in a new package called `target` anticipating that the rest of the target model will go there rather than `gem.math`.  Of course `target` is excluded by `.gitignore` by default so that may cause problems?

2) I made an `HorizonsType` enum in the database, anticipating that we'll want to store that, along with the string designation, in the non-sidereal target table.  It isn't really needed yet though.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/enum/HorizonsType.scala
+++ b/modules/core/shared/src/main/scala/gem/enum/HorizonsType.scala
@@ -1,0 +1,46 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package enum
+
+import cats.syntax.eq._
+import cats.instances.string._
+
+/**
+ * Enumerated type for horizons non-sidereal target type.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class HorizonsType(
+  val tag: String,
+  val name: String
+)
+
+object HorizonsType {
+
+  /** @group Constructors */ case object Comet extends HorizonsType("Comet", "Comet")
+  /** @group Constructors */ case object AsteroidNew extends HorizonsType("AsteroidNew", "Asteroid New")
+  /** @group Constructors */ case object AsteroidOld extends HorizonsType("AsteroidOld", "Asteroid Old")
+  /** @group Constructors */ case object MajorBody extends HorizonsType("MajorBody", "Major Body")
+
+  /** All members of HorizonsType, in canonical order. */
+  val all: List[HorizonsType] =
+    List(Comet, AsteroidNew, AsteroidOld, MajorBody)
+
+  /** Select the member of HorizonsType with the given tag, if any. */
+  def fromTag(s: String): Option[HorizonsType] =
+    all.find(_.tag === s)
+
+  /** Select the member of HorizonsType with the given tag, throwing if absent. */
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def unsafeFromTag(s: String): HorizonsType =
+    fromTag(s).getOrElse(throw new NoSuchElementException(s))
+
+  /** @group Typeclass Instances */
+  implicit val HorizonsTypeEnumerated: Enumerated[HorizonsType] =
+    new Enumerated[HorizonsType] {
+      def all = HorizonsType.all
+      def tag(a: HorizonsType) = a.tag
+    }
+
+}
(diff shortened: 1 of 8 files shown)


modules/core/shared/src/main/scala/gem/enum/HorizonsType.scala
```diff
@@ -0,0 +1,46 @@
...
+sealed abstract class HorizonsType(
+  val tag: String,
+  val name: String
+)
+
+object HorizonsType {
+
+  /** @group Constructors */ case object Comet extends HorizonsType("Comet", "Comet")
+  /** @group Constructors */ case object AsteroidNew extends HorizonsType("AsteroidNew", "Asteroid New")
+  /** @group Constructors */ case object AsteroidOld extends HorizonsType("AsteroidOld", "Asteroid Old")
```
tpolecat: These will eventually be shown to the user so `Asteroid (New Format)` or similar might be more helpful.

modules/core/shared/src/main/scala/gem/parser/HorizonsDesignation.scala
```diff
@@ -0,0 +1,44 @@
...
+import gem.target.HorizonsDesignation._
+
+import atto._, Atto._
+import cats.implicits._
+
+/** Parser for [[gem.target.HorizonsDesignation]]. */
+trait HorizonsDesignationParsers {
+
+  private def des[A,B](s: String, p: Parser[A]): Parser[A] =
+    string(s"${s}_") ~> p
```
tpolecat: Probably don't need type param `B` here.

modules/core/shared/src/main/scala/gem/parser/HorizonsDesignation.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  private def textDes[A](s: String)(f: String => A): Parser[A] =
+    des(s, takeText.map(f))
+
+  private def numDes[A](s: String)(f: Int => A): Parser[A] =
+    des(s, int.map(f))
+
+  val comet: Parser[Comet] =
+    textDes("Comet"      )(Comet.apply      ) namedOpaque "comet"
```
tpolecat: whitespace

modules/core/shared/src/main/scala/gem/parser/HorizonsDesignation.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  val asteroidNew: Parser[AsteroidNew] =
+    textDes("AsteroidNew")(AsteroidNew.apply) namedOpaque "asteroidNew"
+
+  val asteroidOld: Parser[AsteroidOld] =
+    numDes ("AsteroidOld")(AsteroidOld.apply) namedOpaque "asteroidOld"
+
+  val major: Parser[MajorBody] =
```
tpolecat: Can we call this `majorBody` for consistency with the name?

modules/core/shared/src/main/scala/gem/parser/HorizonsDesignation.scala
```diff
@@ -0,0 +1,44 @@
...
+    numDes ("MajorBody"  )(MajorBody.apply  ) namedOpaque "majorBody"
+
+  val horizonsDesignation: Parser[HorizonsDesignation] =
+      (comet      .widen[HorizonsDesignation] |
+       asteroidNew.widen[HorizonsDesignation] |
+       asteroidOld.widen[HorizonsDesignation] |
+       major      .widen[HorizonsDesignation] ) named "horizonsDesignation"
```
tpolecat: This is irritating, sorry.

modules/core/shared/src/main/scala/gem/parser/HorizonsDesignation.scala
swalker2m: Oops, that's left over from an earlier version.

tpolecat:
The database enum makes sense to me because it guarantees that the designations will at least be well-formed, so 👍 on that.

tpolecat:
I think the `target` package will end up being problematic, so we might reconsider that one. I have no good suggestion though.

swalker2m:
> I had a few minor comments. Can we add a test that verifies round-tripping through the parser?

D'oh, that was ignored by `.gitignore`.  I'll force it in.

tpolecat:
Yeah let's definitely rename `target/`

tpolecat:
Honestly I would be fine with this in `gem`. We can subdivide when it gets too crowded. I was thinking of pushing `Enumerated` and `Location` into `gem.util` since they don't really have anything specific to do with the model.

tpolecat:
Looks great! 👍

PR #129, opened 2017-08-28
move Location and Enumerated into gem.util

tpolecat:
This moves `Location` and `Enumerated` into `gem.util` as proposed in #128. 

(Will need to be rebased after #128 which introduces a new enum.)


tpolecat:
Rebased to accommodate `HorizonsType`. Should be good now.

PR #130, opened 2017-08-28
update flyway and doobie versions

tpolecat:
This updates flyway to 4.2.0 and doobie to 0.5.0-M6 (which required changing a bunch of imports to avoid deprecation warnings).


tpolecat:
Ah, I added `sql` to the root project `aggregate` list so it would participate in the automatic dependency update checking, and as a side-effect it also gets the scalastyle check.

tpolecat:
Fixed scalastyle, still need to rebase after #129

PR #131, opened 2017-08-29
switch to monocle lenses

tpolecat:
This removes our janky `Lens` class and switches to [Monocle](http://julien-truffaut.github.io/Monocle/). I would also like to change all our `A => B`, `B => Option[A]` format/parse pairs into `Prism`s but I'll do that in another PR.

Hopefully this is the last infrastructure change for a while.

Good intro video [here](https://www.youtube.com/watch?v=6nyGVgGEKdA).


modules/core/shared/src/main/scala/gem/HorizonsDesignation.scala
```diff
@@ -44,53 +44,41 @@ object HorizonsDesignation {
...
    * the query string `NAME=C/1973 E1;CAP`.
    */
-  final case class Comet(des: String) extends HorizonsDesignation(s"NAME=$des;CAP")
-
-  object Comet {
-    val des: Comet @> String =
-      Lens((a, b) => a.copy(des = b), _.des)
-  }
+  @Lenses final case class Comet(des: String) extends HorizonsDesignation(s"NAME=$des;CAP")
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
```
tpolecat: `@Lenses` generates lenses for all case class fields. They show up in the companion object as values of the same name. Do `Comet.des` is the lens for the `des` field. Unfortuntely generated lenses don't have ascribed types, which causes WartRemover to complain. I opened an issue for this https://github.com/julien-truffaut/Monocle/issues/530 that should be easy to fix.

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -91,16 +91,16 @@ object StaticConfigDao {
         ro <- Statements.Gmos.selectCustomRoiEntry(sid, GmosN).list.map(_.toSet)
         ns <- Statements.Gmos.selectNodAndShuffle(sid, GmosN).option
         gn <- Statements.Gmos.selectNorth(sid).unique
-        gnʹ = GmosNorth.CustomRois.set(gn, ro)
-      } yield GmosNorth.NodAndShuffle.set(gnʹ, ns)
+        gnʹ = GmosNorth.customRois.set(ro)(gn)
+      } yield GmosNorth.nodAndShuffle.set(ns)(gnʹ)
```
tpolecat: Note that `set(a, b)` becomes `set(b)(a)`

modules/core/shared/src/main/scala/gem/config/StaticConfig.scala
```diff
@@ -47,35 +48,38 @@ object StaticConfig {
...
+
   }
 
-  trait GmosNorthLenses {
-    val Common: GmosNorth @> GmosCommonStaticConfig =
-      Lens((a, b) => a.copy(common = b), _.common)
+  trait GmosNorthLenses { this: GmosNorth.type =>
+
+    lazy val customRois: Lens[GmosNorth, Set[GmosCustomRoiEntry]] =
+      common ^|-> GmosCommonStaticConfig.customRois
```
tpolecat: `a >=> b` becomes `a ^|-> b`.

swalker2m: No big deal but I think I like >=> more.

modules/core/shared/src/main/scala/gem/config/StaticConfig.scala
```diff
@@ -47,35 +48,38 @@ object StaticConfig {
...
 
-    val CustomRois: GmosNorth @> Set[GmosCustomRoiEntry] =
-      Common >=> GmosCommonStaticConfig.CustomRois
+    lazy val nodAndShuffle: Lens[GmosNorth, Option[GmosNodAndShuffle]] =
```
swalker2m: Perhaps we should add a `@>` type alias?

jdnavarro:
> I would also like to change all our A => B, B => Option[A] format/parse pairs into Prisms but I'll do that in another PR.

It's tricky to make a lawful `Prism` from a parser/pretty printer because there are usually multiple ways different formats could be valid (parseable). I ran into several headaches in a previous project because of this, so please watch out for the `Prism` laws because if they are not abiding we can have problems when composing them.

modules/core/shared/src/main/scala/gem/HorizonsDesignation.scala
jdnavarro: Is it too cumbersome to create the `Lens` manually? Just curious of how much the macros do for us here, I'm not suggesting to do it by hand.

cquiroz:
@jdnavarro Another thing I ran with prisms is that with an ADT you can easily miss defining your prisms for all the elements, especially if new ones are added afterwards. it is worth to run a test with that as well as checking the laws

modules/core/shared/src/main/scala/gem/config/StaticConfig.scala
jdnavarro: I'm used to the `>=>` for composing Monadic functions with a pointfree style ([Kleisli composition](https://hackage.haskell.org/package/base-4.10.0.0/docs/Control-Monad.html#v:-62--61--62-)). If `^|->` is an entire different thing, I think it's worth the rename, otherwise I'm not so sure.

swalker2m: I propose the name "wind vane" for referring to `^|->`.

jdnavarro: image

Sorry, I couldn't help it.

modules/core/shared/src/main/scala/gem/HorizonsDesignation.scala
tpolecat: Once we get that warning fixed it's still two lines of code per lens because our style guidlines require an explicit return type. You can see the available variations [here](http://julien-truffaut.github.io/Monocle/optics/lens.html).

modules/core/shared/src/main/scala/gem/config/StaticConfig.scala
tpolecat: I'd prefer to use the lib with its intended API for a while and see how it goes, but if it ends up being a pain then sure, we can add aliases. Same with `Either` and `\/` … I thought it would bug me but so far I don't care.

tpolecat:
Monocle provides law-checking for all the optics so we'll certainly use that.

PR #133, opened 2017-08-30
EphemerisKey

swalker2m:
This is the first of a series of PRs related to adding ephemeris support.  In this installment, I'm just generalizing the old `HorizonsDesignation` into an `EphemerisKey` which incorporates Horizons ids and user-supplied ephemeris ids.  I'm anticipating using a database `SEQUENCE` to generate user-supplied ephemeris ids.


modules/core/shared/src/main/scala/gem/EphemerisKey.scala
```diff
@@ -0,0 +1,135 @@
...
+    this match {
+      case AsteroidNew(_)  => EphemerisKeyType.AsteroidNew
+      case AsteroidOld(_)  => EphemerisKeyType.AsteroidOld
+      case Comet(_)        => EphemerisKeyType.Comet
+      case MajorBody(_)    => EphemerisKeyType.MajorBody
+      case UserSupplied(_) => EphemerisKeyType.UserSupplied
+    }
+
+  /** Exports an ephemeris key to a `String` in a format that can be read by the
+    * `gem.parsers.EphemerisKeyParsers` method.
```
tpolecat: `EphemerisKey.parse`?

modules/core/shared/src/main/scala/gem/EphemerisKey.scala
```diff
@@ -0,0 +1,135 @@
...
+    override def des: String =
+      num.toString
+  }
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  object MajorBody
+
+  /** Identifies a user-supplied collection of ephemeris data, where the number
+    * comes from a database sequence.
+    */
+  @Lenses final case class UserSupplied(num: Int) extends EphemerisKey {
```
tpolecat: I think I might prefer `id` here, just to alleviate potential confusion about `num` actually meaning something.

swalker2m:
> A couple comments

I'll fix them in the next PR to avoid dismissing the review.

PR #134, opened 2017-09-04
Ephemeris Table and DAO

swalker2m:
This is the next installment in porting ephemeris support from the old OCS.  Here we introduce an `ephemeris` table and a DAO for getting data into and out of it.

This PR also changes the format of RA and Dec to `HH:MM:SS.µµµµµµ` and `[-+]DD:MM:SS.µµµµµµ` respectively to support sorting by RA and Dec.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
+++ b/modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
@@ -36,17 +36,17 @@ final class InstantMicros private (val toInstant: Instant) extends AnyVal {
   /** Creates an updated instance of InstantMicros by applying the given
     * function to its wrapped Instant.
     */
-  def updated(f: Instant => Instant): InstantMicros =
+  def mod(f: Instant => Instant): InstantMicros =
     InstantMicros.truncate(f(toInstant))
 
   def plusMillis(millisToAdd: Long): InstantMicros =
-    updated(_.plusMillis(millisToAdd))
+    mod(_.plusMillis(millisToAdd))
 
   def plusMicros(microsToAdd: Long): InstantMicros =
-    updated(_.plusNanos(microsToAdd * 1000))
+    mod(_.plusNanos(microsToAdd * 1000))
 
   def plusSeconds(secondsToAdd: Long): InstantMicros =
-    updated(_.plusSeconds(secondsToAdd))
+    mod(_.plusSeconds(secondsToAdd))
 
   override def toString: String =
     toInstant.toString
--- a/modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
+++ b/modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
@@ -1,0 +1,93 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.util
+
+import cats._
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit.MICROS
+
+
+/** InstantMicros wraps a `java.util.Instant` that is truncated to microsecond
+  * resolution.  This allows InstantMicros to roundtrip to/from the database
+  * where timestamps support only microsecond resolution.
+  *
+  * @param toInstant
+  */
+final class InstantMicros private (val toInstant: Instant) extends AnyVal {
+
+  /** Gets the number of seconds from the Java epoch of 1970-01-01T00:00:00Z. */
+  def epochSecond: Long =
+    toInstant.getEpochSecond
+
+  /** Gets the number of microseconds after the start of the second returned
+    * by `epochSecond`.
+    */
+  def µs: Long =
+    toInstant.getNano / 1000l
+
+  /** Converts this instant to the number of milliseconds from the epoch of
+    * 1970-01-01T00:00:00Z.
+    */
+  def toEpochMilli: Long =
+    toInstant.toEpochMilli
+
+  /** Creates an updated instance of InstantMicros by applying the given
+    * function to its wrapped Instant.
+    */
+  def updated(f: Instant => Instant): InstantMicros =
+    InstantMicros.truncate(f(toInstant))
+
+  def plusMillis(millisToAdd: Long): InstantMicros =
+    updated(_.plusMillis(millisToAdd))
+
+  def plusMicros(microsToAdd: Long): InstantMicros =
+    updated(_.plusNanos(microsToAdd * 1000))
+
+  def plusSeconds(secondsToAdd: Long): InstantMicros =
+    updated(_.plusSeconds(secondsToAdd))
+
+  override def toString: String =
+    toInstant.toString
+}
+
+object InstantMicros {
+
+  val Min: InstantMicros = InstantMicros.truncate(Instant.MIN)
+  val Max: InstantMicros = InstantMicros.truncate(Instant.MAX)
+
+
+  /** Creates an InstantMicro from the given Instant, assuring that the time
+    * value recorded has a round number of microseconds.
+    *
+    * @group Constructors
+    */
+  def truncate(i: Instant): InstantMicros =
+    new InstantMicros(i.truncatedTo(MICROS))
+
+  /** Creates an InstantMicro representing the current time, truncated to the
+    * last integral number of microseconds.
+    *
+    * @group Constructors
+    */
+  def now(): InstantMicros =
+    truncate(Instant.now())
+
+  /** Creates an InstantMicro representing the current time using milliseconds
+    * from the Java time epoch.
+    *
+    * @group Constructors
+    */
+  def ofEpochMilli(epochMilli: Long): InstantMicros =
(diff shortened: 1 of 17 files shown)


swalker2m:
This does not yet include support for knowing the version of ephemeris data, when it was last updated, when an update check was performed, etc.  I think that information belongs in a separate table.

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
```diff
@@ -0,0 +1,93 @@
...
+/** InstantMicros wraps a `java.util.Instant` that is truncated to microsecond
+  * resolution.  This allows InstantMicros to roundtrip to/from the database
+  * where timestamps support only microsecond resolution.
+  *
+  * @param toInstant
+  */
+final class InstantMicros private (val toInstant: Instant) extends AnyVal {
+
+  /** Gets the number of seconds from the Java epoch of 1970-01-01T00:00:00Z. */
+  def epochSecond: Long =
```
cquiroz: Should this be named consistently with `toEpochMilli`?

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
```diff
@@ -0,0 +1,93 @@
...
+  val Max: InstantMicros = InstantMicros.truncate(Instant.MAX)
+
+
+  /** Creates an InstantMicro from the given Instant, assuring that the time
+    * value recorded has a round number of microseconds.
+    *
+    * @group Constructors
+    */
+  def truncate(i: Instant): InstantMicros =
+    new InstantMicros(i.truncatedTo(MICROS))
```
cquiroz: Will truncate just ignore nanos? Or would it approximate up/down?

modules/core/shared/src/test/scala/gem/util/InstantMicrosSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+
+
+@SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.NonUnitStatements"))
+final class InstantMicrosSpec extends CatsSuite {
+
+  // Laws
+  checkAll("InstantMicro", OrderLaws[InstantMicros].order)
+
+  test("Construction should truncate Instant nanoseconds to microseconds") {
+    forAll { (i: InstantMicros) =>
```
cquiroz: Should we test the round-trip?

modules/db/src/main/scala/gem/dao/EphemerisDao.scala
```diff
@@ -0,0 +1,131 @@
...
+      .segmentN(4096)                                                   // Stream[M, Segment[EphemerisRow, Unit]]
+      .flatMap { rows =>
+        Stream.eval(Statements.insert.updateMany(rows.toVector).transact(xa))
+      }
+
+  def delete(k: EphemerisKey): ConnectionIO[Int] =
+    Statements.delete(k).run
+
+  def update(k: EphemerisKey, e: Ephemeris): ConnectionIO[Unit] =
+    (delete(k) *> insert(k, e)).void
```
cquiroz: i don't know enough doobie, so this is more a question than a comment. Is this usual? to delete and insert rather than do a straight update?

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
swalker2m: Good question.  I chose to follow Scala conventions over consistency with Java's `Instant`.  I don't know which is better.  `toEpochMilli` works just as well for Scala though.

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
swalker2m: It truncates them so 999 nanoseconds becomes 0 microseconds.  It's all academic given computer clock and especially horizons time resolution, but admittedly it's not satisfying.  Maybe we should round.

modules/core/shared/src/test/scala/gem/util/InstantMicrosSpec.scala
swalker2m: They are round-tripped in the `EphemerisDaoSpec` because they serve as `Ephemeris` keys.  An arbitrary `InstantMicros` may have any number of microseconds and the must be faithfully written and read from the database to pass the tests.

cquiroz: 👍

modules/db/src/main/scala/gem/dao/EphemerisDao.scala
swalker2m: This is intended to support a change in the horizons output where we want to wholesale replace all the ephemeris element rows.  The rows are logically keyed by the id and timestamp and the timestamps themselves are potentially being updated with new horizons output so it seemed like the best option to just wipe them out and write the new data.

Now that I'm looking at this again though, I probably should have actually made an actual primary key out of the triple (`key_type`, `key`, `timestamp`).  I'll go back and do that.

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
```diff
@@ -0,0 +1,93 @@
...
+    * 1970-01-01T00:00:00Z.
+    */
+  def toEpochMilli: Long =
+    toInstant.toEpochMilli
+
+  /** Creates an updated instance of InstantMicros by applying the given
+    * function to its wrapped Instant.
+    */
+  def mod(f: Instant => Instant): InstantMicros =
+    InstantMicros.truncate(f(toInstant))
```
jdnavarro: In scalaz the pointfree version of these functions would be something like `(InstantMicros.truncate _) compose f` which isn't worth the trouble, but do you know if there an nicer syntax for pointfree style in `cats`?

modules/db/src/main/scala/gem/dao/EphemerisDao.scala
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+  def streamInsert[M[_]: Monad](k: EphemerisKey, s: Stream[M, Ephemeris.Element], xa: Transactor[M]): Stream[M, Int] =
+    Stream.constant(k)                                                  // Stream[Pure, EphemerisKey]
+      .zip(s)                                                           // Stream[M, (EphemerisKey, Ephemeris.Element)]
+      .map { case (k, (i, c)) => (k, i, c, c.ra.format, c.dec.format) } // Stream[M, EphemerisRow]
+      .segmentN(4096)                                                   // Stream[M, Segment[EphemerisRow, Unit]]
+      .flatMap { rows =>
+        Stream.eval(Statements.insert.updateMany(rows.toVector).transact(xa))
+      }
```
jdnavarro: This is fs2-0.10, right?

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
```diff
@@ -0,0 +1,93 @@
...
+import java.time.temporal.ChronoUnit.MICROS
+
+
+/** InstantMicros wraps a `java.util.Instant` that is truncated to microsecond
+  * resolution.  This allows InstantMicros to roundtrip to/from the database
+  * where timestamps support only microsecond resolution.
+  *
+  * @param toInstant
+  */
+final class InstantMicros private (val toInstant: Instant) extends AnyVal {
```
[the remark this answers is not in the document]
swalker2m: The `private` just hides the constructor but the `val` makes `toInstant` public.  I've always thought that syntax to be a bit awkward.  Here I wanted to hide the constructor because not just any `Instant` can be used -- it has to be a truncated instant.

modules/db/src/main/scala/gem/dao/EphemerisDao.scala
swalker2m: Yes, 0.10.0-M6

modules/core/shared/src/main/scala/gem/util/InstantMicros.scala
```diff
@@ -0,0 +1,93 @@
...
+    */
+  def truncate(i: Instant): InstantMicros =
+    new InstantMicros(i.truncatedTo(MICROS))
+
+  /** Creates an InstantMicro representing the current time, truncated to the
+    * last integral number of microseconds.
+    *
+    * @group Constructors
+    */
+  def now(): InstantMicros =
```
tpolecat: This should really be `IO[InstantMicros]` … we're not using cats-effect in core but we could pull it in I guess.

modules/core/shared/src/test/scala/gem/util/InstantMicrosSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+      i.toInstant.getNano % 1000l == 0
```
tpolecat: `L` pls

modules/db/src/main/scala/gem/dao/EphemerisDao.scala
tpolecat: We might also want `streamUpdate` here. I guess we'll see how it performs.

swalker2m:
Addressed the feedback and rebased.

PR #135, opened 2017-09-07
Add site to the ephemeris table

swalker2m:
Following up on our HipChat conversation, here is a simple proposal to add site information to the ephemeris table.  In a nut:

```
ALTER TABLE ephemeris
    DROP CONSTRAINT ephemeris_pkey,
    ADD COLUMN site identifier NOT NULL REFERENCES e_site ON DELETE CASCADE,
    ADD PRIMARY KEY (key_type, key, site, timestamp);
```


swalker2m:
I could probably be convinced otherwise, but I think a site-agnostic database is probably the right thing to do.

PR #137, opened 2017-09-07
schemaspy

tpolecat:
This adds Schema Spy to the build so you can generate a data dictionary locally. Eventually this will be produced and published as part of the automated build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ code in the `sql` project. In any case, you can [re]-generate the enumerated typ
 sbt genEnums
 ```
 
+### Generating Schema Documentation
+
+You can do `sbt schemaSpy` to generate a little website about the database. It will appear in `modules/sql/target/schemaspy`.
+
 
 ### Importing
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -143,12 +147,6 @@ Multiple main classes detected, select one to run:
 If you pick the program importer, it will import everything which is the same as explicitly passing in `Int.MaxValue`.
 
 
-### Enumerated Types
-
-Enumerated types are represented by tables named `e_whatever` which are the source for generated code on the Scala side. After compiling if you look in `modules/core/target/scala-2.11/src_managed/gem` you will see the source. The rationale for this is that the database becomes the source of truth, which makes things like filter wavelengths, etc., available for querying and reporting.
```
swalker2m: This still seems like useful information.

build.sbt
```diff
@@ -27,6 +27,10 @@ git.uncommittedSignifier in ThisBuild := Some("UNCOMMITTED")
 // check for library updates whenever the project is [re]load
 onLoad in Global := { s => "dependencyUpdates" :: s }
 
+// some extra commands for us
+addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum; headerCreate")
+addCommandAlias("schemaSpy", "sql/runMain org.schemaspy.Main -t pgsql -port 5432 -db gem -o modules/sql/target/schemaspy -u postgres -host localhost -s public")
```
swalker2m: Curious why `genEnums` needs a semicolon.

README.md
tpolecat: I'll add it back. It kind of duplicates some information above.

build.sbt
tpolecat: `genEnums` has two commands, `runMain` and `headerCreate`. So you need the `;` (which is itself a command).

tpolecat:
Ok updated the readme.

PR #142, opened 2017-09-14
Ephemeris Query

swalker2m:
This installment adds `EphemerisQuery` which may be used to stream ephemeris elements from Horizons.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -294,7 +294,8 @@ lazy val ephemeris = project
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "org.typelevel" %% "cats-testkit" % catsVersion % "test"
+      "org.http4s"    %% "http4s-blaze-client" % http4sVersion,
+      "org.typelevel" %% "cats-testkit"        % catsVersion    % "test"
     )
   )
 
--- a/build.sbt
+++ b/build.sbt
@@ -290,14 +290,15 @@ lazy val ocs2 = project
 lazy val ephemeris = project
   .in(file("modules/ephemeris"))
   .enablePlugins(AutomateHeaderPlugin)
-  .dependsOn(coreJVM, db, sql)
+  .dependsOn(coreJVM % "compile->compile;test->test", db, sql)
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
       "co.fs2"        %% "fs2-io"              % fs2Version     % "test",
       "org.http4s"    %% "http4s-blaze-client" % http4sVersion,
       "org.typelevel" %% "cats-testkit"        % catsVersion    % "test"
-    )
+    ),
+    testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "gem.test.Tags.RequiresNetwork") // by default, ignore network tests
   )
 
 lazy val service = project
--- a/build.sbt
+++ b/build.sbt
@@ -294,6 +294,7 @@ lazy val ephemeris = project
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
+      "co.fs2"        %% "fs2-io"              % fs2Version     % "test",
       "org.http4s"    %% "http4s-blaze-client" % http4sVersion,
       "org.typelevel" %% "cats-testkit"        % catsVersion    % "test"
     )
--- a/modules/ephemeris/src/main/scala/gem/horizons/EphemerisParser.scala
+++ b/modules/ephemeris/src/main/scala/gem/horizons/EphemerisParser.scala
@@ -100,7 +100,8 @@ object EphemerisParser {
     * @return pipe for a `Stream[F, String]` into a `Stream[F, Either[String, Ephemeris.Element]]`
     */
   def eitherElements[F[_]]: Pipe[F, String, Either[String, Ephemeris.Element]] =
-    in => parsedElements(in).map(_.either)
+    _.through(parsedElements)
+     .map(_.either)
 
   /** An `fs2.Pipe` that converts a `Stream[F, String]` of ephemeris data from
     * horizons into a `Stream[F, Ephemeris.Element]`.  If there is a parse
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -290,12 +290,15 @@ lazy val ocs2 = project
 lazy val ephemeris = project
   .in(file("modules/ephemeris"))
   .enablePlugins(AutomateHeaderPlugin)
-  .dependsOn(coreJVM, db, sql)
+  .dependsOn(coreJVM % "compile->compile;test->test", db, sql)
```
swalker2m: This is done to pick up the new test case tags, which are used to skip network tests.

build.sbt
```diff
@@ -290,12 +290,15 @@ lazy val ocs2 = project
...
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "org.typelevel" %% "cats-testkit" % catsVersion % "test"
-    )
+      "co.fs2"        %% "fs2-io"              % fs2Version     % "test",
+      "org.http4s"    %% "http4s-blaze-client" % http4sVersion,
+      "org.typelevel" %% "cats-testkit"        % catsVersion    % "test"
+    ),
+    testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "gem.test.Tags.RequiresNetwork") // by default, ignore network tests
```
swalker2m: This will skip tests tagged with `RequiresNetwork`.  I couldn't figure out how to explicitly include these tests on the command line though, since you can't seem to override this value with `-n` as I had expected.

modules/ephemeris/src/test/resources/gem/horizons/mars-error.eph
```diff
@@ -0,0 +1,217 @@
...
+Table cut-offs 1: Elevation (-90.0deg=NO ),Airmass (>38.000=NO), Daylight (NO )
+Table cut-offs 2: Solar elongation (  0.0,180.0=NO ),Local Hour Angle( 0.0=NO )
+Table cut-offs 3: RA/DEC angular rate (     0.0=NO )
+*******************************************************************************
+ Date__(UT)__HR:MN:SC.fff     R.A.___(ICRF/J2000.0)___DEC
+*********************************************************
+$$SOE
+ 2017-Aug-01 00:00:00.000  m  08 39 26.1113 +19 33 27.498
+ 2017-Aug-02 20:09:36.000 *m  08 44 16.7664 +19 15 25.619
+ 2017-Aug-04 16:19:12.000 --  08 49 06.2770 +18 56 56.576
```
swalker2m: This element shouldn't parse because the solar presence flag is not correct.

modules/ephemeris/src/test/scala/gem/horizons/EphemerisParserSpec.scala
```diff
@@ -105,20 +107,64 @@ final class EphemerisParserSpec extends CatsSuite {
...
 
-  private def eph(elems: (String, String)*): TreeMap[InstantMicros, Coordinates] =
-    TreeMap(elems.map { case (i, c) => time(i) -> coords(c) }: _*)
+object EphemerisParserSpec {
+  private def inputStream(n: String): InputStream =
+    getClass.getResourceAsStream(s"$n.eph")
+
+  private def stream(n: String): Stream[IO, String] =
+    fs2.io.readInputStream(IO(inputStream(n)), 128)
+          .through(fs2.text.utf8Decode)
```
swalker2m: Something curious is going on here.  The tests are sensitive to the value for  the buffer size.  Usually you'd want a value of 8K or something, but I thought it would be good to make sure that the entire file doesn't fit in a single chunk.  If the buffer size is below 110 bytes, everything inexplicably fails.

modules/ephemeris/src/main/scala/gem/horizons/EphemerisQuery.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  // Format coordinates and altitude as expected by horizons.
+  private def formatCoords(s: Site): String =
+    f"'${s.longitude.toDoubleDegrees}%1.5f,${s.latitude.toSignedDoubleDegrees}%1.6f,${s.altitude.toDouble/1000.0}%1.3f'"
+
+  // URL encode query params.
+  private def formatQuery(params: List[(String, String)]): String =
+    params
+      .map { case (k, v) => s"$k=${URLEncoder.encode(v, UTF_8.name)}" }
+      .mkString("&")
```
tpolecat: I prefer `.intercalate("&")` because `.mkString` calls `.toString` on the elements which can be unsafe. Doesn't matter here since it's strings already, but just FYI.

modules/ephemeris/src/test/scala/gem/horizons/EphemerisParserSpec.scala
tpolecat: Hmm, weird. Can you minimize this failure?

build.sbt
swalker2m: The one that asks for 90K elements is kind of slow.  It didn't seem kosher to me to fail the build if you don't have network access or if there is an issue with Horizons.  I guess I think about these issues more than most people since I live in the country and only have spotty access.  If you guys prefer, I could turn it around and leave them on by default and then I could explicitly turn them off when I run it at home.

modules/ephemeris/src/main/scala/gem/horizons/EphemerisQuery.scala
swalker2m: Makes sense, I'll switch it.

build.sbt
tpolecat: No it's fine with me to leave them out by default. Can we pass something to sbt to make them run in CI?

swalker2m: I'll work on it some more.  I was defeated on a first pass at it.

tpolecat:
Looks fine. You might ask on sbt/sbt gitter if you can't get the test thing working.

modules/ephemeris/src/main/scala/gem/horizons/EphemerisParser.scala
```diff
@@ -47,15 +55,77 @@ object EphemerisParser {
...
+  /** An `fs2.Pipe` that converts a `Stream[F, String]` of ephemeris data from
+    * horizons into a `Stream[F, ParseResult[Ephemeris.Element]]`.
+    *
+    * @tparam F effect to use
+    *
+    * @return pipe for a `Stream[F, String]` into a `Stream[F, ParseResult[Ephemeris.Element]]`
+    */
+  def parsedElements[F[_]]: Pipe[F, String, ParseResult[Ephemeris.Element]] =
+    _.through(fs2.text.lines)
+     .dropThrough(_.trim =!= SOE)
```
swalker2m: The problem with buffer size that I reported seems to have something to do with `dropThrough`. If I switch the `dropThrough` for

```
.dropWhile(_.trim =!= SOE)
.drop(1)
```

It works with any buffer size.

tpolecat: I think fs2 would appreciate a bug report if you can make a standalone reproduction.

PR #143, opened 2017-09-19
initial target model

tpolecat:
This adds a target table and basic CRUD operations which in the end are simple but I had to do a few other things, which I will explain in inline comments.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,11 +70,11 @@ lazy val gemWarts =
   Warts.allBut(
     Wart.Any,                // false positives
     Wart.Nothing,            // false positives
-    Wart.Null,               // false positives
     Wart.Product,            // false positives
     Wart.Serializable,       // false positives
     Wart.Recursion,          // false positives
-    Wart.ImplicitConversion  // we know what we're doing
+    Wart.ImplicitConversion, // we know what we're doing
+    Wart.ImplicitParameter   // false positives
   )
 
 lazy val commonSettings = Seq(
@@ -232,10 +232,10 @@ lazy val db = project
       "org.tpolecat" %% "doobie-scalatest" % doobieVersion % "test"
     ),
     initialCommands += """
-      |import cats._, cats.data._, cats.implicits._, scalaz.effect.IO
+      |import cats._, cats.data._, cats.implicits._, cats.effect._
       |import doobie._, doobie.implicits._
-      |import gem._, gem.enum._, gem.dao._
-      |val xa = DriverManagerTransactor[IO](
+      |import gem._, gem.enum._, gem.math._, gem.dao._, gem.dao.meta._, gem.dao.composite._
+      |val xa = Transactor.fromDriverManager[IO](
       |  "org.postgresql.Driver",
       |  "jdbc:postgresql:gem",
       |  "postgres",
--- a/modules/core/shared/src/main/scala/gem/enum/EphemerisKeyType.scala
+++ b/modules/core/shared/src/main/scala/gem/enum/EphemerisKeyType.scala
@@ -16,7 +16,7 @@ sealed abstract class EphemerisKeyType(
   val tag: String,
   val shortName: String,
   val longName: String
-)
+) extends Product with Serializable
 
 object EphemerisKeyType {
 
--- a/modules/core/shared/src/main/scala/gem/math/RadialVelocity.scala
+++ b/modules/core/shared/src/main/scala/gem/math/RadialVelocity.scala
@@ -6,15 +6,28 @@ package gem.math
 import cats.{ Order, Show }
 import cats.implicits._
 import gem.math.PhysicalConstants.SpeedOfLight
+import scala.math.sqrt
 
-/** Radial Velocity represented as integral meters per second, positive if receding. */
+/**
+ * Radial Velocity represented as integral meters per second, positive if receding. We can also
+ * view velocity in terms of redshift (a dimensionless value called '''z''').
+ * @see Wikipedia on [[https://en.wikipedia.org/wiki/Radial_velocity Radial Velocity]]
+ * @see Wikipedia on [[https://en.wikipedia.org/wiki/Redshift Redshift]]
+ */
 final case class RadialVelocity(toMetersPerSecond: Int) {
 
   // Sanity check
   assert(toMetersPerSecond.abs <= SpeedOfLight, "Radial velocity exceeds the speed of light.")
 
   def toKilometersPerSecond: Double =
-    toMetersPerSecond.toDouble / 1000
+    toMetersPerSecond.toDouble / 1000.0
+
+  def toRedshift: Double = {
+    val v = toMetersPerSecond.toDouble
+    val C = SpeedOfLight.toDouble
+    val t = (1.0 + v / C) / (1.0 - v / C)
+    sqrt(t) - 1.0
+  }
 
   override def toString =
     f"RadialVelocity($toKilometersPerSecond%3.3fkm/s)"
@@ -29,6 +42,12 @@ object RadialVelocity {
   /** Construct a [[RadialVelocity]] from floating point kilometers per second. */
   def fromKilometersPerSecond(kms: Double): RadialVelocity =
     RadialVelocity((kms * 1000.0).toInt)
+
+  def fromRedshift(z: Double): RadialVelocity = {
+    val C = SpeedOfLight.toDouble
+    val m = C * ((z + 1.0) * (z + 1.0) - 1.0) / ((z + 1.0) * (z + 1.0) + 1.0)
+    RadialVelocity(m.toInt)
+  }
 
   /** Instances are ordered by their `.toMetersPerSecond` values. */
   implicit val RadialVelocityOrder: Order[RadialVelocity] =
--- a/modules/db/src/main/scala/gem/dao/TargetDao.scala
+++ b/modules/db/src/main/scala/gem/dao/TargetDao.scala
@@ -9,7 +9,6 @@ import doobie.implicits._
 import gem.dao.meta._
 import gem.dao.composite._
 import gem.enum.TrackType
-import shapeless._
 
 object TargetDao {
   import EnumeratedMeta._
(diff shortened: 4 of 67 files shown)


build.sbt
```diff
@@ -70,11 +70,11 @@ lazy val gemWarts =
   Warts.allBut(
     Wart.Any,                // false positives
     Wart.Nothing,            // false positives
-    Wart.Null,               // false positives
```
tpolecat: False positives here have gone away so I turned this one back on.

build.sbt
```diff
@@ -70,11 +70,11 @@ lazy val gemWarts =
...
     Wart.Product,            // false positives
     Wart.Serializable,       // false positives
     Wart.Recursion,          // false positives
-    Wart.ImplicitConversion  // we know what we're doing
+    Wart.ImplicitConversion, // we know what we're doing
+    Wart.ImplicitParameter   // false positives
```
tpolecat: There is an instance where I'm taking a typeclass instance but it's still triggering this warning about non-typeclass implicits. So I turned it off.

build.sbt
```diff
@@ -232,10 +232,10 @@ lazy val db = project
       "org.tpolecat" %% "doobie-scalatest" % doobieVersion % "test"
     ),
     initialCommands += """
-      |import cats._, cats.data._, cats.implicits._, scalaz.effect.IO
+      |import cats._, cats.data._, cats.implicits._, cats.effect._
       |import doobie._, doobie.implicits._
-      |import gem._, gem.enum._, gem.dao._
-      |val xa = DriverManagerTransactor[IO](
+      |import gem._, gem.enum._, gem.math._, gem.dao._, gem.dao.meta._, gem.dao.composite._
+      |val xa = Transactor.fromDriverManager[IO](
```
tpolecat: Forgot to update this for doobie 0.5

modules/core/shared/src/main/scala/gem/enum/EphemerisKeyType.scala
```diff
@@ -16,7 +16,7 @@ sealed abstract class EphemerisKeyType(
   val tag: String,
   val shortName: String,
   val longName: String
-)
+) extends Product with Serializable
```
tpolecat: This gives better LUBs, which simplifies some code that I'll note lower down.

modules/core/shared/src/main/scala/gem/math/RadialVelocity.scala
```diff
@@ -6,15 +6,28 @@ package gem.math
...
   def toKilometersPerSecond: Double =
-    toMetersPerSecond.toDouble / 1000
+    toMetersPerSecond.toDouble / 1000.0
+
+  def toRedshift: Double = {
+    val v = toMetersPerSecond.toDouble
+    val C = SpeedOfLight.toDouble
+    val t = (1.0 + v / C) / (1.0 - v / C)
+    sqrt(t) - 1.0
+  }
```
tpolecat: The redshift conversion doesn't really need to be part of this PR but I couldn't figure out how to pull it out. In any case it's simple math taken from Carlos's implementation in OCS.

modules/db/src/main/scala/gem/dao/TargetDao.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      // We map only the ephemeris key portion of the nonsidereal target here, and we only need to
+      // consider the Option[Nonsidereal] case because this is what the coproduct encoding needs.
+      implicit val compositeOptionNonsidereal: Composite[Option[Nonsidereal]] =
+        Composite[Option[EphemerisKey]].imap(_.map(Nonsidereal.empty))(_.map(_.ephemerisKey))
+
+      // Construct an encoder for track constructors, tagged by TrackType.
+      val enc = Tag[Sidereal](TrackType.Sidereal)       :+:
+                Tag[Nonsidereal](TrackType.Nonsidereal) :+: TNil
+
```
tpolecat: Without the change to enum above the tag type here is inferred as `TrackType with Product with Serializable` which doesn't have a `Meta` instance. With the change it's just `TrackType` which is what we want.

modules/db/src/main/scala/gem/dao/composite/Coordinates.scala
```diff
@@ -4,19 +4,28 @@
...
+    CoordinatesCompositeLemmas.CoordinatesComposite
+
+  implicit val CoordinatesOptionComposite: Composite[Option[Coordinates]] =
+    CoordinatesCompositeLemmas.CoordinatesOptionComposite
 
 }
 object CoordinatesComposite extends CoordinatesComposite
+
+/** Derivation of Composite instances for Coordinates. */
+private object CoordinatesCompositeLemmas {
```
tpolecat: The issue here is that we're deriving the instance implicitly but also want to *export* it implicitly, which we can't do without causing an infinite loop. So we're pushing them into a helper object and delegating to them from `CoordinatesComposite` above.

modules/db/src/main/scala/gem/dao/composite/TaggedCoproduct.scala
```diff
@@ -0,0 +1,124 @@
...
+
+package gem.dao.composite
+
+import cats._
+import cats.implicits._
+import doobie._
+import shapeless._
+import shapeless.ops.coproduct.Inject
+
+object TaggedCoproduct {
```
tpolecat: This is what took so long. It's simple in the end but took me a while to figure out how to do it.

modules/db/src/main/scala/gem/dao/TargetDao.scala
tpolecat: The implementation of `Tag` and `TNil` is below in `TaggedCoproduct.scala`.

project/plugins.sbt
```diff
@@ -10,7 +10,7 @@ addSbtPlugin("org.scalastyle"   %% "scalastyle-sbt-plugin" % "0.8.0")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-native-packager"   % "1.1.4")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-git"               % "0.8.5")
 addSbtPlugin("de.heikoseeberger" % "sbt-header"            % "2.0.0")
-addSbtPlugin("org.wartremover"   % "sbt-wartremover"       % "2.1.1")
+addSbtPlugin("org.wartremover"   % "sbt-wartremover"       % "2.1.1") // 2.2.0 PublicInference fails
```
tpolecat: There is a bug in WartRemover 2.2.0 which crashes the compiler when checking for inferred types on public members. Because this check is important we should stick with 2.1.1 for now.

modules/db/src/main/scala/gem/dao/TargetDao.scala
```diff
@@ -0,0 +1,93 @@
...
+package gem
+package dao
+
+import doobie._
+import doobie.implicits._
+import gem.dao.meta._
+import gem.dao.composite._
+import gem.enum.TrackType
+import shapeless._
```
swalker2m: Are we crossing the Rubicon here, or maybe we already did and I didn't notice.  @cquiroz any thoughts?  I guess I'd like to know whether we're going to try to only use certain parts, or only use it in certain sub-projects, or maybe we're ready to fully commit?  It seems a little less frightening after a brief intro at Scala World but it seems like one of those things that if it spreads very far there's no going back.

modules/db/src/main/scala/gem/dao/composite/Coordinates.scala
swalker2m: If you can type `implicitly` and summons an instance below, why do we need to export an implicit value in `CoordinatesComposite`?

modules/db/src/main/scala/gem/dao/TargetDao.scala
cquiroz: I totally agree. I'd like that we somehow get all up to speed on shapeless before going too far. I guess it also depends a lot on what parts we use. Just using `HList` in same places is one thing, impilict recursive typeclass derivation is quite another, and there are far more advanced (esoteric) features than that 😄

cquiroz: @tpolecat could you briefly explain what you need shapless here for? I admit I haven't read the PR, possibly it is quite obvious

modules/db/src/main/scala/gem/dao/composite/Coordinates.scala
tpolecat: Yeah, if you import the RA and Dec instances you can then get the Coordinates one for free. I just wanted to provide it as a direct import. The problem is that you can get an error like `can't find Composite[ComplexType]` and you don't know which part of it is missing and you have to go through this debugging process where you break the type up try to figure out which particular part isn't available. So I was trying to avoid that by providing composites for bigger types which means fewer imports in client code (and probably faster compilation because the instances are derived once). But I don't feel strongly about it. We can get rid of this and add the necessary "smaller" instances at the call sites.

swalker2m: No, no need to change anything.  That makes sense to me.

modules/db/src/main/scala/gem/dao/TargetDao.scala
tpolecat: On line 48 `enc.composite` has type `Composite[Sidereal :+: Nonsidereal :+: CNil]` and shapeless is needed here to provide the implicits required for the call to `unify` which extracts the value as a `Track` (which is the LUB of `Sidereal` and `Nonsidereal`).

I could probably provide this as a method on `TaggedCoproductEncoder` so we don't need shapeless at the call site, if you'd prefer to keep it under wraps.

swalker2m: > I could probably provide this as a method on TaggedCoproductEncoder so we don't need shapeless at the call site, if you'd prefer to keep it under wraps.

That would probably be good, at least until we're all confident and competent with Shapeless.

tpolecat: Ok I'll give that a try.

tpolecat:
Ok I added a method to push the `_.unify` down a layer so we don't need to import shapeless in `TargetDao`.

modules/db/src/main/scala/gem/dao/TargetDao.scala
```diff
@@ -44,8 +43,10 @@ object TargetDao {
       val enc = Tag[Sidereal](TrackType.Sidereal)       :+:
                 Tag[Nonsidereal](TrackType.Nonsidereal) :+: TNil
```
swalker2m: Hmm, from where does it pick up `:+:` without Shapeless?

tpolecat: That's a different `:+:` ... it's the little DSL for constructing tagged coproduct encoders. `TNil` is the empty one.

tpolecat:
Book available [here](https://underscore.io/books/) FYI.

PR #144, opened 2017-09-20
WIP - upgrade to sbt 1.0

tpolecat:
I don't think our build will need any changes other than updating the plugins.

 - [x] flyway-sbt https://github.com/flyway/flyway/issues/1741
 - [x] scalastyle-sbt-plugin 
 - [x] sbt-native-packager   
 - [x] sbt-git               
 - [x] sbt-header            
 - [x] sbt-wartremover       
 - [x] sbt-scalajs           
 - [x] sbt-dependency-graph  https://github.com/jrudolph/sbt-dependency-graph/issues/134
 - [x] sbt-updates

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
+++ b/modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
@@ -113,6 +113,7 @@ object ProperMotion {
    * @param elapsedYears    elapsed time in epoch years
    * @return (ra, dec) in radians, corrected for proper motion
    */
+  // scalastyle:off method.length
   private def properMotionʹ(
     baseCoordinates: Vec2,
     daysPerYear:     Double,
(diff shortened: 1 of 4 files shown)


cquiroz:
Congratulations on getting to sbt 1.0. I don't even want to think about doing that for
Ocs3

tpolecat:
I'm not convinced we want to move. I'm messing around with it in doobie and it's behaving strangely.

cquiroz:
Sbt and strange behavior is the regular state 😄. I guess for doobie you don't need to cross-sbt build but maybe for tut?

tpolecat:
We had to fork tut. There was no way to make it compile for 0.13 and 1.0

tpolecat:
sbt 1.0.3 fixes the issues I ran into, and all the plugins are now available so I think we can merge this if it goes green.

tpolecat:
Hm wonder why scalastyle just now saw that stuff.

PR #146, opened 2017-09-26
HorizonsNameQuery

swalker2m:
Adds an horizons name resolution query.  This is code extracted from the old OCS project's `HS2`, the `HorizonsService2`.  For the most part, it is just copy/pasted from there with a few necessary changes:

* Ported to `cats`
* Uses `http4s`
* Work with scalastyle checks

I also split out some common functionality between the `HorizonsEphemerisQuery` and the `HorizonsNameQuery` into `HorizonsClient`.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/EphemerisKey.scala
+++ b/modules/core/shared/src/main/scala/gem/EphemerisKey.scala
@@ -49,6 +49,10 @@ object EphemerisKey {
     */
   sealed abstract class Horizons(val queryString: String) extends EphemerisKey
 
+  /** Horizons designation for asteroids, old and new style.
+    */
+  sealed abstract class Asteroid(s: String) extends Horizons(s)
+
   /** Designation for a comet, in the current apparition. Example: `C/1973 E1`
     * for Kohoutek, yielding the query string `NAME=C/1973 E1;CAP`.
     */
@@ -59,14 +63,14 @@ object EphemerisKey {
   /** Designation for an asteroid under modern naming conventions. Example:
     * `1971 UC1` for 1896 Beer, yielding a query string `ASTNAM=1971 UC1`.
     */
-  @Lenses final case class AsteroidNew(des: String) extends Horizons(s"ASTNAM=$des")
+  @Lenses final case class AsteroidNew(des: String) extends Asteroid(s"ASTNAM=$des")
   @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
   object AsteroidNew
 
   /** Designation for an asteroid under "old" naming conventions. These are
     * small numbers. Example: `4` for Vesta, yielding a query string `4;`
     */
-  @Lenses final case class AsteroidOld(num: Int) extends Horizons(s"$num;") {
+  @Lenses final case class AsteroidOld(num: Int) extends Asteroid(s"$num;") {
     override def des: String =
       num.toString
   }
--- a/modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
+++ b/modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
@@ -5,6 +5,7 @@ package gem.horizons
 
 import gem.enum.Site
 
+import cats.ApplicativeError
 import cats.data.Reader
 import cats.effect.IO
 import cats.implicits._
@@ -76,7 +77,7 @@ object HorizonsClient {
     * raising an error if there is a problem parsing the request.
     */
   val uri: ParamReader[IO[Uri]] =
-    urlString.map(Uri.fromString).map(_.fold(IO.raiseError, IO.pure))
+    urlString.map(Uri.fromString).map(ApplicativeError[IO, Throwable].fromEither)
 
   /** Creates an http4s Request representing the request. */
   val request: ParamReader[IO[Request[IO]]] =
--- a/modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
+++ b/modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
@@ -33,9 +33,13 @@ object HorizonsClient {
   val client: Client[IO] =
     PooledHttp1Client[IO](maxTotalConnections = 1)
 
+  /** Horizons service URL. */
   val Url: String =
     "https://ssd.jpl.nasa.gov/horizons_batch.cgi"
 
+  /** DateTimeFormatter that can format instants into the string expected by
+    * the horizons service: {{{yyyy-MMM-d HH:mm:ss.SSS}}}.
+    */
   val DateFormat: DateTimeFormatter =
     DateTimeFormatter.ofPattern("yyyy-MMM-d HH:mm:ss.SSS", US).withZone(UTC)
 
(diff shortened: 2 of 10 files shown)


modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
```diff
@@ -0,0 +1,243 @@
...
+  /** Result of performing an horizons name resolution. */
+  type Result[A] = EitherT[IO, Error, A]
+
+  object Result {
+    def fromEither[A](e: Either[Error, A]): Result[A] =
+      EitherT.fromEither[IO](e)
+
+    def delay[A](a: => A): Result[A] =
+      EitherT.rightT[IO, Error](a)
+  }
```
swalker2m: This is part of the old code that is only used in the OT.  I left it here for now but I'm not sure whether we'll need it.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
```diff
@@ -0,0 +1,243 @@
...
+
+  object Search {
+    final case class Comet(partial: String)     extends Search[EphemerisKey.Comet    ](s"NAME=$partial*;CAP")
+    final case class Asteroid(partial: String)  extends Search[EphemerisKey.Asteroid ](s"ASTNAM=$partial*"  )
+    final case class MajorBody(partial: String) extends Search[EphemerisKey.MajorBody](s"$partial"          )
+  }
+
+  sealed trait Error extends Product with Serializable
+
+  final case class HorizonsException(e: Throwable)                  extends Error
```
swalker2m: Not used, yet.  I think it would only be needed if the client wants to map an error running the service.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  /** Computes the URL string that matches the given parameters. */
+  val urlString: ParamReader[String] =
+    formatQuery.map { s => s"$Url?$s" }
+
+  /** Creates an http4s Uri representing the request and lifts it into IO,
+    * raising an error if there is a problem parsing the request.
+    */
+  val uri: ParamReader[IO[Uri]] =
+    urlString.map(Uri.fromString).map(_.fold(IO.raiseError, IO.pure))
```
tpolecat: I think I would prefer `.map(ApplicativeError[IO, Throwable].fromEither)` here, despite it being more characters.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
tpolecat: I think I'd prefer deleting stuff that's unused, especially if it's trivial like this.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
```diff
@@ -0,0 +1,243 @@
...
+    // from --- -------- --- -----
+    type Offsets = List[(Int, Int)]
+
+    // Parse many results based on an expected header pattern. This will read
+    // through the tail until --- separators are found, then parse the
+    // remaining lines (until a blank line is encountered) using the function
+    // constructed by `f`, if any (otherwise it is assumed that there were not
+    // enough columns found). Returns a list of values or an error message.
+    def parseMany[A](header: String, tail: List[String], headerPattern: Regex)(f: Offsets => Option[String => A]): Either[String, List[A]] =
+      headerPattern.findFirstMatchIn(header).toRight("Header pattern not found: " + headerPattern.regex) *>
```
tpolecat: This code makes me really nervous but it was the best I could do at the time. It would be nice to revisit this at some point and see if it can be made more obvious. Ok for now though.

modules/core/shared/src/main/scala/gem/EphemerisKey.scala
```diff
@@ -49,6 +49,10 @@ object EphemerisKey {
     */
   sealed abstract class Horizons(val queryString: String) extends EphemerisKey
 
+  /** Horizons designation for asteroids, old and new style.
+    */
+  sealed abstract class Asteroid(s: String) extends Horizons(s)
```
jdnavarro: In this particular case why `Asteroid` is not a `trait`? It's still not clear to me when to choose a trait or an abstract class.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
```diff
@@ -0,0 +1,235 @@
...
+      def case1: ParsedMajorBodies =
+        """  +(.*?) / \((.+?)\)  +(\d+) *$""".r.findFirstMatchIn(header).map { m =>
+          List(Resolution(EphemerisKey.MajorBody(m.group(3).toInt), m.group(1)))
+        }.toRight("Could not match 'Charon / (Pluto)     901' header pattern.")
+
+      // First one that works, otherwise Nil because it falls through to small-body search
+      case0 orElse
+      case1 orElse Nil.asRight
+    }
+  }
```
jdnavarro: I guess the Horizons service doesn't offer a more orthodox format like JSON. Is the Horizons format formally documented somewhere? Why not use `atto` in this case?

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
```diff
@@ -0,0 +1,235 @@
...
+    final case class MajorBody(partial: String) extends Search[EphemerisKey.MajorBody](s"$partial"          )
+  }
+
+  sealed trait Error extends Product with Serializable
+
+  final case class HorizonsException(e: Throwable)                  extends Error
+  final case class ParseError(input: List[String], message: String) extends Error
+
+  /** Result of performing an horizons name resolution. */
+  type Result[A] = EitherT[IO, Error, A]
```
jdnavarro: Some would argue this `EitherT IO` monad transformer stack is an [anti-pattern](https://www.schoolofhaskell.com/user/commercial/content/exceptions-best-practices#the-bad). I don't have a very strong opinion on the matter, but I'd like to see some some code snippet where the `Error` value and a raised IO exception are being handled from a `Result`.

modules/core/shared/src/main/scala/gem/EphemerisKey.scala
swalker2m: It can't be a trait because it extends an abstract class.  The `Horizons` class it extends is an abstract class because it contains a `queryString` field.  That could have been specified as a method and then implemented in the various instances but this seemed easier.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
swalker2m: I'm putting things together from the bottom-up.  There aren't really clients yet. This code was ported from some original work in the OCS package that used `EitherT`.  If there's a better way to do error handling though, we can change it.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsNameQuery.scala
swalker2m: Unfortunately Horizons doesn't really offer formatted data in its responses and we end up having to parse information that was meant to be read by a human.  In this case the various cases are tricky so I opted to just keep the regex parsing solution that we've had in place for a few years instead of trying to re-implement it.

PR #148, opened 2017-09-29
ephemeris_meta for update tracking

swalker2m:
The PR adds a table for tracking information related to ephemeris updates.  In particular it records the last update time, the last update check time, and the horizons solution reference for each ephemeris key.  At the moment, we only have solution reference information for comets and asteroids so this may change if we get more information from horizons about major body versions.  User-supplied ephemeris data also won't have a solution reference of course.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/EphemerisMeta.scala
+++ b/modules/core/shared/src/main/scala/gem/EphemerisMeta.scala
@@ -1,0 +1,31 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+import gem.util.InstantMicros
+
+import cats.{ Eq, Show }
+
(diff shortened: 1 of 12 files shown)


swalker2m:
Obviously the ephemeris elements in the `ephemeris` table are related to those in `ephemeris_meta` but nothing enforces this.  Missing meta data for a given key just means that an update is required.  The opposite situation, missing ephemeris elements when there is meta data, means we have uncollected garbage.  I'm not sure whether we should do anything about this at the data model level though.  For example, `ephemeris` elements could reference the corresponding `ephemeris_meta` but I'm not sure whether it's worth the cost of that.

tpolecat:
If both tables referenced `target` somehow then it would prevent unmoored ephemerides and metadata, but maybe we can do that in another PR. I think it might be harder than usual because it's not a key relationship.

tpolecat:
We could restructure the target table so it has slices for sidereal and nonsidereal, which would make the key relationship here straightforward. I'm open to doing that.

PR #149, opened 2017-10-04
Update atto, doobie, http4s, wartremover

tpolecat:
There were some changes to the http4s API that are good simplifications but I had to refactor `Gatekeeper` a little bit. The new version of WartRemover is a lot better. There's still one false positive for `PublicInference` which I suppressed.

### DIFF
--- a/modules/web/src/main/scala/Gatekeeper.scala
+++ b/modules/web/src/main/scala/Gatekeeper.scala
@@ -4,14 +4,17 @@
 package gem
 package web
 
+import cats.data._
 import cats.effect.IO
+import cats.implicits._
 import gem.json._
 import gem.{ Service => GemService }
 import io.circe.generic.auto._
 import java.time.Instant
 import org.http4s._
 import org.http4s.circe._
-import org.http4s.dsl._
+import org.http4s.dsl.io._
+import org.http4s.implicits._
 import org.http4s.server._
 
 /**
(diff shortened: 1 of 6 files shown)


modules/web/src/main/scala/Gatekeeper.scala
```diff
@@ -79,40 +82,50 @@ object Gatekeeper {
...
-
     }
 
+  /**
+   * Construct the gatekeeper middleware. We need a server environment to figure out how to do
+   * do this because we need a way to encode/decode tokens, as well as way to log users in.
+   */
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  def apply(env: Environment): AuthMiddleware[IO, GemService[IO]] = delegate =>
+    login(env) <+> authenticate(env, delegate)
```
swalker2m: What does this do?  I can't find `<+>` except in scalaz docs.

tpolecat: `<+>` is append for `SemigroupK` (called `Plus` in scalaz), which is like `|+|` for things of kind `* -> *` that ignores the element type. So for `Option[A]` you can say `a |+| b` given `Semigroup[A]` and it smashes the values together if both are `Some`. But `a <+> b` means `a orElse b` … it doesn't require `Semigroup[A]` because it never looks at the elements. 

For `HttpService[IO]`, which is an alias for `Kleisli[OptionT[IO, ?], Request[IO], Response[IO]]` appending `a <+> b` means "send the request to `a` and if it yields `None` then try again with `b`", so it has the same hand-wavey meaning as `SemigroupK[Option]`; i.e., try this and if it doesn't work try the other thing. So this lets you compose (and prioritize) big routes out of smaller ones. Cachai?

cquiroz: That used to be done with `|+|` earlier but it was changed for same reason on http4s

tpolecat: Yeah the default instance from cats is actually wrong … we're fixing it. https://github.com/typelevel/cats/issues/1928

PR #150, opened 2017-10-05
Ephemeris Element Velocity

swalker2m:
This PR incorporates rate of change information associated with ephemeris elements from horizons.

```sql
ALTER TABLE ephemeris
   ADD COLUMN delta_ra  NUMERIC(10,6) NOT NULL,
   ADD COLUMN delta_dec NUMERIC(10,6) NOT NULL;

COMMENT ON COLUMN ephemeris.delta_ra  IS 'arcsec / hour * cos(dec)';
COMMENT ON COLUMN ephemeris.delta_dec IS 'arcsec / hour';
```

It does so for two purposes:

1. This information is required by the TCS and we will need it when we write out its 1440 line ephemeris files for each night.

2. We can use this information to drastically reduce the number of ephemeris elements that need to be stored, following an [analysis by @andrewwstephens](https://docs.google.com/document/d/1UeKXhgYjt7q2asYWBlhuoLYNKb_ee2ru0_YpcRVxSzM/edit).  In particular, we only keep subsequent elements that differ in velocity by more than 0.1 arcseconds and rely on interpolation to fill in any missing gaps. 

A new type, `EphemerisCoordinates`, tracks with velocity information.  Whereas before, `Ephemeris` was essentially a `Map[InstantMicros, Coordinates]` and an `Ephemeris.Element` was the pair `(InstantMicros, Coordinates)`, now `Ephemeris` is a `Map[InstantMicros, EphemerisCoordinates]` and `Ephemeris.Element` changes correspondingly. 

`EphemerisCoordinates` are just a product of `Coordinates` and an `Offset` where the `Offset` is simply understood to be a tracking rate.  In horizons data it is arcsec per hour.

I'm not 100% satisfied with this work for a number of reasons:

* The lack of unit information in the tracking rate is disturbing.
* I suspect that we could do a better, more accurate job of minimizing the ephemeris data.
* My results don't quite line up with Andy's in one case in particular (asteroid 2014 UR)


modules/ephemeris/src/main/scala/gem/horizons/EphemerisParser.scala
```diff
@@ -87,7 +96,8 @@ object EphemerisParser {
     */
   def parsedElements[F[_]]: Pipe[F, String, ParseResult[Ephemeris.Element]] =
     _.through(fs2.text.lines)
-     .dropThrough(_.trim =!= SOE)
+     .dropWhile(_.trim =!= SOE)
+     .drop(1)
```
swalker2m: This *should* be equivalent but I believe there is an [FS2 bug](https://github.com/functional-streams-for-scala/fs2/issues/934) here.

modules/ephemeris/src/main/scala/gem/horizons/VelocityCompression.scala
```diff
@@ -0,0 +1,41 @@
...
+    * @param µas difference in velocity below this threshold result in
+    *            skipping elements
+    */
+  def velocityCompression[F[_]](µas: Long): Pipe[F, Ephemeris.Element, Ephemeris.Element] =
+    _.zipWithNext
+     .map { case (e, on) => (e, e._2.velocity, on.isEmpty) }
+     .filterWithPrevious { case ((_, pv, _), (_, v, last)) =>
+       // always include the last element regardless of Δ velocity
+       last || (pv.toMicroarcseconds - v.toMicroarcseconds).abs >= µas
+     }.map(_._1)
```
swalker2m: Here is where we filter out elements whose delta velocity is less than the threshold of 0.1 arcseconds.

modules/ephemeris/src/test/scala/gem/horizons/VelocityCompressionSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    assert(actual == expected)
+  }
+
+  // Andy's result: 99%
+  test("2014 UR compresses to 79% of original size", RequiresNetwork) {
+    val start = LocalDate.of(2015, Month.SEPTEMBER, 15)
+    val end   = LocalDate.of(2015, Month.NOVEMBER,  15)
+
+    testCompression("2014 UR", ur_2014, 0.79, start, end)
```
swalker2m: This is the test case that differs the most from Andy's results.  I need to see the exact test setup he used to figure out what is going on.

modules/ephemeris/src/main/scala/gem/horizons/VelocityCompression.scala
```diff
@@ -0,0 +1,41 @@
...
+  val StandardVelocityLimitµas: Long =
+    Angle.fromMilliarcseconds(100).toMicroarcseconds
+
+  /** An `fs2.Pipe` that passes only elements which differ by more than the
+    * provided delta velocity.
+    *
+    * @param µas difference in velocity below this threshold result in
+    *            skipping elements
+    */
+  def velocityCompression[F[_]](µas: Long): Pipe[F, Ephemeris.Element, Ephemeris.Element] =
```
cquiroz: neat!

jdnavarro:
I know nothing on the subject but I'm guessing PostgreSQL will do some kind of automatic compression of the data when storing it to disk and that there may be a way to tweak the compression settings for a specific table or similar.

I don't think any compression algorithm will beat the actual removal of large amount of unneeded data like you are doing here, but I wonder what would be the compression ratio when naively using a state-of-the-art ([lzo](https://en.wikipedia.org/wiki/Lempel–Ziv–Oberhumer) or [zstd](https://facebook.github.io/zstd/)) on the tables representing the old `Ephemeris` maps. Also, I'd be curious to know what would be the effect of using compression at the filesystem/VM image level.

tpolecat:
I would be interested to see how many elements we need for best and worst cases. How many points would we need for something under a lot of acceleration like Titan, vs something pokey like Pluto?

swalker2m:
Andy's document goes in to detail comparing best and worst cases.  Here are the numbers I come up for some of his examples.  Titan is actually fairly constant and compresses down drastically.  2014 UR on the other hand changes rapidly. 

```
2014 UR    AsteroidNew(2014 UR)   => 69689 / 87840 elements, 79.34% of original
2015 QT3   AsteroidNew(2015 QT3)  => 29310 / 87840 elements, 33.37% of original
Churyumov  Comet(67P)             =>  2721 / 89280 elements,  3.05% of original
Titan      MajorBody(606)         =>   502 / 89280 elements,  0.56% of original
Beer       AsteroidNew(1971 UC1)  =>  3247 / 87840 elements,  3.70% of original
Io         MajorBody(501)         => 20489 / 87840 elements, 23.33% of original
```

tpolecat:
Hm yeah worst case is pretty bad because those numbers are for two months, not the whole semester. But storage is cheap, Postgres can handle billions of rows no problem, and we'll never need to have the whole thing in memory so I still wonder if it's worth worrying about. I don't think it's worth complicating the design until we know for sure that it's an issue.

swalker2m:
> Where will you hook up the pipe fitting?

Coming soon.  I think the ephemeris cron job (or whatever takes its place) will need to do a few max element queries to cover the semester.  You can append them and then pass it through velocity compression: `Monoid.combineAll(streams).through(standardVelocityCompression)`.  When I do this for the 2015A semester for IO I get:

```
Io         MajorBody(501)        => 76658 / 349920 elements, 21.91% of original
```

swalker2m:
>  I don't think it's worth complicating the design until we know for sure that it's an issue.

This was my inclination too, until the migration issue came up and I also realized we need the velocity information in the model and in the database in order to do the TCS ephemeris files.  The only additional complication is whether to pass the element stream through the velocity compression pipe at the end.  We can still chose not to do that.  I'll go ahead and merge this unless there are objections.

tpolecat:
No objections.

PR #152, opened 2017-10-16
Preparing for Ephemeris Updater

swalker2m:
This is a collection of a few commits related only in that they prepare the way for an Ephemeris Updater program that ties together all the pieces in previous PRs.  In particular it:

* Improves streaming in `EphemerisDao`, avoiding multiple transactions and adding a streaming update.

* Adds a `paging` solution to `HorizonsEphemerisQuery`.  Paging allows the caller to specify a start and end time and a desired time interval between elements, returning a `List[HorizonsEphemerisQuery]` that would be necessary to obtain the all the data.  Note, since horizons caps results at around 90K elements, this cannot always be done in a single query.  Paging will allow us to get, say, 1 minute ephemeris data for an entire semester + padding.

* Increases the response header timeout on queries since I started to see timeouts.

* Adds lenses to `EphemerisMeta`, since that will be updated frequently.


modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisQuery.scala
```diff
@@ -83,76 +117,100 @@ object HorizonsEphemerisQuery {
...
+  def pagingCurrentSemester(
+        key:      HorizonsDesignation,
+        site:     Site,
+        step:     Duration,
+        padding:  Period): IO[List[HorizonsEphemerisQuery]] =
+
+    Semester.current(site).map(pagingSemester(key, site, _, step, padding))
+
+  def streamAll(qs: List[HorizonsEphemerisQuery]): Stream[IO, Ephemeris.Element] =
+    Monoid.combineAll(qs.map(_.streamEphemeris))
```
tpolecat: This should be equivalent to `qs.foldMap(_.streamEphemeris)`

swalker2m: Right, thanks.  I'll update it in the next one.

PR #153, opened 2017-10-18
EphemerisUpdater

swalker2m:
This PR offers the culmination of the previous `ephemeris?` PRs in that it provides a utility for creating/updating an ephemeris as necessary.  The `HorizonsEphemerisUpdater` is the main objective, but in order to exercise it I included a test application in `src/test/scala` that can be launched from SBT.  I don't think the test app will survive and ultimately we'll want to do periodic updates from, say, a cron job.  Perhaps we should rework it into a telnetd command.

The logging and configuration is a bit awkward so I'd especially welcome any feedback about that.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,6 +27,8 @@ git.uncommittedSignifier in ThisBuild := Some("UNCOMMITTED")
 // check for library updates whenever the project is [re]load
 onLoad in Global := { s => "dependencyUpdates" :: s }
 
+cancelable in Global := true
+
 // some extra commands for us
 addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum; headerCreate")
 addCommandAlias("schemaSpy", "sql/runMain org.schemaspy.Main -t pgsql -port 5432 -db gem -o modules/sql/target/schemaspy -u postgres -host localhost -s public")
--- a/modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
+++ b/modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
@@ -66,7 +66,11 @@ final case class HorizonsEphemerisUpdater(
       time <- InstantMicros.now
       sem  <- Semester.current(site)
       ctx  <- report(key, site)
-      _    <- calcUpdate(ctx, time, sem).transact(xa)
+      _    <- (if (ctx.isUpToDateFor(sem))
+                 recordUpdateCheck(ctx, time)
+               else
+                 doUpdate(ctx, time, sem)
+              ).transact(xa)
 
     } yield ()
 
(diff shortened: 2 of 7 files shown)


modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
```diff
@@ -0,0 +1,191 @@
...
+
+  private def streamEphemeris(
+                key:      EphemerisKey.Horizons,
+                site:     Site,
+                semester: Semester): Stream[ConnectionIO, Ephemeris.Element] = {
+
+    val qs = HorizonsEphemerisQuery.pagingSemester(key, site, semester, StepSize, Padding)
+
+    qs.foldMap(_.streamEphemeris)
+      .through(standardVelocityCompression)
```
tpolecat: image

modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  private def recordUpdateCheck(ctx:  Context,
+                                time: InstantMicros): ConnectionIO[Unit] =
+
+    for {
+      _ <- logMessage(s"${ctx.key}@${ctx.site} already up-to-date. Record update check.")
+      _ <- ctx.meta.fold(().pure[ConnectionIO]) { m =>
+            updateMeta(ctx.key, ctx.site, EphemerisMeta.lastUpdateCheck.set(time)(m))
+          }
```
tpolecat: There should be a combinator for this. You can say `OptionT(ctx.meta).semiflatMap(m => ...).value` but it's not really much better.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
```diff
@@ -0,0 +1,191 @@
...
+
+  private def updateMeta(key: EphemerisKey.Horizons, site: Site, m: EphemerisMeta): ConnectionIO[Unit] =
+    log.log(user, s"updateMeta($key, $site, $m)") {
+      EphemerisDao.updateMeta(key, site, m).void
+    }
+
+  private def streamEphemeris(
+                key:      EphemerisKey.Horizons,
+                site:     Site,
+                semester: Semester): Stream[ConnectionIO, Ephemeris.Element] = {
```
tpolecat: I like to wrap long signatures like

```scala
private def streamEphemeris(
  key:      EphemerisKey.Horizons,
  site:     Site,
  semester: Semester
): Stream[ConnectionIO, Ephemeris.Element] = {

  val qs = HorizonsEphemerisQuery
  ...
```

but it's not worth arguing about. Just putting it out there as an option.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisUpdater.scala
```diff
@@ -0,0 +1,191 @@
...
+
+      rnge <- log.log(user, s"EphemerisDao.selectTimes($key, $site)") {
+                EphemerisDao.selectTimes(key, site)
+              }.transact(xa)
+
+      soln <- log.log(user, s"HorizonsSolutionRefQuery($key).lookup") {
+                HorizonsSolutionRefQuery(key).lookup.liftIO[ConnectionIO]
+              }.transact(xa)
+
+    } yield Context(key, site, meta, rnge, soln)
```
tpolecat: So, as this stands we have three transactions here and I'm not sure if that's what we want. It seems like at least the first two operations should be done at once. The third operation is only in `ConnectionIO` to placate the logger so it need not have a connection at all. We can leave it as-is for now but we should open an issue to fix it. I tried but ran into a cats bug I think.

build.sbt
```diff
@@ -27,6 +27,8 @@ git.uncommittedSignifier in ThisBuild := Some("UNCOMMITTED")
 // check for library updates whenever the project is [re]load
 onLoad in Global := { s => "dependencyUpdates" :: s }
 
+cancelable in Global := true
```
cquiroz: 👍

PR #154, opened 2017-10-23
Ephemeris Compression Round 2

swalker2m:
This PR makes a couple of changes to ephemeris "compression", besides renaming it to `EphemerisCompression` instead of `VelocityCompression`:

1. Re-implements the delta velocity approach using the same method that @andrewwstephens uses.  The compression numbers now match his.

2. Add his average acceleration-based approach that produces smaller ephemeris that are effectively no less accurate.

I think we can probably delete the original delta velocity algorithm eventually but I thought I'd let it settle in for a bit first.

```
2014 UR Δv   AsteroidNew(2014 UR)  => 87793 / 87840 elements, 99.95% of original
2014 UR a    AsteroidNew(2014 UR)  =>  8881 / 87840 elements, 10.11% of original
2015 QT3 Δv  AsteroidNew(2015 QT3) => 30071 / 87840 elements, 34.23% of original
2015 QT3 a   AsteroidNew(2015 QT3) =>  4960 / 87840 elements,  5.65% of original
Churyumov Δv Comet(67P)            =>  3446 / 89280 elements,  3.86% of original
Churyumov a  Comet(67P)            =>  1116 / 89280 elements,  1.25% of original
Titan Δv     MajorBody(606)        =>   682 / 89280 elements,  0.76% of original
Titan a      MajorBody(606)        =>   504 / 89280 elements,  0.56% of original
Beer Δv      AsteroidNew(1971 UC1) =>  4070 / 87840 elements,  4.63% of original
Beer a       AsteroidNew(1971 UC1) =>  1210 / 87840 elements,  1.38% of original
Io Δv        MajorBody(501)        => 20848 / 87840 elements, 23.73% of original
Io a         MajorBody(501)        =>  2844 / 87840 elements,  3.24% of original
```


modules/core/shared/src/main/scala/gem/syntax/Stream.scala
```diff
@@ -0,0 +1,66 @@
...
+
+package gem.syntax
+
+import fs2.{ Chunk, Pull, Stream }
+
+final class StreamOps[F[_], O](val self: Stream[F, O]) {
+
+  /** A filter/fold combination that uses the accumulated result of folding
+    * over the elements since the last element was emitted to determine whether
+    * the current element under examination should be emitted.  The filter
```
swalker2m: Ok, this is admittedly a bit whacky.  I could be easily convinced that it doesn't really belong here and should be relegated to the ephemeris package where it is actually used.

tpolecat: Looks good. I think you should PR this back to fs2 since it's completely generic, and then we can then delete it from our codebase.

modules/ephemeris/src/main/scala/gem/horizons/EphemerisCompression.scala
```diff
@@ -0,0 +1,147 @@
...
+    * provided delta velocity.
+    *
+    * @param µas difference in velocity below this threshold result in
+    *            skipping elements
+    */
+  def velocityCompression[F[_]](µas: Long): Pipe[F, Ephemeris.Element, Ephemeris.Element] = {
+    def Δv(e0: Element, e1: Element): Long = {
+      val dp = e0.µasP - e1.µasP
+      val dq = e0.µasQ - e1.µasQ
+      Math.sqrt(dp * dp + dq * dq).round
```
tpolecat: I think `Math.hypot(dp, dq)` is more efficient.

modules/ephemeris/src/main/scala/gem/horizons/EphemerisCompression.scala
```diff
@@ -0,0 +1,147 @@
...
+      // time since previous
+      val d = Duration.between(prev._1.toInstant, cur._1.toInstant)
+
+      // time since previous as decimal hours
+      val h = d.toMillis.toDouble / Duration.ofHours(1).toMillis.toDouble
+
+      val dp = (cur.µasP - prev.µasP)/h
+      val dq = (cur.µasQ - prev.µasQ)/h
+
+      AvgAcceleration(math.sqrt(dp*dp + dq*dq), h)
```
tpolecat: `hypot` here as well, I think.

PR #155, opened 2017-11-05
cats 1.0.0-RC1, remove mouse

tpolecat:
This updates gem to cats-1.0.0-RC1, which was all mechanical to accommodate some re-organizing in cats for consistency. Mouse has not yet been released for cats-1.0.0-RC1 so I removed the dependency and added a string syntax class, but we can add it back later if it seems like we need it.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/EphemerisKey.scala
+++ b/modules/core/shared/src/main/scala/gem/EphemerisKey.scala
@@ -6,6 +6,7 @@ package gem
 import gem.enum.EphemerisKeyType
 import gem.parser.EphemerisKeyParsers
 import gem.syntax.parser._
+import gem.syntax.string._
 
 import cats.{ Eq, Show }
 import monocle.macros.Lenses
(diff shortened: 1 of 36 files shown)


cquiroz:
There is some scalacheck build failure

modules/core/shared/src/main/scala/gem/EphemerisKey.scala
```diff
@@ -6,6 +6,7 @@ package gem
 import gem.enum.EphemerisKeyType
 import gem.parser.EphemerisKeyParsers
 import gem.syntax.parser._
+import gem.syntax.string._
```
cquiroz: So we use `mouse` only for strings?

tpolecat: Yes, and it doesn't provide everything we need (`parseBigDecimal` for instance). I would consider contributing all this back but for now it's so minor it seems like less of a hassle to just maintain it ourselves. We can contribute general stuff back once we have a better idea of what we're going to need.

Issue #88, opened 2017-07-24
observations should not need to know the program they're in


tpolecat:
In the construction [here](https://github.com/gemini-hlsw/gem/blob/master/modules/db/src/test/scala/gem/dao/StepDaoSpec.scala#L20-L47) we see that the program id is repeated in each observation, which is a potential consistency problem. There is also the opportunity for consistency issues with the indices, which must be unique.

So maybe something like

```scala

case class Observation[+S, +D](title: String, staticConfig: S, steps: List[D])

case class Program[+A](id: Program.Id, title: String, observations: Map[Int, A]) {
  def observationsWithIds: List[(Observation.Id, A)]
}

```

So if you need to know an observation's id you need to pass a pair around.

swalker2m:
Also, if the Observation definitions do not include IDs they can be directly shared across libraries and multiple programs perhaps.

swalker2m:
You weren't thinking this would extend down to the schema data model?  That would require a program/observation link table and make observation queries awkward I would think.

PR #160, opened 2017-11-23
Remove id from Observation

swalker2m:
This is a fairly mechanical refactoring to implement issue #88.  The idea is that observations should not keep up with their own ids because it references the program id and creates an opportunity for inconsistency--a program could contain an observation whose id references a different program.

Similarly, a program should keep its observations in a map keyed by number rather than in a `List`, since nothing stops two observations with the same number from appearing in today's model (or the same observation appearing twice).

I didn't change the database model because I think it probably makes sense to keep the observation id in the observation table.  Otherwise queries involving observations become more complicated.

I chose a `TreeMap` instead of a hash map in order to keep a sensible `Traverse[Program]` implementation.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Observation.scala
+++ b/modules/core/shared/src/main/scala/gem/Observation.scala
@@ -3,7 +3,11 @@
 
 package gem
 
-import cats.{ Applicative, ApplicativeError, Eval, Order, Show, Bitraverse }
+import gem.syntax.string._
+
+import mouse.boolean._
+
+import cats.{ Applicative, Eval, Order, Show, Bitraverse }
 import cats.implicits._
 
 /**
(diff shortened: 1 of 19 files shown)


modules/core/shared/src/main/scala/gem/Observation.scala
```diff
@@ -14,17 +18,45 @@ import cats.implicits._
...
   title: String,
   staticConfig: S,
   steps: List[D])
 
 object Observation {
 
+  /** A positive, non-zero integer for use in ids. */
+  sealed abstract case class Index(toInt: Int) {
+    def format: String =
+      s"$toInt"
```
sraaphorst: Just curious if there's some reason you did it this was instead of `toInt.toString`?

cquiroz: I bet because `WartRemover` complains about `toString`

cquiroz: i've been defining lots of `Show.fromToString` to get around that `WartRemover` feature. I don't think it's that helpful

PR #161, opened 2017-11-29
Add magnitude band

jdnavarro:
In the end I left out the brightness units that seemed related to ITC. This is good enough for now to continue with GNIRS for now.

You may want to check [MagnitudeBand.scala](https://github.com/gemini-hlsw/ocs/blob/8966cd9222294505cd1e661cceb536ae02238471/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala) in OCS.

### DIFF
--- a/modules/sql/src/main/resources/db/migration/V033__Magnitude_Band.sql
+++ b/modules/sql/src/main/resources/db/migration/V033__Magnitude_Band.sql
@@ -1,0 +1,60 @@
+--
+-- Name: magnitude_system; Type: TYPE; Schema: public; Owner: postgres
+--
+
+-- Note: this may need to be turned into a table once we start taking into
+-- account the units needed for ITC Watts and Ergs.
+CREATE TYPE magnitude_system AS ENUM (
+    'Vega',
+    'AB',
+    'Jy'
+);
+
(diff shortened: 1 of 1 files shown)


jdnavarro:
I just realized I missed the `AP` band. I'm adding it now.

modules/sql/src/main/resources/db/migration/V033__Magnitude_Band.sql
```diff
@@ -0,0 +1,59 @@
...
+
+COMMENT ON COLUMN e_magnitude_band.center IS 'nm';
+COMMENT ON COLUMN e_magnitude_band.width  IS 'nm';
+
+--
+-- Data for Name: e_magnitude_band; Type: TABLE DATA; Schema: public; Owner: postgres
+--
+
+COPY e_magnitude_band (id, short_name, long_name, center, width, default_system) FROM stdin;
+SloanU	u	UV	356	46	AB
```
cquiroz: I hadn't realized that you could just put the values as plain text here. neat

sraaphorst: I didn't know that either. Cool.

PR #163, opened 2017-11-30
Initial porting of GNIRS parameters

jdnavarro:
I expect this to be further tweaked as we write the Shapeless records. I also have some small doubts about the instrument that I need to ask Andy, the most important one will be about the constrains for when cross-dispersed is on. But in any case I hope this is good enough to get started with GNIRS.

I focused on [GNIRSParams](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java) to get the parameters, but let me know if there are some other places where more parameters are defined.

Don't mind the commit messages, I'll squash-merge.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/enum/GnirsCamera.scala
+++ b/modules/core/shared/src/main/scala/gem/enum/GnirsCamera.scala
@@ -1,0 +1,49 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package enum
+
+import cats.instances.string._
+import cats.syntax.eq._
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for GNRIS Camera.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GnirsCamera(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val pixelScale: BigDecimal
+) extends Product with Serializable
+
+object GnirsCamera {
+
+  /** @group Constructors */ case object ShortBlue extends GnirsCamera("ShortBlue", "Short blue", "Short blue camera", 0.15)
+  /** @group Constructors */ case object LongBlue extends GnirsCamera("LongBlue", "Long blue", "Long blue camera", 0.05)
+  /** @group Constructors */ case object ShortRed extends GnirsCamera("ShortRed", "Short red", "Short red camera", 0.15)
+  /** @group Constructors */ case object LongRed extends GnirsCamera("LongRed", "Long red", "Long red camera", 0.05)
+
+  /** All members of GnirsCamera, in canonical order. */
+  val all: List[GnirsCamera] =
+    List(ShortBlue, LongBlue, ShortRed, LongRed)
+
+  /** Select the member of GnirsCamera with the given tag, if any. */
+  def fromTag(s: String): Option[GnirsCamera] =
+    all.find(_.tag === s)
+
+  /** Select the member of GnirsCamera with the given tag, throwing if absent. */
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def unsafeFromTag(s: String): GnirsCamera =
+    fromTag(s).getOrElse(throw new NoSuchElementException(s))
+
+  /** @group Typeclass Instances */
+  implicit val GnirsCameraEnumerated: Enumerated[GnirsCamera] =
+    new Enumerated[GnirsCamera] {
+      def all = GnirsCamera.all
+      def tag(a: GnirsCamera) = a.tag
+    }
+
+}
--- a/modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
+++ b/modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
@@ -1,0 +1,52 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package enum
+
+import cats.instances.string._
+import cats.syntax.eq._
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for GNRIS Decker.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GnirsDecker(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val obsolete: Boolean
+) extends Product with Serializable
+
+object GnirsDecker {
+
+  /** @group Constructors */ case object Acquisition extends GnirsDecker("Acquisition", "Acquisition", "Acquisition", false)
+  /** @group Constructors */ case object PupilViewer extends GnirsDecker("PupilViewer", "Pupil", "Pupil viewer", false)
+  /** @group Constructors */ case object ShortCamLongSlit extends GnirsDecker("ShortCamLongSlit", "Shot camera slit", "Short camera long slit", false)
+  /** @group Constructors */ case object ShortCamCrossDispersed extends GnirsDecker("ShortCamCrossDispersed", "Short camera XD", "Short camera cross dispersed", false)
+  /** @group Constructors */ case object Ifu extends GnirsDecker("Ifu", "IFU", "Integral field unit", true)
+  /** @group Constructors */ case object LongCamLongSlit extends GnirsDecker("LongCamLongSlit", "Long camera slit", "Long camera long slit", false)
+  /** @group Constructors */ case object Wollaston extends GnirsDecker("Wollaston", "Wollaston", "Wollaston", true)
+
+  /** All members of GnirsDecker, in canonical order. */
+  val all: List[GnirsDecker] =
+    List(Acquisition, PupilViewer, ShortCamLongSlit, ShortCamCrossDispersed, Ifu, LongCamLongSlit, Wollaston)
+
+  /** Select the member of GnirsDecker with the given tag, if any. */
+  def fromTag(s: String): Option[GnirsDecker] =
+    all.find(_.tag === s)
+
+  /** Select the member of GnirsDecker with the given tag, throwing if absent. */
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def unsafeFromTag(s: String): GnirsDecker =
+    fromTag(s).getOrElse(throw new NoSuchElementException(s))
+
+  /** @group Typeclass Instances */
+  implicit val GnirsDeckerEnumerated: Enumerated[GnirsDecker] =
+    new Enumerated[GnirsDecker] {
+      def all = GnirsDecker.all
+      def tag(a: GnirsDecker) = a.tag
+    }
+
+}
--- a/modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
+++ b/modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
@@ -1,0 +1,61 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package enum
+
+import cats.instances.string._
+import cats.syntax.eq._
+import gem.math.Wavelength
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for GNRIS Disperser Order.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GnirsDisperserOrder(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val count: Int,
+  val defaultWavelength: Wavelength,
+  val minWavelength: Wavelength,
+  val maxWavelength: Wavelength,
+  val deltaWavelength: Wavelength,
+  val band: Option[MagnitudeBand],
+  val cross_dispersed: Boolean
+) extends Product with Serializable
+
+object GnirsDisperserOrder {
+
+  /** @group Constructors */ case object One extends GnirsDisperserOrder("One", "1", "One", 1, Wavelength.unsafeFromAngstroms(48500), Wavelength.unsafeFromAngstroms(43000), Wavelength.unsafeFromAngstroms(60000), Wavelength.unsafeFromAngstroms(0), Some(MagnitudeBand.M), false)
+  /** @group Constructors */ case object Two extends GnirsDisperserOrder("Two", "2", "Two", 2, Wavelength.unsafeFromAngstroms(34000), Wavelength.unsafeFromAngstroms(27000), Wavelength.unsafeFromAngstroms(43000), Wavelength.unsafeFromAngstroms(0), Some(MagnitudeBand.L), false)
+  /** @group Constructors */ case object Three extends GnirsDisperserOrder("Three", "3", "Three", 3, Wavelength.unsafeFromAngstroms(22200), Wavelength.unsafeFromAngstroms(18600), Wavelength.unsafeFromAngstroms(27000), Wavelength.unsafeFromAngstroms(6), Some(MagnitudeBand.K), true)
+  /** @group Constructors */ case object FourXD extends GnirsDisperserOrder("FourXD", "4XD", "FourXD", 4, Wavelength.unsafeFromAngstroms(16500), Wavelength.unsafeFromAngstroms(14200), Wavelength.unsafeFromAngstroms(18600), Wavelength.unsafeFromAngstroms(5), Some(MagnitudeBand.H), true)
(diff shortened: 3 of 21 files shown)


cquiroz:
I'm curious about this on the description?
"I expect this to be further tweaked as we write the Shapeless records"

Where are we using shapeless?

tpolecat:
> Where are we using shapeless?

We use it to save some boilerplate in generating enum classes from the database tables. It's just machinery.

swalker2m:
It looks great to me so far.

modules/sql/src/main/scala/enum/GnirsEnums.scala
```diff
@@ -0,0 +1,29 @@
...
+    List(
+
+      EnumDef.fromQuery("GnirsCrossDispersed", "GNRIS Cross Dispersed") {
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'pixelScale -> BigDecimal`.T
+        sql"""SELECT id, id tag, short_name, long_name, pixel_scale FROM e_gnirs_cross_dispersed""".query[(String, E)]
+      },
+
+      EnumDef.fromQuery("GnirsFpuSlit", "GNRIS FPU Slit") {
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'slitWidth -> Arcseconds`.T
+        sql"""SELECT id, id tag, short_name, long_name, pixel_scale FROM e_gnirs_cross_dispersed""".query[(String, E)]
```
swalker2m: Oops, this is reading from the wrong table.  Should be `e_gnirs_fpu_slit` or whatever it is called.

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
```diff
@@ -0,0 +1,60 @@
...
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val count: Int,
+  val defaultWavelength: gem.math.Wavelength,
+  val minWavelength: gem.math.Wavelength,
+  val maxWavelength: gem.math.Wavelength,
+  val deltaWavelength: gem.math.Wavelength,
+  val band: Option[gem.enum.MagnitudeBand],
+  val cross_dispersed: Boolean
```
swalker2m: For other `Boolean` cases, we've made an enumerated type unique to that case. For example, see `MosPreimaging`. I guess that would be ideal, but certainly a hassle.  It's probably fine the way it is.

jdnavarro:
I still have to push add more the parameters before the final review. The ones left will be much more straightforward than DisperserOrder. I'll let you know when everything is ready ready.

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
jdnavarro: Yes, that cross_dispersed boolean everywhere bothers me too. I'll rework it in another PR where I'll try to include the cross-dispersion constrains.

modules/core/shared/src/main/scala/gem/enum/GnirsCamera.scala
```diff
@@ -0,0 +1,48 @@
...
+package gem
+package enum
+
+import cats.syntax.eq._
+import cats.instances.string._
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for GNRIS Camera.
+ * @group Enumerations (Generated)
```
cquiroz: This is unrelated to this PR but in ocs3 i have seen errors in scaladoc with these custom annotations. or are we using a custom set?

tpolecat: This is built into scaladoc but you have to pass a flag to turn it on. See https://github.com/gemini-hlsw/gem/blob/master/build.sbt#L162

cquiroz: Is there a list of those. In ocs3 we fail with `@typeclass`

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
tpolecat: Where does this value enter into our calculations? ITC? AGS? I don't know how big a deal this is.

modules/core/shared/src/main/scala/gem/enum/GnirsCamera.scala
tpolecat: http://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html

tpolecat:
This looks good so far. 👍

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
jdnavarro: This was a late addition, initially we only had the band.

In addition to ITC, it's also used indirectly in [EdCompInstGNIRS.java](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gnirs/EdCompInstGNIRS.java#L403-L404). I guess it has to do something with some label in UI, perhaps?

tpolecat: Hm. Not really knowing how this is used I guess I'm ok with it staying as a boolean for now. If we end up in a position where we're having to check it a lot and are in danger of forgetting to check (this is the issue I think) then we might want to have distinct types.

jdnavarro: The thing with cross dispersed mode (XD) is that several instrument parameters are only available when XD is on/off. In this case, some orders are only available in XD mode other orders only when no XD.

For these orders we could split the table in 2: `disperser_order_xd` and `disperser_order_mirror`. But my intention was to rework the whole XD feature everywhere after getting more expertise with GNIRS, because we have more XD constrains with other parameters and I'm not sure what would be the best way to encode them overall.

I was trying to play this PR conservatively and stick with to how it was written in OCS, before trying something more experimental that deviates more from the original code.

I could file ticket for this XD in order to not forget, or we could put this PR on hold and try to work this XD out before merging.

tpolecat: > It seems GNIRS behaves as an almost entirely different instrument when the XD is on.

Ok, it sounds like this needs some further design work before we commit to a config implementation. I'm fine merging this as-is, with the expectation that we will revisit it.

andrewwstephens: If the software is to model the instrument we may want to consider replacing the XD "mode" with the actual mechanism, a prism turret with 4 positions:
1. A mirror which is used for long-slit spectroscopy (no cross-dispersion)
2. The SXD prism which cross-disperses orders by ~8"
3. The LXD prism which cross-disperses orders by ~6"
4. The Wollaston prism

modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object GnirsDecker {
+
+  /** @group Constructors */ case object Acquisition extends GnirsDecker("Acquisition", "ACQ", "Acquisition", false)
+  /** @group Constructors */ case object PupilViewer extends GnirsDecker("PupilViewer", "PV", "Pupil viewer", false)
+  /** @group Constructors */ case object ShortCamLongSlit extends GnirsDecker("ShortCamLongSlit", "SHORT_CAM_LONG_SLIT", "Short camera long slit", false)
+  /** @group Constructors */ case object ShortCamCrossDispersed extends GnirsDecker("ShortCamCrossDispersed", "SHORT_CAM_XD", "Short camera cross dispersed", false)
+  /** @group Constructors */ case object Ifu extends GnirsDecker("Ifu", "IFU", "IFU", true)
+  /** @group Constructors */ case object LongCamLongSlit extends GnirsDecker("LongCamLongSlit", "LONG_CAM_LONG_SLIT", "Long camera long slit", false)
+  /** @group Constructors */ case object Wollaston extends GnirsDecker("Wollaston", "WOLLASTON", "Wollaston", true)
```
andrewwstephens: Why is the Wollaston in the GnirsDecker?  (it's really in the prism turret)

jdnavarro: It was there in the original OCS code, but it felt odd to me too. I'll remove it.

modules/core/shared/src/main/scala/gem/enum/GnirsFpuOther.scala
```diff
@@ -0,0 +1,50 @@
...
+sealed abstract class GnirsFpuOther(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val obsolete: Boolean
+) extends Product with Serializable
+
+object GnirsFpuOther {
+
+  /** @group Constructors */ case object Ifu extends GnirsFpuOther("Ifu", "IFU", "IFU", true)
```
andrewwstephens: Technically speaking, the GNIRS decker slide supports two IFUs (1 and 2).  Currently neither are installed although this may change in the future.

jdnavarro: We also need to take into account that at some point we'll need to import previous observations, so if we already have observations with IFU enabled, I think it's worth including it even if it's no longer installed.

modules/core/shared/src/main/scala/gem/enum/GnirsReadMode.scala
```diff
@@ -0,0 +1,51 @@
...
+) extends Product with Serializable
+
+object GnirsReadMode {
+
+  /** @group Constructors */ case object Bright extends GnirsReadMode("Bright", "Bright", "Bright Objects", 600, 30, 1)
+  /** @group Constructors */ case object Faint extends GnirsReadMode("Faint", "Faint", "Faint Objects", 9000, 10, 16)
+  /** @group Constructors */ case object VeryFaint extends GnirsReadMode("VeryFaint", "VeryFaint", "Very Faint Objects", 18000, 7, 32)
+  /** @group Constructors */ case object VeryBright extends GnirsReadMode("VeryBright", "VeryBright", "Very Bright/Acq./High Bkgrd.", 200, 155, 1)
+
+  /** All members of GnirsReadMode, in canonical order. */
```
andrewwstephens: Could we put these in *logical* order instead? (e.g. Very Bright, Bright, Faint, VeryFaint)

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
jdnavarro: I'll add Mirror and the Wollaston prism to a new table called `prism_turret`. I could use that table as the main source of constrains for the rest of parameters on other tables.

modules/core/shared/src/main/scala/gem/enum/GnirsReadMode.scala
tpolecat: @jdnavarro you may need to add an artificial sort key to the database table in order to select these in the required order.

modules/sql/src/main/resources/db/migration/V035__Add_Gnirs.sql
```diff
@@ -0,0 +1,339 @@
...
+
+--
+-- Data for Name: e_gnirs_cross_dispersed; Type: TABLE DATA; Schema: public; Owner: postgres
+--
+
+COPY e_gnirs_cross_dispersed (id, short_name, long_name, pixel_scale) FROM stdin;
+Sxd	SXD	SXD	0.15
+Lxd	LXD	LXD	0.05
+\.
+
```
andrewwstephens: The pixel scale is set entirely by the camera, so it may not be necessary here.

modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
andrewwstephens: Sorry, I understand now.  The decker has a position for the Wollaston prism, and this must set that position.

modules/core/shared/src/main/scala/gem/enum/GnirsCrossDispersed.scala
```diff
@@ -0,0 +1,47 @@
...
+  val shortName: String,
+  val longName: String,
+  val pixelScale: BigDecimal
+) extends Product with Serializable
+
+object GnirsCrossDispersed {
+
+  /** @group Constructors */ case object Sxd extends GnirsCrossDispersed("Sxd", "SXD", "SXD", 0.15)
+  /** @group Constructors */ case object Lxd extends GnirsCrossDispersed("Lxd", "LXD", "LXD", 0.05)
+
```
andrewwstephens: I realize that I may be commenting on old (and fully functional) code, however, it is surprising to specify the pixel scale here as a parameter of the cross-disperser.  The pixel scale is uniquely determined by the camera.  In addition, the SXD prism may be used with either camera, and therefore have either pixel scale.

modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
andrewwstephens: Here are the Decker positions from the GNIRS control software:
spectacle w26012

jdnavarro:
This is ready for another review.

The constraints for the XD mode are still missing but it's not that straightforward to me. It may be easier once we have some UIs in place and we know for sure how the instrument is going to be set in practice.

modules/core/shared/src/main/scala/gem/enum/GnirsDecker.scala
```diff
@@ -0,0 +1,52 @@
...
+  val shortName: String,
+  val longName: String,
+  val obsolete: Boolean
+) extends Product with Serializable
+
+object GnirsDecker {
+
+  /** @group Constructors */ case object Acquisition extends GnirsDecker("Acquisition", "ACQ", "Acquisition", false)
+  /** @group Constructors */ case object PupilViewer extends GnirsDecker("PupilViewer", "PV", "Pupil viewer", false)
+  /** @group Constructors */ case object ShortCamLongSlit extends GnirsDecker("ShortCamLongSlit", "SHORT_CAM_LONG_SLIT", "Short camera long slit", false)
```
tpolecat: The short name is intended to be seen by users so `SHORT_CAM_LONG_SLIT` is probably not what we want here. If you want to include an internal name I think it should be another field.

modules/core/shared/src/main/scala/gem/enum/GnirsDisperserOrder.scala
```diff
@@ -0,0 +1,61 @@
...
+  /** @group Constructors */ case object Two extends GnirsDisperserOrder("Two", "2", "Two", 2, Wavelength.unsafeFromAngstroms(34000), Wavelength.unsafeFromAngstroms(27000), Wavelength.unsafeFromAngstroms(43000), Wavelength.unsafeFromAngstroms(0), Some(MagnitudeBand.L), false)
+  /** @group Constructors */ case object Three extends GnirsDisperserOrder("Three", "3", "Three", 3, Wavelength.unsafeFromAngstroms(22200), Wavelength.unsafeFromAngstroms(18600), Wavelength.unsafeFromAngstroms(27000), Wavelength.unsafeFromAngstroms(6), Some(MagnitudeBand.K), true)
+  /** @group Constructors */ case object FourXD extends GnirsDisperserOrder("FourXD", "4", "FourXD", 4, Wavelength.unsafeFromAngstroms(16500), Wavelength.unsafeFromAngstroms(14200), Wavelength.unsafeFromAngstroms(18600), Wavelength.unsafeFromAngstroms(5), Some(MagnitudeBand.H), true)
+  /** @group Constructors */ case object Four extends GnirsDisperserOrder("Four", "4", "Four", 4, Wavelength.unsafeFromAngstroms(16300), Wavelength.unsafeFromAngstroms(14200), Wavelength.unsafeFromAngstroms(18600), Wavelength.unsafeFromAngstroms(5), Some(MagnitudeBand.H), true)
```
tpolecat: Here we have two elements with the same short name, which will be confusing.

modules/core/shared/src/main/scala/gem/enum/GnirsWaveLengthSuggestion.scala
```diff
@@ -0,0 +1,51 @@
...
+sealed abstract class GnirsWaveLengthSuggestion(
+  val tag: String,
+  val shortName: String,
+  val longName: String
+) extends Product with Serializable
+
+object GnirsWaveLengthSuggestion {
+
+  /** @group Constructors */ case object One extends GnirsWaveLengthSuggestion("One", "4.85", "4.85")
+  /** @group Constructors */ case object Two extends GnirsWaveLengthSuggestion("Two", "3.40", "3.40")
```
tpolecat: If we ever intend to use these values as numbers we will also want numeric fields for them.

jdnavarro:
I changed now several short names. I was previously sticking mostly to the names used in the OT, but it seems the UI labels weren't too much of a reference before anyway.

PR #164, opened 2017-12-01
Generate magnitude band enumerations

jdnavarro:
This kind of works but I'm getting the feeling I'm doing something wrong.


modules/sql/src/main/scala/MagnitudeSystem.scala
```diff
@@ -0,0 +1,13 @@
...
+package gem.sql
+
+/** Minimal MagnitudeSystem classes to support reading from the enum
+  * tables.
+  */
+
+final case class MagnitudeSystem(override val toString: String)
+object MagnitudeSystem {
+  def fromString(s: String): MagnitudeSystem = MagnitudeSystem(s)
+}
```
jdnavarro: This doesn't feel right.

modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -54,6 +56,7 @@ object EnumDef {
 
     implicit val caseWavelengthNm = at[Wavelength.Nm ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
     implicit val caseWavelengthUm = at[Wavelength.Um ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseMagnitudeSystem = at[MagnitudeSystem](a => s"gem.sql.MagnitudeSystem.fromString(${a.toString})")
```
tpolecat: Here and above you need `gem.enum.MagnitudeSystem` because this is the generated code that goes in core.

modules/sql/src/main/scala/MagnitudeSystem.scala
tpolecat: You don't need the factory method here. I would just say `MagnitudeSystem(id: String)` and call it good. The only purpose is to tag it as something other than a `String`.

modules/core/shared/src/main/scala/gem/enum/MagnitudeBand.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Enumerated type for magnitude band.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class MagnitudeBand(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val center: gem.math.Wavelength,
+  val width: gem.math.Wavelength,
```
swalker2m: I wonder if it makes sense to say that the width is a wavelength.

jdnavarro: You are right, it looks like wavelength but it isn't.

modules/core/shared/src/main/scala/gem/enum/MagnitudeBand.scala
```diff
@@ -0,0 +1,67 @@
...
+  /** @group Constructors */ case object H extends MagnitudeBand("H", "H", "H", gem.math.Wavelength.unsafeFromAngstroms(16500), gem.math.Wavelength.unsafeFromAngstroms(3000), gem.enum.MagnitudeSystem.unsafeFromTag("Vega"))
+  /** @group Constructors */ case object K extends MagnitudeBand("K", "K", "K", gem.math.Wavelength.unsafeFromAngstroms(22000), gem.math.Wavelength.unsafeFromAngstroms(4100), gem.enum.MagnitudeSystem.unsafeFromTag("Vega"))
+  /** @group Constructors */ case object L extends MagnitudeBand("L", "L", "L", gem.math.Wavelength.unsafeFromAngstroms(37600), gem.math.Wavelength.unsafeFromAngstroms(7000), gem.enum.MagnitudeSystem.unsafeFromTag("Vega"))
+  /** @group Constructors */ case object M extends MagnitudeBand("M", "M", "M", gem.math.Wavelength.unsafeFromAngstroms(47700), gem.math.Wavelength.unsafeFromAngstroms(2400), gem.enum.MagnitudeSystem.unsafeFromTag("Vega"))
+  /** @group Constructors */ case object N extends MagnitudeBand("N", "N", "N", gem.math.Wavelength.unsafeFromAngstroms(104700), gem.math.Wavelength.unsafeFromAngstroms(52300), gem.enum.MagnitudeSystem.unsafeFromTag("Vega"))
```
tpolecat: Can you use references for enum values; i.e., instead of `gem.enum.MagnitudeSystem.unsafeFromTag("Vega")` generate `gem.enum.MagnitudeSystem.Vega`?

jdnavarro: I was trying something similar but didn't manage to get it working. I'll fiddle a bit more.

modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -54,6 +56,7 @@ object EnumDef {
 
     implicit val caseWavelengthNm = at[Wavelength.Nm ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
     implicit val caseWavelengthUm = at[Wavelength.Um ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseMagnitudeSystem = at[MagnitudeSystem](a => s"""gem.enum.MagnitudeSystem.unsafeFromTag("${a.id}")""")
```
tpolecat: I think just change this line to say `gem.enum.MagnitudeSystem.${a.id}`)

modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -49,11 +54,14 @@ object EnumDef {
...
-    implicit val caseDegrees      = at[Degrees](a => s"gem.math.Angle.fromDoubleDegrees(${a.toDegrees})")
+    implicit val caseDegrees      = at[Degrees ](a => s"gem.math.Angle.fromDoubleDegrees(${a.toDegrees})")
     implicit val caseZoneId       = at[ZoneId  ](a => s"""java.time.ZoneId.of("${a.toString}")""")
 
-    implicit val caseWavelengthNm = at[Wavelength.Nm ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
-    implicit val caseWavelengthUm = at[Wavelength.Um ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseWavelengthNm = at[Wavelength.Nm](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseWavelengthUm = at[Wavelength.Um](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseSlitWidthNm  = at[SlitWidth.Nm ](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
+    implicit val caseSlitCenterNm = at[SlitCenter.Nm](a => s"gem.math.Wavelength.unsafeFromAngstroms(${a.toAngstrom})")
```
jdnavarro: Do you think it's worth having a separate Slit dimensions class in `math`? I'm thinking SlitWidth and SlitCenter are just placeholders that eventually need to be replaced by a more general solution for units with alternative representations.

modules/sql/src/main/scala/Slit.scala
```diff
@@ -0,0 +1,63 @@
...
+
+/** Minimal Slit classes to support reading from the enum tables.
+  *
+  * NOTE: For now this is a placeholder for slit dimensions replicating
+  * Wavelength. Scientifically it seems the fundamental unit should be
+  * arcseconds from which other representations could be derived. How to
+  * implement this properly should be clearer after we identify the commonality
+  * with other instruments or we come with a more general solution about
+  * handling units in general in gem.
+  */
```
jdnavarro: Perhaps it's not worth creating this placeholder now. Maybe it's better to put GNIRS implementation on hold and try to come up with a more permanent solution for the units problem now.

modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -29,15 +29,20 @@ object EnumDef {
     implicit def caseDegrees [S <: Symbol] = at[(S, Degrees) ] { case (s, _) => "  val " + s.name + ": gem.math.Angle"}
     implicit def caseZoneId  [S <: Symbol] = at[(S, ZoneId)  ] { case (s, _) => "  val " + s.name + ": java.time.ZoneId"}
 
-    implicit def caseWavelengthNm[S <: Symbol] = at[(S, Wavelength.Nm ) ] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
-    implicit def caseWavelengthUm[S <: Symbol] = at[(S, Wavelength.Um ) ] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
+    implicit def caseWavelengthNm[S <: Symbol] = at[(S, Wavelength.Nm)] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
+    implicit def caseWavelengthUm[S <: Symbol] = at[(S, Wavelength.Um)] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
+
+    implicit def caseSlitWidthNm [S <: Symbol] = at[(S, SlitWidth.Nm) ] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
+    implicit def caseSlitCenterNm[S <: Symbol] = at[(S, SlitCenter.Nm)] { case (s, _) => "  val " + s.name + ": gem.math.Wavelength"}
```
swalker2m: It doesn't make much sense to talk about slit widths in terms of a wavelength. They are measurement of length though, just like a wavelength.  What we've been doing is punting in cases like these and just storing unit-less primitives for now.

modules/sql/src/main/scala/EnumDef.scala
swalker2m: I think if we were using Libra slit widths would just be nm.  `Wavelength` doesn't really make sense here but it's probably not worth having a `SlitWidth` type?  Until then I wouldn't look askance at just using an `Int` and "knowing" it is nm... We do that type of thing in other places.

modules/sql/src/main/scala/EnumDef.scala
jdnavarro: I can see wavelength doesn't make sense for slit but here Wavelength was intended as a placeholder since it was already quite similar, but I'll use `Int` instead as you mentioned in another comment.

modules/sql/src/main/scala/EnumDef.scala
cquiroz: i thought we had discarded no use Libra, is that being reconsidered?

swalker2m: I didn't think it was discarded, but rather under-evaluated.  I had some concerns about it but I also have some concerns about the status quo.

cquiroz: By status quo do you mean using `Int` or `squants`?

tpolecat: My concern is that it's new and adopting it would be a "deep" dependency that would infect the lowest parts of the model (i.e., removing it later would be hard). I'm generally positive, but we will probably need to make improvements to it first. In any case I think Someone™ needs to take some time and evaluate it properly.

swalker2m: I mean using primitives with no unit information as we do in a few places.  The problem with Squants, I think, is that it stores values as a `Double`:

```scala
final class Length private (val value: Double, val unit: LengthUnit)
```

cquiroz: Right, forgot about that. I was meaning to progress on generalizing that for squants 2.0. Mabye @arturog8m wants to support that work

PR #166, opened 2017-12-07
Replace Floating types by arbitrary precision decimals

### LINKED ISSUE #162: Replace SQL double types with numeric
We don't really need floating point arithmetic (and all the gotchas that come with it) so most numeric types in gem should be able to be represented by types backed by normal or arbitrary precision integers.

- [ ] Postgres: `double` -> `numeric(m,n)`.
- [ ] Enumerations: `Double` -> `BigDecimal`. At some point we should be using proper unit types with `squants` or `libra` but this is still under discussion.

There are some camelCase renaming that could be done in the same migration.

jdnavarro:
I think I got all the floating types but I may have missed some. Do you any way to make sure there are no `double`s or `real`s in an entire postgres database?

Fixes #162.

### DIFF
--- a/modules/sql/src/main/resources/db/migration/V034__Replace_Floating.sql
+++ b/modules/sql/src/main/resources/db/migration/V034__Replace_Floating.sql
@@ -1,0 +1,17 @@
+--
+-- Replace double and real by numeric(m,n)
+--
+
+ALTER TABLE e_site
+      ALTER COLUMN longitude TYPE numeric(10,7),
+      ALTER COLUMN latitude  TYPE numeric(10,7);
+
+-- Some precision is lost during conversion
+UPDATE e_site SET longitude = -155.469055, latitude =  19.8238068 WHERE id = 'GN';
+UPDATE e_site SET longitude = -70.7366867, latitude = -30.2407494 WHERE id = 'GS';
+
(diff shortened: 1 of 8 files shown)


modules/sql/src/main/resources/db/migration/V034__Replace_Floating.sql
```diff
@@ -0,0 +1,17 @@
...
+      ALTER COLUMN longitude TYPE numeric(10,7),
+      ALTER COLUMN latitude  TYPE numeric(10,7);
+
+-- Some precision is lost during conversion
+UPDATE e_site SET longitude = -155.469055, latitude =  19.8238068 WHERE id = 'GN';
+UPDATE e_site SET longitude = -70.7366867, latitude = -30.2407494 WHERE id = 'GS';
+
+ALTER TABLE e_f2_lyot_wheel
+      ALTER COLUMN plate_scale TYPE numeric(4,3),
+      ALTER COLUMN pixel_scale TYPE numeric(3,2);
```
jdnavarro: I'm assuming the original numbers had some floating point precision weirdness in them, but if they are really those numbers let me know. https://github.com/jdnavarro/gem/blob/double/modules/sql/src/main/resources/db/migration/V001__Initial_Setup.sql#L789-L794

PR #167, opened 2017-12-11
update doobie, flyway, fs2, http4s, kp

tpolecat:


tpolecat:
Doh. Guess I should have tried compiling first. Standby.

modules/core/jvm/src/main/scala/gem/syntax/Stream.scala
```diff
@@ -49,7 +49,7 @@ final class StreamOps[F[_], O](val self: Stream[F, O]) {
             val aʹ = f(a, o)
             if (p(aʹ)) (matching :+ o, z) else (matching, aʹ)
           }).flatMap { case (matching, aʹ) =>
-            Pull.output(Chunk.vector(matching)) *> go(aʹ, tl)
+            Pull.outputChunk(Chunk.vector(matching)) >> go(aʹ, tl)
```
tpolecat: There are some subtle strictness issues here that caused them to replace `*>` with `>>` … not worth worrying about but there is an explanation.

modules/core/shared/src/main/scala/gem/math/EphemerisCoordinates.scala
```diff
@@ -37,6 +37,9 @@ final case class EphemerisCoordinates(
 
 object EphemerisCoordinates {
 
+  val Zero: EphemerisCoordinates =
+    EphemerisCoordinates(Coordinates.Zero, Offset.Zero)
+
```
tpolecat: Not sure how the test that uses this value compiled before. I think there was some unnoticed LUBbing happening.

modules/db/src/test/scala/gem/dao/EphemerisDaoSpec.scala
```diff
@@ -183,7 +183,7 @@ class EphemerisDaoSpec extends PropSpec with PropertyChecks with DaoTest {
       val p  = (mʹ.toList.traverse { case (ks, e) =>
         Stream.emits(e.toMap.toList).covary[ConnectionIO]
           .to(EphemerisDao.streamInsert(ks.key, ks.site))
-          .runSync
+          .run
```
tpolecat: The `Effect` instance requirement here has been lifted so `runSync` is no longer needed.

modules/ocs2/src/main/scala/gem/ocs2/ImportServer.scala
```diff
@@ -12,13 +12,12 @@ import gem.config.{ StaticConfig, DynamicConfig }
 import gem.ocs2.pio.{PioDecoder, PioError}
 import gem.ocs2.pio.PioError._
 
-import fs2.Stream
+import fs2.{ Stream, StreamApp }
```
tpolecat: `StreamApp` is now provided by fs2 so the one in http4s is deprecated.

tpolecat:
This should be the last milestone for fs2 0.10 … they're planning to release with cats 1.0 which is due in the next few days.

PR #168, opened 2017-12-11
Target Environment Discussion

swalker2m:
This is a content-free PR (:-1:) that I wanted to offer now before going any further on the target environment.  Target environment is rather involved and I think I've lived with it for so long that it's hard for me to see simplifications.  What follows are initial ideas for how to represent the target environment in a relational database.  I'm not terribly proud of any of this though and it remains terribly complicated so I'd appreciate any comments or help that you have to offer.

### DIFF
--- a/TargetEnvironmentDataModel.md
+++ b/TargetEnvironmentDataModel.md
@@ -1,0 +1,194 @@
+
+# Target Environment Notes
+
+The target environment describes the collection of targets associated with an observation.  This document explores ideas for how we might represent this collection in a relational database. The target environment consists of three distinct parts:
+
+1. _Asterism_. The science targets, or "asterism".  Before GHOST, this was always a single target. With the arrival of GHOST, we will need instrument-specific asterisms where particular targets are associated with various instrument features like the state of guide fibers and fiber "agitators".
+
+2. _Guide Stars_. Selected guide stars, the probes with which they are associated and alternatives in case the chosen stars are not adequate for some reason. 
+
+3. _User Targets_. So-called "user" targets for blind offsets, tuning stars, etc.
+
+A goal is to keep the data model as flat as possible, so my initial thought is that there is no need for a `target_environment` table to tie these things together.  Instead, we have `observation` of course and then tables for the specific parts of the environment.
+
+
+## Asterism
+
+Asterisms for new instruments like GHOST incorporate instrument-specific features.  It seems clear that we'll need a structure somewhat like the one we have for static instrument configuration wherein there is a "base" table with an asterism id and asterism type along with type-specific tables holding the details.  Currently we have four types of asterism:
+
+* Single Target
+* GHOST Dual Target
+* GHOST Beam Switching 
+* GHOST High Resolution
+
+Clearly asterisms can be relevant only to particular instruments.  It would be nice to ensure that the observation's instrument corresponds to the asterism's instrument so we cannot, for instance, use a GHOST Dual Target asterism with a GMOS observation.
+
+First a discriminator:
+
+```sql
+CREATE TYPE asterism_type AS ENUM (
+  'SingleTarget',
+  'GhostDualTarget'
+--  'GhostBeamSwitching',
+--  'GhostHighRes'
+);
+```
+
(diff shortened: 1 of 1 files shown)


tpolecat:
This looks good at first glance. I'll read it in more detail in a bit.

TargetEnvironmentDataModel.md
```diff
@@ -0,0 +1,194 @@
...
+  target_id_1   integer       NOT NULL  REFERENCES target(target_id),
+  target_id_2   integer       NOT NULL  REFERENCES target(target_id),
+  -- more configuration here for guide fibers, etc.
+  PRIMARY KEY (asterism_id, asterism_type, instrument) REFERENCES asterism ON DELETE CASCADE,
+  CONSTRAINT is_ghost_dual (asterism_type = 'GhostDualTarget'),
+  CONSTRAINT is_ghost      (instrument    = 'GHOST')
+);
+```
+
+An observation will be deleted if its asterism is deleted, but we'll need a trigger to delete the asterism when the corresponding observation is deleted.
```
tpolecat: I think we should not cascade here because the observation is logically the owner; deleting the asterism out from under an observation seems like a logic error. The trigger on the other end makes sense.

TargetEnvironmentDataModel.md
```diff
@@ -0,0 +1,194 @@
...
+CREATE TABLE guide_star (
+  guide_star_id   SERIAL      PRIMARY_KEY,
+  guide_probe     guide_probe NOT NULL,
+  guide_group_id  integer     NOT NULL    REFERENCES guide_group ON DELETE CASCADE,
+  target_id       integer     NOT NULL    REFERENCES target      ON DELETE CASCADE,
+  UNIQUE (guide_group_id, guide_probe) -- No group should have more than one guide star assigned to a particular probe
+);
+```
+
+Nothing here prevents us from adding guide probes that don't correspond to the observation's instrument (e.g., a GMOS OI guide star for a GHOST observation).  I don't see a good way to lock that down though.
```
tpolecat: Could we make the guide probe a lookup table with the instrument as a column, then add it to `guide_star` and `guide_group` and then a composite FK to `observation`? I assume you went through this and there's a reason it won't work.

tpolecat:
I like this design. Just saying ok the asterims, guiding, and user targets are independent things makes the model much simpler. Doing away with ordering as well. So 👍

TargetEnvironmentDataModel.md
swalker2m: Some guide probes, like PWFS2, have no associated instrument.  In that case the `instrument` column is null in the `e_guider` table holding the guide probes and the foreign key constraint referencing it from `guide_star` cannot be satisfied.  Or maybe there's a better way?

swalker2m: I guess one possibility is to have two instrument columns, one for the guide probe instrument and one for the observation instrument.  So a FK into `e_guider` with its instrument and a FK into `observation` with its instrument and a `CHECK` constraint that either the two instruments are the same or else the `e_guider` instrument `IS NULL`.

```sql
CREATE TABLE guide_star (
  id                SERIAL      PRIMARY KEY,
  group_id          integer     NOT NULL    REFERENCES guide_group(id) ON DELETE CASCADE,
  target_id         integer     NOT NULL    REFERENCES target(id),

  guider            identifier  NOT NULL    REFERENCES e_guider(id),
  guider_instrument identifier              REFERENCES e_instrument,
  FOREIGN KEY (guider, guider_instrument)   REFERENCES e_guider(id, instrument),

  program_id        text        NOT NULL,
  observation_index id_index    NOT NULL,
  instrument        identifier  NOT NULL    REFERENCES e_instrument,
  FOREIGN KEY (program_id, observation_index, instrument) REFERENCES observation ON DELETE CASCADE,

  UNIQUE (group_id, guider), -- No group should have more than one guide star assigned to a particular probe
  CHECK ((guider_instrument IS NULL) OR (instrument = guider_instrument))
);
```

swalker2m:
Removing up this old discussion PR.

PR #171, opened 2017-12-15
gemctl re-do

tpolecat:
This is a re-do of the deployment stuff and it's quite a bit simpler, but it will be hard to review because there are so many changes. It probably makes sense is to just squash-merge it and then go through the whole thing together next week, since nothing here effects anything in the application itself. I'll add some inline comments about build changes.


build.sbt
```diff
@@ -20,40 +22,32 @@ lazy val shapelessVersion    = "2.3.2"
...
 
-git.uncommittedSignifier in ThisBuild := Some("UNCOMMITTED")
+version in ThisBuild := imageManifest.formatVersion
 
 // check for library updates whenever the project is [re]load
-onLoad in Global := { s => "dependencyUpdates" :: s }
+onLoad in Global := { s => 
+  if (sys.props.contains("gem.skipDependencyUpdates")) s
+  else "dependencyUpdates" :: s 
+}
```
tpolecat: In Jenkins I'm passing `-Dgem.skipDependencyUpdates` to speed things up. We should do this in Travis as well I guess.

build.sbt
```diff
@@ -323,19 +320,32 @@ lazy val telnetd = project
...
+    // Install nc before changing the user
+    dockerCommands       ++= dockerCommands.value.flatMap {
+      case c @ Cmd("USER", args @ _*) =>
+        Seq(
+          ExecCmd("RUN", "apt-get", "update"),
+          ExecCmd("RUN", "apt-get", "--assume-yes", "install", "netcat-openbsd"),
+          c
+        )
+      case cmd => Seq(cmd)
+    }
```
tpolecat: The idea here is to use netcat to probe the telnet port to see if it's open, as a health check. But it leaves the telnet server in a weird state and I haven't been able to figure out why. So for now this is unused and will probably go away. When we add the web service to the app we can just use curl for the health check.

modules/ctl/src/main/scala/Parsers.scala
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.ctl
```
tpolecat: This is more or less the same commandline parsing code from before but it's better organized.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.0.3
+sbt.version=1.1.0-RC1
```
tpolecat: 1.1 adds LSP support, so you can get error highlighting and jump-to-class in VS-Code. Dale and Eugene say it's good to go so I don't think there's any harm in upgrading.

cquiroz: Should we wait for the final release. It should be quite soon
I want to add scalameta support to ocs3 along 1.1.0. WDYT?

build.sbt
cquiroz: Do we really want to have 2 CIs setup? Could we just use travis and let github check the state from these somehow?

cquiroz:
It seems the way to select versions is in this PR. i'd like that we review versioning to harmonize it with ocs3, especially now that we are close to a first release of the seqexec

swalker2m:
> It probably makes sense is to just squash-merge it and then go through the whole thing together next week, since nothing here effects anything in the application itself.

👍 (modulo travis ci failure)

build.sbt
tpolecat: We're not really doing CI on Jenkins … we're just packaging and deploying. We could move CI to Jenkins and have it poke GitHub but I haven't looked into this.

cquiroz: Ok, then don't worry. I checked this a while ago and github can ping a local jenkins but we need to open the firewall which is a hassle

project/build.properties
tpolecat: Sure, I can roll this back to 1.0.4 for now.

tpolecat:
> i'd like that we review versioning to harmonize it with ocs3

Sure, do you want to do that here? It might be nice to have a record of the discussion.

My thinking is as follows:
- The commit hash is the only trustworthy way to identify a software version. There's no human involved so you can't get it wrong and can't change it after the fact. 
- In gem we need to ensure that a new deployment is a descendent of the current deployment, which we can only do via commits.
- I don't think the Julian date prefix I'm using now is worth the trouble and want to come up with another strategy.
- A tag is just a human-readable label for a commit, so I think these ideas are compatible. So maybe we could pass the tag (if any, see below) as an argument to `main()` so the software can show its tag to the user.
  - For test deployments we could pass the most recent tag, an offset, and the latest short hash; like `1.2.3+2:4b602b4`.
  - For production deployments we could either require a tag or offer to create one prior to deployment, and deploy `+0` could be displayed simply as `1.2.3`.

What's your thinking?

PS - It probably makes sense to spin off `gemctl` into its own repo. At this point it doesn't depend on anything in gem.

tpolecat:
This needs another 👍 , sorry

cquiroz:
BTW, if you want to make sure that nobody can ever change a hash we should disable forced push on master

cquiroz:
I agree with most of your comments, the only think that worries me a bit is to present users with the hash, but maybe science folks are ok with that. Having a tag for deployments is probably good enough and for test we can probably do whatever we want

The other question is what type of versioning to use. We have historically put the yearsemester in the version. Do we want to keep that? or start on 1.0.x?

tpolecat:
My preference would be to use hashes for versions and not tie them to semesters at all; i.e., sometime before 2020A starts we would do whatever we need to support 2020A programs and it just works when programs start appearing. I want there to be literally zero work to be done at the semester boundary.

tpolecat:
(We do have force-push to master disabled).

swalker2m:
>  I want there to be literally zero work to be done at the semester boundary.

Aren't there often filter changes, etc. at semester boundaries?

tpolecat:
>>  I want there to be literally zero work to be done at the semester boundary.
> Aren't there often filter changes, etc. at semester boundaries?

Yes, but I think we need to model this. The filter set will depend on the program's semester, or the scheduling block for the obs, or something.

swalker2m:
> Yes, but I think we need to model this. The filter set will depend on the program's semester, or the scheduling block for the obs, or something.

Makes sense.

tpolecat:
Yes, I'm going to redo the doc. I just didn't want to keep sitting on these changes. With this we'll get continuous deployment to test and on-demand to prod and can start start working on incremental improvements.

PR #173, opened 2017-12-20
new main app

tpolecat:
This PR
- Adds a new `main` project which runs the telnet server and web server together. You can do `main/run` to bring it up locally. Press enter to shut it down.
- Leaves behind similar standalone runners for `web` and `telnetd`. Note that these are not configurable and don't run database migrations at startup. They're just for development.
- Adds distinct configuration data types for database, telnet, and web.
- Adds `Stream` constructors for services that need to be cleaned up, which simplifies lifecycle management. The entry points now use `StreamApp`.

This PR does **not**
- Add a configuration library like [PureConfig](https://github.com/pureconfig/pureconfig), but we will need to eventually.
- Add a homepage to the web app. That will be next.


cquiroz:
Just as general note we have been using `knobs` for configuration on ocs3

modules/main/src/main/scala/MainConfiguration.scala
```diff
@@ -0,0 +1,25 @@
...
+  web:      WebConfiguration,
+  telnetd:  TelnetdConfiguration
+)
+
+object MainConfiguration {
+
+  val forTesting: MainConfiguration =
+    apply(
+      DatabaseConfiguration.forTesting,
+      WebConfiguration.forTesting,
```
cquiroz: I wish we could coordinate with ocs3 on these details otherwise it will be harder to merge. We already have some configuration classes for Web and other subsystems

modules/web/src/main/scala/WebServer.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.http4s.server.blaze.BlazeBuilder
+
+object WebServer {
+
+  // Single-element stream that creates and yields a web server, guaranteeing cleanup.
+  private def server(cfg: WebConfiguration.WebServer, root: HttpService[IO]): Stream[IO, Server[IO]] =
+    Stream.bracket(
+      BlazeBuilder[IO]
+        .bindHttp(cfg.port, cfg.host)
+        .mountService(root, "/")
```
cquiroz: is `root` providing some sort of rest api or the UI?

modules/web/src/main/scala/WebServer.scala
```diff
@@ -0,0 +1,33 @@
...
+        .start)(
+      s => Stream(s),
+      _.shutdown
+    )
+
+  /** A single-element stream that creates and yields a web server, guaranteeing cleanup. */
+  def stream(web: WebConfiguration, db: DatabaseConfiguration): Stream[IO, Server[IO]] =
+    Environment.stream(web, db).flatMap { env =>
+      server(env.config.webServer, Gatekeeper(env)(Application.service))
```
cquiroz: What's `Gatekeeper`?

cquiroz:
I haven't followed in much detail the progress in `gem` but it worries me that we are duplicating stuff already solved in ocs3 like configuration, the launcher, etc

I hope we could converge onto some commonality to make the future merge simpler

modules/web/src/main/scala/WebServer.scala
tpolecat: It's the thing that authenticates requests and does login processing.

tpolecat:
> Just as general note we have been using knobs for configuration on ocs3

Yeah there are a few reasonable-looking options. I don't feel very strongly, although knobs won't work because it depends on scalaz. PureConfig and Case Classy both have cats instances.

> I haven't followed in much detail the progress in gem but it worries me that we are duplicating stuff already solved in ocs3 like configuration, the launcher, etc

There will certainly be some de-duplication when we merge stuff. I don't have a good sense of the timeframe for doing this.

cquiroz:
I'd like to promote doing the merge sooner than later, as time passes the merge gets harder. We'd certainly benefit from having the gem model instead of the ocs2 model for the seqexec and would help to test gem as well. 

We should port the seqexec to cats/fs2 before that though. We have discussed to do that soon after we make the first release, but let's see.

tpolecat:
Oops, had to rebase. Will need another approval assuming CI remains green.

PR #174, opened 2017-12-20
update gem to expose ports properly

tpolecat:
This changes the internal and docker-exposed ports to be `9090` (http) and `9091` (telnet) so that's consistent now. Also had to publish the web service on `0.0.0.0` to make it externally visible. Once this is merged the services should be visible on `sbfocstest-lv1`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,9 +28,9 @@ def imageManifest = ImageManifest.current("postgres:9.6.0").unsafeRunSync
 version in ThisBuild := imageManifest.formatVersion
 
 // check for library updates whenever the project is [re]load
-onLoad in Global := { s => 
+onLoad in Global := { s =>
   if (sys.props.contains("gem.skipDependencyUpdates")) s
-  else "dependencyUpdates" :: s 
+  else "dependencyUpdates" :: s
 }
 
 cancelable in Global := true
@@ -367,7 +367,7 @@ lazy val main = project
   .settings(
     packageName in Docker := "gem",
     dockerBaseImage       := "openjdk:8u141",
-    dockerExposedPorts    := List(6666),
+    dockerExposedPorts    := List(9090, 9091),
     dockerRepository      := Some("sbfocsdev-lv1.cl.gemini.edu"),
     dockerLabels          := imageManifest.labels,
 
--- a/modules/ctl/src/main/scala/hi/Containers.scala
+++ b/modules/ctl/src/main/scala/hi/Containers.scala
@@ -46,7 +46,8 @@ object Containers {
                 "--label", s"gem.role=gem",
                 // "--health-cmd", if (r) "\"nc -z localhost 6666\""
                 //                 else     "nc -z localhost 6666",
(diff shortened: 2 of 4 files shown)


build.sbt
```diff
@@ -367,7 +367,7 @@ lazy val main = project
   .settings(
     packageName in Docker := "gem",
     dockerBaseImage       := "openjdk:8u141",
-    dockerExposedPorts    := List(6666),
+    dockerExposedPorts    := List(9090, 9091),
```
cquiroz: Do these also need to be opened in the host's firewall somehow?

tpolecat: I don't think the VMs have firewalls running since they're internal. But eventually yes.

cquiroz: right, can docker do it automagically? Does it run as root?

cquiroz:
Looks good 👍

build.sbt
tpolecat: Containers can run services on any port because they're sandboxed. These ports can be seen by other containers on the same machine if they join the same virtual network. This is how Gem can see Postgres (but nobody else can).

To see a container's port *externally* it must be mapped to a unique port on the host. We expose gem's ports 9090 and 9091 this way, using the same port numbers on the host for simplicity.

Docker itself runs as `docker` (not `root`) so I don't think you can map a container port to a host port < 1024 without special permissions. I don't think it has anything to say about a firewall running on the host.

tpolecat:
Ok so the ports are all set up and deployment to test is automatic.

```
~$ curl -i -d '{ "uid": "root", "pass": "" }' sbfocstest-lv1:9090/login
HTTP/1.1 200 OK
Content-Type: text/plain; charset=UTF-8
Set-Cookie: gem.jwt=<stuff>
Date: Thu, 21 Dec 2017 07:07:17 GMT
Content-Length: 10

Logged in.~$ 
~$ 
~$ curl -i -b gem.jwt=<stuff> sbfocstest-lv1:9090/api/query/program
HTTP/1.1 200 OK
Content-Type: application/json
Set-Cookie: gem.jwt=<new-stuff>
Date: Thu, 21 Dec 2017 07:09:28 GMT
Content-Length: 2

[]~$ 
~$ 
~$ telnet sbfocstest-lv1 9091
Trying 172.16.90.172...
Connected to gem-test.
Escape character is '^]'.
Welcome to Gem
Username: root
Password: 
Welcome, local time is Thu Dec 21 07:09:58 UTC 2017.
root@gem> exit
Goodbye.
Connection closed by foreign host.
~$ 
```

build.sbt
cquiroz: nice, container isolation cannot be bad while letting them talk to each other
BTW i always like to use the standard ports like 80 and 443 for the web part. What I've found the easiest is to get ITS to forward those ports to your app in the local firewall

tpolecat: I'm not sure docker can forward those ports but yeah you're right. Might as well if I can. I'll try it.

Issue #175, opened 2017-12-21
Use `text` instead of `character varying(n)`?


swalker2m:
The schema employs many `varchar(n)` types.  I'm not sure what we're gaining by specifying an explicit limit.  According to the postgres doc, where the "three types" are

1) char(n) - blank padded with explicit limit
2) varchar(n) - explicit limit
3) varchar / text - no limit

> There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs and slower sorting. In most situations text or character varying should be used instead.

Suggest changing all `character varying(n)` to `text`.

tpolecat:
This seems reasonable. 👍

PR #178, opened 2017-12-22
varchar(n) to text

swalker2m:
Issue #175, replacing `varchar(n)` with `text`.  Also switches to the `identifier` domain in a few places where it should have been used.


cquiroz:
This is probably a non-issue but would you loose data when converting. I suppose any `varchar(n)` will fit in `text`

PR #181, opened 2017-12-29
Gpi Enums

cquiroz:
This a first pass at defining GPI enums, I'm sending it for review as I'm a bit unsure on the details

Some notes:
* I copied some definitions from the GNIRS PR to use `MagnitudeBand`
* The observing mode is quite tricky as it refers to other enumerations
* `correspondingHMode` is pending as there are problems with self references


modules/sql/src/main/scala/enum/GpiEnums.scala
```diff
@@ -0,0 +1,93 @@
...
+        sql"""SELECT id, id tag, short_name, long_name, obsolete FROM e_gpi_sampling_mode""".query[(String, E)]
+      },
+
+      EnumDef.fromQuery("GpiCassegrain", "GPI Cassegrain") {
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'value -> Int`.T
+        sql"""SELECT id, id tag, short_name, long_name, value FROM e_gpi_cassegrain""".query[(String, E)]
+      },
+
+      EnumDef.fromQuery("GpiObservingMode", "GPI ObservingMode") {
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'filter -> Option[EnumRef['GpiFilter]], 'filterIterable -> Boolean, 'apodizer -> Option[EnumRef['GpiApodizer]], 'fpm -> Option[EnumRef['GpiFPM]], 'lyot -> Option[EnumRef['GpiLyot]], 'brightLimitPrism -> Option[Double], 'brightLimitWollaston -> Option[Double], 'obsolete  -> Boolean`.T
```
cquiroz: `correspondingHMode` is pending

modules/sql/src/main/scala/EnumRefs.scala
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.sql
+
+/** Minimal classes to refer to other enums by id. This is required to construct
+ *  enums refering to others like Gpi Observing Mode. */
+object EnumRefs {
+  final case class EnumRef[T <: Symbol](tag: String)
```
cquiroz: This let's me refer to another enumeration, passing the name of the enumeration as a literal type. I'd rather have the real type but they are not yet available at the time code is generated

modules/sql/src/main/resources/db/migration/V038__Add_Gpi.sql
```diff
@@ -0,0 +1,386 @@
...
+  short_name              text          NOT NULL,
+  long_name               text          NOT NULL,
+  filter                  identifier    REFERENCES e_gpi_filter,
+  filter_iterable         boolean       NOT NULL,
+  apodizer                identifier    REFERENCES e_gpi_apodizer,
+  fpm                     identifier    REFERENCES e_gpi_fpm,
+  lyot                    identifier    REFERENCES e_gpi_lyot,
+  bright_limit_prism      numeric(4,3)  NULL,
+  bright_limit_wollaston  numeric(4,3)  NULL,
+  corresponding_h_mode    identifier    REFERENCES e_gpi_observing_mode, -- To avoid self references, null its assumed to point to itsel	f
```
cquiroz: This is a bit tricky to handle. it can't be null at the end according to the old model

modules/sql/src/main/resources/db/migration/V038__Add_Gpi.sql
```diff
@@ -0,0 +1,386 @@
...
+
+--
+-- Name: e_gpi_pupil_camera; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE e_gpi_pupil_camera (
+  id          identifier  PRIMARY KEY,
+  short_name  text        NOT NULL,
+  long_name   text        NOT NULL,
+  value       boolean     NOT NULL
```
cquiroz: Maybe we don't need value?

modules/sql/src/main/resources/db/migration/V038__Add_Gpi.sql
tpolecat: I don't understand what the problem is here. A row can refer to itself.

swalker2m: I think he's referring to the issue when generating code -- if he supplies as the value the same thing as the enum id then the code generation fails.  The solution instead would be to go with the `LazyRef` trick.

tpolecat: Oh right, it's the generated code that's an issue. I'll pull your branch and see if I can fix it.

tpolecat:
Ok I opened a PR on @cquiroz's fork with `LazyEnumRef`. It seems to work. https://github.com/cquiroz/gem/pull/1

modules/sql/src/main/resources/db/migration/V038__Add_Gpi.sql
cquiroz: Yes, it was about the generated code as I hit the compiler limitation to refer to itself. The comment is bogus as it can’t be null on the mode

tpolecat:
Are there other things that you can make `NOT NULL` via `LazyEnumRef`? If not this looks good to me. 👍

cquiroz:
Only one reference to band that should be marked as `NOT NULL`

modules/sql/src/main/resources/db/migration/V038__Add_Gpi.sql
```diff
@@ -27,7 +27,7 @@ CREATE TABLE e_gpi_filter (
   id          identifier  PRIMARY KEY,
   short_name  text        NOT NULL,
   long_name   text        NOT NULL,
-  band        identifier  REFERENCES e_magnitude_band,
+  band        identifier  NOT NULL REFERENCES e_magnitude_band,
```
tpolecat: Do you need to change the code to use `EnumDef` here?

cquiroz: Good question. is `MagnitudeBand` an enum? It would seem so

cquiroz: Yes, I changed it, seems more consistent now

cquiroz:
Great to get my first relevant PR in 😄

PR #183, opened 2018-01-03
JulianDate

swalker2m:
The TCS export of ephemeris data requires the equivalent Julian Date for each ephemeris element.  I noticed that the code we have for this in `Epoch.Scheme` computes the Julian Day Number, but not the Julian Date itself (which includes the fractional day) so I decided to add a `JulianDate` class.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/JulianDate.scala
+++ b/modules/core/shared/src/main/scala/gem/math/JulianDate.scala
@@ -1,0 +1,87 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.math
+
+import cats.{Eq, Show}
+
+import java.time.{Instant, LocalDateTime}
+import java.time.ZoneOffset.UTC
+
+/** Astronomical time representation of continuous days since noon, November 24,
+  * 4714 BC.
+  *
+  * @param dayNumber      whole Julian Day number
+  * @param nanoAdjustment fraction of a day adjustment in nanoseconds
+  *
+  * @see The Wikipedia [[https://en.wikipedia.org/wiki/Julian_day article]]
+  */
+sealed abstract case class JulianDate(
+  dayNumber:      Int,
+  nanoAdjustment: Long
+) {
+
+  import JulianDate.{ NanoPerDay, MaxAdjustment, MinAdjustment }
+
+  // Guaranteed by the JulianDate constructors, double checked here.
+  assert(dayNumber      >= 0,             s"dayNumber >= 0")
+  assert(nanoAdjustment >= MinAdjustment, s"nanoAdjustment >= $MinAdjustment")
+  assert(nanoAdjustment <= MaxAdjustment, s"nanoAdjustment <= $MaxAdjustment")
+
+
+  /** Julian date value as a Double, including Julian Day Number and fractional
+    * day since the preceding noon.
+    */
+  val toDouble: Double =
+    dayNumber + nanoAdjustment.toDouble / NanoPerDay.toDouble
+}
+
+
+object JulianDate {
+
+  // One half day of seconds, ignoring leap seconds since Java time API ignores
+  // them as well.
+  private val SecPerDay: Int      = 86400
+  private val SecPerHalfDay: Int  = 43200
+
+  private val Billion: Int        = 1000000000
+  private val NanoPerDay: Long    = SecPerDay.toLong * Billion.toLong
+
+  private val MinAdjustment: Long = -SecPerHalfDay.toLong * Billion.toLong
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/gem/math/JulianDate.scala
```diff
@@ -0,0 +1,87 @@
...
+
+/** Astronomical time representation of continuous days since noon, November 24,
+  * 4714 BC.
+  *
+  * @param dayNumber      whole Julian Day number
+  * @param nanoAdjustment fraction of a day adjustment in nanoseconds
+  *
+  * @see The Wikipedia [[https://en.wikipedia.org/wiki/Julian_day article]]
+  */
+sealed abstract case class JulianDate(
```
cquiroz: Should you override equals/hashCode in this case?

modules/core/shared/src/test/scala/gem/math/JulianDateSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import cats.tests.CatsSuite
+import cats.{ Eq, Show }
+import gem.arb._
+
+import java.time.LocalDateTime
+
+
+@SuppressWarnings(Array("org.wartremover.warts.ToString", "org.wartremover.warts.Equals"))
+final class JulianDateSpec extends CatsSuite {
```
cquiroz: Maybe we could add a test for `Eq` using `discipline`

modules/core/shared/src/main/scala/gem/math/JulianDate.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  /** Convert an `Instant` to a Julian Date.
+    */
+  def ofInstant(i: Instant): JulianDate =
+    ofLocalDateTime(LocalDateTime.ofInstant(i, UTC))
+
+  /** JulianDate from a `LocalDateTime` assumed to represent a time at UTC.
+    */
+  def ofLocalDateTime(ldt: LocalDateTime): JulianDate = {
+    val y   = ldt.getYear
```
cquiroz: Where are these formulas? on the wikipedia page?

modules/core/shared/src/main/scala/gem/math/JulianDate.scala
tpolecat: No, you still get a generated `equals/hashCode` but no `copy` and no companion `apply`.

modules/core/shared/src/main/scala/gem/math/JulianDate.scala
```diff
@@ -0,0 +1,87 @@
...
+              d - 32075
+
+    // Whole seconds since midnight
+    val secs = ldt.getHour * 3600 + ldt.getMinute * 60 + ldt.getSecond
+    val adj  = (secs - SecPerHalfDay).toLong * Billion + ldt.getNano
+
+    new JulianDate(jdn, adj) {}
+  }
+
+  implicit val JulianDateEq: Eq[JulianDate] =
```
tpolecat: Any reason not to have `Order` here?

modules/core/shared/src/test/scala/gem/math/JulianDateSpec.scala
tpolecat: Agree, see [here](https://github.com/gemini-hlsw/gem/blob/master/modules/core/shared/src/test/scala/gem/math/AngleSpec.scala#L17) for an example; there's an equivalent for `Order` should we choose to upgrade.

modules/core/shared/src/main/scala/gem/math/JulianDate.scala
cquiroz: so `abstract case class` just forbids construction but the rest of the case class sematics apply?

modules/core/shared/src/main/scala/gem/math/JulianDate.scala
swalker2m: The computation for Julian Day is reproduced in the wikiwiki page and elsewhere.  They attribute it to an "explanatory supplement" of an astronomical almanac from 1992 where it is attributed to "Fliegel and Van Flandern (1968)":

screenshot 2018-01-04 09 42 07

I suppose to be correct it should check that the date is actually actually after -4713 November 23 and return the result wrapped in a corresponding `Option`.  Or maybe that is overkill.

The fractional part is just the fraction of a day corresponding to the time.

swalker2m: Fliegel, H.F. and Van Flandern, T.C. (1968). "A Machine Algorithm for Processing Cal-
endar Dates" Communications ol the Association of Computing Machines ll, 6sT.

swalker2m:
Alright, updated to `Order`.  I should have done that originally.  I just spaced it.

tpolecat:
Looks great! 👍

PR #184, opened 2018-01-05
WIP on Format abstraction

tpolecat:
This is an attempt to unify the format+parse pairs in a formal way, via `Format[A, B]` which is a pair `A => Option[B]` and `B => A`. It's like `Prism` but less strict in that parsing can be lenient (i.e., non-injective: many `A` might map to the same `Option[B]`).

In any case this has some benefits.
- We know which format+parse pairs go together without relying on naming conventions.
- We can test them with a standard set of laws. The setup here mirrors the setup in cats.
- Given `Format[A, B]` and `Meta[A]` we can get a `Meta[B]`, so these relationships can be re-used in the database layer, and similarly in the JSON layer.

Anyway it needs to be built out more but you get the idea. Wanted to get some feedback before continuing.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
+++ b/modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
@@ -5,7 +5,6 @@ package gem
 
 import cats.{ Eq, Show }
 import cats.kernel.laws.discipline._
-import monocle.law.discipline._
 import cats.tests.CatsSuite
 import gem.arb._
 import gem.enum.{ Site, DailyProgramType }
@@ -20,7 +19,6 @@ final class ProgramIdSpec extends CatsSuite {
 
   // Laws
   checkAll("Program.Id", OrderTests[Program.Id].order)
-  checkAll("Program.Id.Formats.Standard", PrismTests(ProgramId.Optics.fromString))
 
   test("Equality must be natural") {
     forAll { (a: ProgramId, b: ProgramId) =>
@@ -95,6 +93,12 @@ final class ProgramIdSpec extends CatsSuite {
   test("Nonstandard must reparse") {
     forAll { (nid: Nonstandard) =>
       Nonstandard.fromString(nid.format) shouldEqual Some(nid)
+    }
+  }
+
+  test("ProgramId must reparse") {
+    forAll { (pid: ProgramId) =>
+      ProgramId.fromString(pid.format) shouldEqual Some(pid)
     }
   }
 
--- a/modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
+++ b/modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
@@ -5,7 +5,7 @@ package gem
 
 import cats.{ Eq, Show }
 import cats.kernel.laws.discipline._
-import gem.laws.discipline._
+import monocle.law.discipline._
 import cats.tests.CatsSuite
 import gem.arb._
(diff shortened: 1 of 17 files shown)


modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
```diff
@@ -19,6 +20,7 @@ final class ProgramIdSpec extends CatsSuite {
 
   // Laws
   checkAll("Program.Id", OrderTests[Program.Id].order)
+  checkAll("Program.Id.Formats.Standard", FormatTests(Program.Id.Formats.Standard).format)
```
tpolecat: A problem here is that `Arbitrary[String]` is unlikely to produce many valid program ids, so we should write some better generators.

cquiroz:
Do you think this could be added to `Monocle` on the future?

modules/core/shared/src/test/scala/gem/laws/discipline/FormatTests.scala
```diff
@@ -0,0 +1,36 @@
...
+package discipline
+
+import cats.Eq
+import cats.instances.option._
+import gem.util.Format
+import org.scalacheck.Arbitrary
+import org.scalacheck.Prop._
+import org.typelevel.discipline.Laws
+
+trait FormatTests[A, B] extends Laws {
```
cquiroz: neat

cquiroz:
This looks great! I wonder if this is somewhat applicable to boopickle's binary format too

modules/core/shared/src/test/scala/gem/ProgramIdSpec.scala
tpolecat: Yeah, I think we need to do this in a few places. We can have local instances or pass them explicitly. `Arbitrary` is kind of a bogus typeclass because you often want specialized instances.

tpolecat: In any case I think another PR is the place to go back over the tests and improve generators.

tpolecat:
> Do you think this could be added to Monocle on the future?

There is an open issue https://github.com/julien-truffaut/Monocle/issues/186 about this. I think the issue is that when you compose `Format` with other optics (in both directions) you will end up with *other* partial optics that will need to get slotted into the hierarchy somewhere, and then *their* composition rules will have to be worked out, and so on.

So I'm not sure how hard this will be. I'll spend a little more time thinking about it.

tpolecat:
Ok so to deal with the bad test data problem I added a new `Format` "law" that requires your generator to provide data that parses but is not in normal form. It's not really a law but it ensures that you're actually testing the normalization behavior. I had to remove the `Format[String, ProgramId]` instance in fact because it doesn't normalize (i.e., it's a lawful `Prism` so `Format` is unnecessarily weak).

I also renamed the methods `parse` → `getOption` and `format` → `reverseGet` to make it consistent with `Prism`. They're not my favorite names but I think consistency here is important.

So what I'd like to do is verify this design and merge it, then go back and start adding all the instances. I don't want to do that work until we're ok with the setup here. To review, we have:

- The `Format` data type.
- The laws, as well as the code setup (based on cats) to let us define our own laws and check them.
- The instance for `Coordinates`, which is in an `Optics` module on the companion. My intent is that other optics go here as well. We will treat `Format` as the same kind of thing.

cquiroz:
👍 for the rename to be consistent with `Prism`

PR #185, opened 2018-01-05
MagnitudeValue

cquiroz:
This PR adds a class for `MagnitudeValue` represented as an Integer. This new value is used by `GpiObservingMode` which contains a couple of magnitude limits


modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
```diff
@@ -0,0 +1,27 @@
...
+package gem.math
+
+import cats.{ Order, Show }
+import cats.instances.int._
+
+/**
+ * Exact magnitude value represented as an int with the original value scaled up
+ * @param value This magnitude integral value
+ */
+final case class MagnitudeValue(val value: Int) extends Product with Serializable
```
cquiroz: I see case clases defined both with/without `extends Product with Serializable`. Should we unify the criteria?

modules/core/shared/src/test/scala/gem/arb/ArbMagnitudeValue.scala
```diff
@@ -0,0 +1,23 @@
...
+
+package gem
+package arb
+
+import gem.math.MagnitudeValue
+import org.scalacheck._
+import org.scalacheck.Arbitrary._
+import org.scalacheck.Cogen._
+
+trait ArbMagnitudeValue {
```
cquiroz: I'm using the class name as the name of the file `ArbMagnitudeValue`. I noted that `Arbitraries`' file names use the orignal name which is a bit confusing when you search by file name

modules/sql/src/main/scala/package.scala
```diff
@@ -27,6 +27,10 @@ package object sql {
   implicit val ZoneIdMeta: Meta[ZoneId] =
     Meta[String].xmap(ZoneId.of, _.toString)
 
-  implicit val MagnitudSystem: Meta[MagnitudeSystem] =
+  implicit val MagnitudSystemMeta: Meta[MagnitudeSystem] =
     Meta[String].xmap(MagnitudeSystem, _.id)
+
+  implicit val MagnitudeValueMeta: Meta[MagnitudeValue] =
```
cquiroz: The `Meta` doesn't seem to be used. Why do we need to define it?

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
swalker2m: Since this is 100x the actual magnitude maybe it shouldn't be visible, or at least not so easy to confuse with the actual magnitude.  Maybe call it `scaledValue` or something and add a `toDouble: Double` method that gets the actual magnitude?

modules/core/shared/src/test/scala/gem/arb/ArbMagnitudeValue.scala
swalker2m: Either way is fine with me, but we should be consistent.

modules/core/shared/src/test/scala/gem/math/WavelengthSpec.scala
```diff
@@ -13,7 +13,7 @@ final class WavelengthSpec extends CatsSuite {
   import ArbWavelength._
 
   // Laws
-  checkAll("Wavelength", EqTests[Wavelength].eqv)
+  checkAll("Wavelength", OrderTests[Wavelength].eqv)
```
swalker2m: Shouldn't that be `OrderTests[Wavelength].order`?

modules/core/shared/src/test/scala/gem/math/MagnitudeValueSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.{ Eq, Show, Order }
+import cats.kernel.laws.discipline._
+import gem.arb._
+
+@SuppressWarnings(Array("org.wartremover.warts.ToString"))
+final class MagnitudeValueSpec extends CatsSuite {
+  import ArbMagnitudeValue._
+
+  // Laws
+  checkAll("MagnitudeValue", OrderTests[MagnitudeValue].eqv)
```
swalker2m: `OrderTests[MagnitudeValue].order` ?

modules/sql/src/main/scala/MagnitudeValue.scala
```diff
@@ -0,0 +1,13 @@
...
+
+package gem.sql
+
+import java.math.RoundingMode.HALF_UP
+
+/** Minimal MagnitudeValue class to support reading from the enum
+  * tables.
+  */
+final case class MagnitudeValue(value: BigDecimal) extends Product with Serializable {
+  def toInt: Int = value.underlying.scaleByPowerOfTen(2).setScale(0, HALF_UP).intValue
```
swalker2m: Minor nitpick but I might call this `toScaledValue` or something since it is 100x `value`.

sraaphorst:
What is the reason for using integers instead of doubles? Is it to avoid floating point representation issues? It looks like everything is the double values multiplied by 100?

swalker2m:
> Is it to avoid floating point representation issues

Yes.  I'd like to get away from storing floating point values anywhere.  It's always problematic.

modules/core/shared/src/test/scala/gem/arb/ArbMagnitudeValue.scala
cquiroz: I'd like to change them just because of the ergonomics of editing. now when I search for a file Angle I get 5 hits

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
cquiroz: 👍

modules/core/shared/src/test/scala/gem/math/MagnitudeValueSpec.scala
cquiroz: doh!

modules/sql/src/main/scala/MagnitudeValue.scala
cquiroz: 👍

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
cquiroz: `scaledValue` is used on test. Would it be ok to declare it `private` for the `gem` package?

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
```diff
@@ -10,7 +10,9 @@ import cats.instances.int._
  * Exact magnitude value represented as an int with the original value scaled up
  * @param value This magnitude integral value
  */
-final case class MagnitudeValue(val value: Int) extends Product with Serializable
+final case class MagnitudeValue(private[gem] val scaledValue: Int) extends Product with Serializable {
+  def toDouble: Double = scaledValue / 10.0
```
swalker2m: Shouldn't this be `/ 100.0`?  When you read the value from the database it is scaled up by 10^2 so it should be scaled down here by the same.

cquiroz: fixed

modules/core/shared/src/test/scala/gem/arb/ArbMagnitudeValue.scala
tpolecat: Let's do a PR that does nothing but rename files, since it will create gigantic deltas.

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
```diff
@@ -0,0 +1,29 @@
...
+package gem.math
+
+import cats.{ Order, Show }
+import cats.instances.int._
+
+/**
+ * Exact magnitude value represented as an int with the original value scaled up
+ * @param value This magnitude integral value
+ */
+final case class MagnitudeValue(private[gem] val scaledValue: Int) extends Product with Serializable {
```
tpolecat: I don't think `scaledValue` needs to be private (and there's no point really because you can still access it with pattern-matching, although `sealed abstract case class` would remedy this).

To be consistent with similar classes I'd prefer this member be called `toCentiwhatever` (whatever the units are, which should also be noted in the class doc) and similarly `toDouble` should be `toDoubleWhatever`.

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
```diff
@@ -0,0 +1,29 @@
...
+  def toDouble: Double = scaledValue / 100.0
+}
+
+object MagnitudeValue {
+
+  final lazy val ZeroMagnitude = MagnitudeValue(0)
+
+  /** @group Typeclass Instances */
+  implicit val MagnitudeValueShow: Show[MagnitudeValue] =
+    Show.fromToString
```
tpolecat: I would prefer that the `toString` show the value humans want to see; something like `s"MagnitudeValue($toDouble)"`

modules/sql/src/main/scala/package.scala
tpolecat: Yeah the only ones we need are `Duration` and `ZoneId`. You can delete the others.

modules/core/shared/src/main/scala/gem/math/MagnitudeValue.scala
cquiroz: Magnitude doesn't really have units

tpolecat: Huh, ok. It's weird but I guess it's fine as-is. Can we make a note in the doc that it's dimensionless and maybe link to an article somewhere? It will be helpful to future me.

tpolecat:
👍  Thanks!

PR #186, opened 2018-01-05
TCS Ephemeris Export

swalker2m:
Support for exporting a night of ephemeris data to a file for consumption by the TCS.  At some point, I'll add telnet commands for this and ultimately it will be driven by a cronjob.


modules/ephemeris/src/main/scala/gem/horizons/tcs/TcsEphemerisExport.scala
```diff
@@ -0,0 +1,97 @@
...
+    * @param start start time for ephemeris data; if there is an element at
+    *              exactly this time it will be the first element (otherwise,
+    *              the element immediately preceeding this time is included)
+    * @param end   end time for ephemeris data; if there is an element at
+    *              exactly this time it will be the last element (otherwise,
+    *              the element immediately following this time is included)
+    */
+  def exportOne(path: Path, key: EphemerisKey, site: Site, start: Instant, end: Instant): M[Unit] = {
+    def query(s: InstantMicros, e: InstantMicros) =
+      EphemerisDao
```
cquiroz: nice!

cquiroz:
I can't comment much on the details but it seems great
Should there be a test validating the exact fileoutput for the TCS, being it so strict on the requirements?

modules/ephemeris/src/main/scala/gem/horizons/tcs/TcsEphemerisExport.scala
```diff
@@ -0,0 +1,97 @@
...
+        .to(file.writeAll(path, List(CREATE, TRUNCATE_EXISTING)))
+        .run
+
+    val s = InstantMicros.truncate(start)
+    val e = InstantMicros.truncate(end)
+
+    for {
+      r <- EphemerisDao.bracketRange(key, site, s, e).transact(xa)
+      _ <- query(r._1, r._2)
+    } yield ()
```
tpolecat: Seems like you should be able to do this in a single transaction. Something like 

```scala
Stream.eval(EphemerisDao.bracketRange(key, site, s, e)).flatMap { case (s, e) => 
  streamRange(key, site, s, e)
}
```

swalker2m:
@tpolecat @cquiroz Thanks for the comments.  I think I've addressed them now and they improved the PR a bit.

tpolecat:
Looks great!

PR #187, opened 2018-01-08
sbt 1.1 and cats 1.0

tpolecat:
This updates to sbt 1.1 and cats 1.0

fs2 is still on a milestone build, as are http4s and doobie which depend on it. There were a few API changes that I think I have fixed correctly.
  
Note that sbt-updates thinks 1.5.0 is newer than 1.5.0-cats so we'll get a warning about monocle, possibly forever. Oh well.


modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
```diff
@@ -33,11 +33,8 @@ object HorizonsClient {
   /** Horizons permits only one request at a time from a given host, so all
     * requests should go through this client.
     */
-  val client: Client[IO] =
-    PooledHttp1Client[IO](
-      maxTotalConnections = 1,
-      config = BlazeClientConfig.defaultConfig.copy(responseHeaderTimeout = 20.seconds)
```
swalker2m: Will there still be 1 `maxTotalConnections`?  I had specified that in hopes of complying with the JPL limit on connections from a single client.

tpolecat: Ah, I don't think so. I will try to figure that out.

tpolecat:
Ok I added a asynchronous mutex to the stream that provides the http4s `Client` used by `HorizonsClient`. It uses `Stream.bracket` which should (heh-heh) guarantee that the lock is released. I wrote a test that shows serial behavior even if you ask for parallel, which I *think* shows what we want but pls give it a think.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
```diff
@@ -30,11 +31,14 @@ import scala.concurrent.duration._
...
+  /** A stream that emits a single client that will be shut down automatically. An internal global
+    * mutex ensures that only one such stream exists at a time.
     */
-  val client: IO[Client[IO]] =
-    Http1Client[IO](BlazeClientConfig.defaultConfig.copy(responseHeaderTimeout = 20.seconds))
+  val client: Stream[IO, Client[IO]] = {
+    val mutex  = Semaphore[IO](1L)(implicitly, scala.concurrent.ExecutionContext.global).unsafeRunSync // note
+    val client = Http1Client.stream[IO](BlazeClientConfig.defaultConfig.copy(responseHeaderTimeout = 20.seconds))
+    Stream.bracket(mutex.decrement)(_ => client, _ =>  mutex.increment)
+  }
```
swalker2m: This is very clever.  I don't think I would have come up with that solution.  It streams clients but only one at a time due to the `mutex`, correct?

tpolecat: Correct. We take the mutex before running the `client` stream, then when the stream terminates the client is cleaned up (internally it also uses `bracket` to guarantee this) and then the lock is released. If the lock is already taken we don't block the thread. We're suspended and later rescheduled on the mutex's `ExecutionContext`.

tpolecat:
Fixed line endings and also found the editor setting that should prevent this in the future.

tpolecat:
I changed `mutex` slightly to make it slightly more futureproof.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
```diff
@@ -31,13 +31,19 @@ import scala.concurrent.duration._
...
+    Semaphore[IO](1L)(implicitly, scala.concurrent.ExecutionContext.global).unsafeRunSync // note
+
+  /** A stream that emits a single client that will be shut down automatically. An global mutex
+    * ensures that only one such stream exists at a time.
     */
   val client: Stream[IO, Client[IO]] = {
-    val mutex  = Semaphore[IO](1L)(implicitly, scala.concurrent.ExecutionContext.global).unsafeRunSync // note
+    // By using the singleton type here we prove that the mutex is a constant value. This guards
+    // against someone accidentally refactoring it into a method.
+    val mutexʹ: mutex.type = mutex
```
cquiroz: Can you explain a bit about `singleton types`? Is it enabled by TLS?

tpolecat:
Hell. The build server only has one thread so the concurrency counterexample fails.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
tpolecat: Not TLC; it's been in Scala for a long time. The TL;DR is that a constant references have singleton types, so we're using that type explicitly to prove this.

Singleton types arise in a few places.

- The built-in literals have them, but that's not what we're doing here.

```scala
scala> val a = 42
a: Int = 42

scala> final val a = 42 // note the return type
a: Int(42) = 42

scala> val a: 42 = 42 // TLS only
a: 42 = 42

scala> val a: shapeless.Witness.`42`.T = 42
a: Int(42) = 42
```

- `object`s have singleton types.

```scala
scala> :t List
scala.collection.immutable.List.type

scala> val x: List.type = List
x: scala.collection.immutable.List.type = scala.collection.immutable.List$@323a351c

scala> x(1,2,3)
res0: List[Int] = List(1, 2, 3)
```

- *variables* have them. This is what we're doing here. Note that I didn't say *values*, although if `a.type =:= b.type` then certainly `a == b`, but it's much stronger than this. It means they're the same value. Note that singleton types won't normally be inferred.

```scala
scala> val a = "foo"
a: String = foo

scala> val b = "foo"
b: String = foo

scala> a == b
res2: Boolean = true

scala> implicitly[a.type =:= b.type]
<console>:14: error: Cannot prove that a.type =:= b.type.
       implicitly[a.type =:= b.type]
                 ^

scala> val b = a
b: String = foo

scala> implicitly[a.type =:= b.type]
<console>:14: error: Cannot prove that a.type =:= b.type.
       implicitly[a.type =:= b.type]
                 ^

scala> val b: a.type = a
b: a.type = foo

scala> implicitly[a.type =:= b.type]
res5: a.type =:= b.type = <function1>
```

- Note further than `var`s and `def`s don't have singleton types, which would be unsound because the values can change and thus their types would need to change too, which can't happen.

```scala
scala> var a = "foo"
a: String = foo

scala> val b: a.type = a
<console>:12: error: stable identifier required, but a.type found.
       val b: a.type = a
               ^

scala> def a = "foo"
a: String

scala> val b: a.type = a
<console>:12: error: stable identifier required, but a.type found.
       val b: a.type = a
               ^
```

There's a lot more to say about it but hopefully this answers your question.

cquiroz: Great explanation. thanks!

cquiroz:
@tpolecat I set the test to run serially in js
https://github.com/gemini-hlsw/gem/blob/master/build.sbt#L218

I don't think that would affect the other test though

swalker2m:
This looks good to me pending some solution for running that one test case on travis.

tpolecat:
My explanation doesn't make sense actually. Maybe I got unlucky since it's nondeterministic. I'll re-run and lower the odds if it still fails.

tpolecat:
Ok I made it much less likely that this test will fail again so I think we should merge and see how it goes. If it keeps failing with any regularity I'll rethink it.

PR #188, opened 2018-01-09
JSON support for Target

swalker2m:
Trying to add a `TargetEnvironment` to an `Observation`, I came across an issue with JSON support. `Target` was missing the required codec definitions.  This turned out to be a little more involved than I assumed because:

* We cannot have a single implicit `Angle` codec because there are so many different contexts in which `Angle` is used and each needs its own representation.

* There is some issue with deriving codecs for `Map[K, V]` even when there are codecs for `K` and `V`.  I'm probably doing something wrong but I ended up having to write them out explicitly as lists of tuples.

At any rate, I submit this PR for further discussion.  Maybe we want to do this differently now and not rely so much on auto-derivation?

PS, I'm not sure what's up with the `PublicInference` warts that I had to turn off.  I think it might be a bug.

### DIFF
--- a/modules/json/src/main/scala/gem/json.scala
+++ b/modules/json/src/main/scala/gem/json.scala
@@ -4,14 +4,15 @@
 package gem
 
 import cats.data.OneAnd
-import gem.enum.GcalArc
+import gem.enum.{ GcalArc, Site }
 import gem.config.GcalConfig.GcalArcs
 import gem.config.GmosConfig._
-import gem.math.{ Angle, Offset, Wavelength }
-import gem.util.Enumerated
+import gem.math._
+import gem.util.{ Enumerated, InstantMicros }
 import io.circe._
 import io.circe.syntax._
-import java.time.Duration
+import io.circe.generic.auto._
+import java.time.{ Duration, Instant }
 import scala.collection.immutable.TreeMap
 
 // These json codecs are provided for primitive types that have no natural mapping that would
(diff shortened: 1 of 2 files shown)


tpolecat:
> Maybe we want to do this differently now and not rely so much on auto-derivation?

Agree. The Json stuff was just kind of a stub to ensure we could go end to end, but it was never really designed. I think we should probably use semi-automatic derivation for things with reasonable default encodings and explicit implementations where necessary, hopefully leaning on the optics stuff I'm working on in #184. This will also speed up compilation, which is really ridiculously slow if you auto-derive everything.

I would like to provide semi-automatic derivation as an option for doobie 0.6 because we run into the same issue with database serialization. People are starting to realize the implicit codec pattern you see with all these libraries ends up being an issue, and something closer to parser combinators (which are never implicit) might work better. But on the other hand things like [SIP-35](https://docs.scala-lang.org/sips/opaque-types.html) would allow us to do this via lightweight newtyping, and if we were to start using [refinement types](https://github.com/fthomas/refined) there would be fewer ambiguous cases. It's a big and unclear design space.

> PS, I'm not sure what's up with the `PublicInference` warts that I had to turn off. I think it might be a bug.

The mechanism used by WartRemover here is suuuuuper sketchy and doesn't really work with synthetic code. So anywhere you have a maco in the mix it's likely to give false warnings. I have heard they're re-doing it with scalameta, which can see the parse tree and can tell you for real whether there's a type annotation or not. Until then I think we should just deal with it. I think the benefit of this wart outweighs the occasional false positive.

cquiroz:
Auto derivation of codecs may create issues on the .js side as sometimes the code size explodes. It's a bit hard to tell if that happens just looking at the code though

modules/json/src/main/scala/gem/json.scala
```diff
@@ -22,18 +23,34 @@ import scala.collection.immutable.TreeMap
 package object json {
 
   private implicit class MoreAngleOps(a: Angle) {
+    private def moveRight(n: Int): BigDecimal =
+      new java.math.BigDecimal(a.toMicroarcseconds).movePointRight(n)
```
cquiroz: What's a `movePointRight`? scaling?

tpolecat: Yes, fixed point math. Multiplication by multiple of 10 with no rounding error.

swalker2m:
I fixed the failing test case in another PR.  If I can merge that then I'll rebase this one.  Whether we want to continue doing JSON this way or not, I'll either need to merge this or else turn off the compile test for observation in order to continue adding targets.

swalker2m:
In order to associate targets with observations and still be able to run test cases, I'll need to do something.  Which do we prefer?

1) Merge this PR to add JSON codecs for target.

2) Remove [this line](https://github.com/gemini-hlsw/gem/blob/master/modules/json/src/test/scala/gem/json/CompilationTest.scala#L31).

3) Something else?

cquiroz:
Slightly unrelated but i think the `json` module should be compiled to js too

Issue #190, opened 2018-01-10
Clip InstantMicros


swalker2m:
`InstantMicros` was invented to work around the fact that SQL `timestamp` supports only microsecond resolution whereas `java.time.Instant` has nanosecond resolution.   To reliably round-trip time values, we truncate `Instant` at microseconds.  Unfortunately postgres `timestamp`s are also [more limited in range](https://www.postgresql.org/docs/9.5/static/datatype-datetime.html) than `Instant`s and therefore `InstantMicros`.

This task is to clip `InstantMicros` to values that can be represented in a `timestamp`.  This implies that constructing an `InstantMicros` will have to produce an `Option` and reject values outside of the range of `timestamp`.

swalker2m:
Solved by #191

PR #191, opened 2018-01-11
Database-safe Timestamps

swalker2m:
This PR implements issue #190.  Because `Instant` covers a range of dates that is much larger than those supported by the postgres timestamp,  we were able to create times that could not be written to the database or used in queries.   Here I limit the valid range of `Instant`s for creating timestamps to roughly those supported by the database and rename `InstantMicros` to `Timestamp` to reflect the fact that it more closely mirrors the postgres `timestamp` type.

The `Option`-returning API is a bit annoying.  Given that `Instant` is also limited in time range and can throw exceptions when adding milliseconds, etc. I also considered just having `Timestamp` behave like `Instant` with a smaller time range.  That felt wrong but I'm not sure and would welcome feedback.

Also, it occurs to me that the name `Timestamp` is used in a couple of other contexts including the JDBC `Timestamp` so perhaps a different name would be better?

### DIFF
--- a/modules/core/shared/src/main/scala/gem/util/Timestamp.scala
+++ b/modules/core/shared/src/main/scala/gem/util/Timestamp.scala
@@ -12,15 +12,15 @@ import java.time.ZoneOffset.UTC
 import java.time.temporal.ChronoUnit.MICROS
 
 
-/** InstantMicros wraps a `java.util.Instant` that is truncated to microsecond
-  * resolution.  This allows InstantMicros to roundtrip to/from the database
-  * where timestamps support only microsecond resolution.  In addition the min
+/** Timestamp wraps a `java.util.Instant` that is truncated to microsecond
+  * resolution.  This allows Timestamps to roundtrip to/from the database
+  * where only microsecond resolution is supported.  In addition the min
   * and max instants that are supported are determined by th postgres limit for
   * timestamps.
   *
   * @param toInstant
   */
-final class InstantMicros private (val toInstant: Instant) extends AnyVal {
+final class Timestamp private(val toInstant: Instant) extends AnyVal {
 
   /** Gets the number of seconds from the Java epoch of 1970-01-01T00:00:00Z. */
   def epochSecond: Long =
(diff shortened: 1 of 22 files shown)


modules/core/shared/src/main/scala/gem/util/Timestamp.scala
```diff
@@ -0,0 +1,127 @@
...
+
+object Timestamp {
+
+  /** Minimum time that can be stored in a postgres timestamp. */
+  val Min: Timestamp =
+    truncate(ZonedDateTime.of( -4712, 1, 1, 0, 0, 0, 0, UTC).toInstant)
+
+  /** Maximum time that can be stored in a postgres timestamp. */
+  val Max: Timestamp =
+    truncate(ZonedDateTime.of(294275, 12, 31, 23, 59, 59, 999999000, UTC).toInstant)
```
swalker2m: I don't think this is the exact upper limit, but the crucial point is that there _is a limit_ and it is close to this value.

modules/core/shared/src/main/scala/gem/util/Timestamp.scala
```diff
@@ -0,0 +1,127 @@
...
+    new Timestamp(i.truncatedTo(MICROS))
+
+  /** Creates a Timestamp representing the current time, truncated to the
+    * last integral number of microseconds.
+    *
+    * @group Constructors
+    */
+  def now: IO[Timestamp] =
+    IO {
+      truncate(Instant.now())
```
swalker2m: Obviously this has a year 294276 bug but I decided not to worry too much about that.

cquiroz:
`Timestamp` is used all over java/scala libs though it's hard to think about a better synonym

modules/core/shared/src/main/scala/gem/util/Timestamp.scala
```diff
@@ -0,0 +1,127 @@
...
+    */
+  def unsafeFromInstant(i: Instant): Timestamp =
+    fromInstant(i).getOrElse(sys.error(s"$i out of Timestamp range"))
+
+  /** Creates a Timestamp from the given Instant, assuring that the time value
+    * recorded has a round number of microseconds. The caller must verify that
+    * the timestamp is in the valid range.
+    */
+  private def truncate(i: Instant): Timestamp =
+    new Timestamp(i.truncatedTo(MICROS))
```
tpolecat: Can we inline this into `fromInstant` and then use `unsafeFromInstant` in the other cases here that use `truncate`? I would prefer to have the scale and limit validation together.

modules/core/shared/src/main/scala/gem/util/Timestamp.scala
```diff
@@ -0,0 +1,127 @@
...
+/** Timestamp wraps a `java.util.Instant` that is truncated to microsecond
+  * resolution.  This allows Timestamps to roundtrip to/from the database
+  * where only microsecond resolution is supported.  In addition the min
+  * and max instants that are supported are determined by the postgres limit for
+  * timestamps.
+  *
+  * @param toInstant
+  */
+final class Timestamp private(val toInstant: Instant) extends AnyVal {
+
```
tpolecat: Can we add range and scale assertions here as a sanity check, as we did for a few other types?

swalker2m: Apparently assertions aren't legal in value classes, but I could just lose the value class-ness of it.

tpolecat:
I'm not super concerned about the name because we're not using `java.sql.Timestamp` and I can't see any usage sneaking past code review. So I think the name is ok.

modules/core/shared/src/main/scala/gem/util/Timestamp.scala
tpolecat: Oh I didn't notice it was a value class. Yeah I would be ok dropping the `extends AnyVal`.

tpolecat:
I will note for the record that we *could* implement this as a [refinement](https://github.com/fthomas/refined) of `Instant`. At this point I'm still not convinced it's worth the additional dependency, but if we accumulate many more of these constrained value classes I think we might want to change strategy.

modules/core/shared/src/main/scala/gem/util/Timestamp.scala
swalker2m: I just went with the `sealed abstract case class` trick to avoid having to write `equals` and `hashCode`, which would have required an `Eq` instance for `Instant` to be defined somewhere due to the `[wartremover:Equals]` check.

tpolecat:
Looks great! 👍

tpolecat:
Probably random. I restarted it.

PR #192, opened 2018-01-11
format for dataset label

tpolecat:
This adds a `Format[String, DataSet.Label]`. A few things to consider:
- I sliced the `Optics` module off into its own trait just to keep the main companion from getting too cluttered. I'm not sure I like that convention much better but it was worth trying.
- I'm not sure what to do with existing `.format` and `.[unsafe]FromString` methods because they're unmoored from their `Format` even if they delegate to it. Do we want users to be explicit and say `Foo.Optics.fromBlah.reverseGet(foo)` rather than `foo.formatBlah`?


modules/core/shared/src/main/scala/gem/Dataset.scala
```diff
@@ -58,6 +57,24 @@ object Dataset {
       Show.fromToString
 
   }
+  trait LabelOptics {
+    object Optics {
+
+      /** Format from Strings into Label and back. */
+      val fromString: Format[String, Label] =
```
cquiroz: Why not putting `fromString` directly into `LabelOptics`?

modules/core/shared/src/test/scala/gem/DatasetLabelSpec.scala
```diff
@@ -7,13 +7,15 @@ import cats.tests.CatsSuite
...
 import gem.arb._
+import gem.laws.discipline._
 
 @SuppressWarnings(Array("org.wartremover.warts.ToString"))
 final class DatasetLabelSpec extends CatsSuite {
   import ArbDataset._
 
   // Laws
   checkAll("DatasetLabel", OrderTests[Dataset.Label].order)
+  checkAll("Optics.fromString", FormatTests(Dataset.Label.Optics.fromString).formatWith(strings))
```
cquiroz: Is `formatWith` part of scalacheck? or did you add it to the test package?

modules/core/shared/src/main/scala/gem/Dataset.scala
swalker2m: I kind of like having the `Optics` namespace for these.  I'm not sure I have a strong opinion on keeping the delegating `fromString` and `format` methods.  Part of me wants to get rid of them to avoid having all that extra code. Part of me thinks they are convenient and more intuitive than  calls that end in `reverseGet` and `getOption`.   I guess we'd get used to it though.  If forced to decide I guess I would remove the delegating methods and just put the `Optics` module directly in the companion instead of in a separate `trait`.

modules/core/shared/src/test/scala/gem/DatasetLabelSpec.scala
tpolecat: It's our code but it follows the [pattern for law checking](https://typelevel.org/cats/typeclasses/lawtesting.html) used by cats. `FormatTests` is a factory that constructs a `RuleSet` that's evaluated by Scala(Test|check).

tpolecat:
Huh incremental compiler missed that one. Anyway when it's green it's ready for another review. I'm working ahead on other types.

tpolecat:
Ok this is ready for another look. The incremental compiler was cheating me.

modules/db/src/main/scala/gem/dao/meta/DatasetLabel.scala
```diff
@@ -7,10 +7,11 @@ import doobie._
 import gem.Dataset
 
 trait DatasetLabelMeta {
+  import FormatMeta._
 
   // Dataset.Label as string
   implicit val DatasetLabelMeta: Meta[Dataset.Label] =
-    Meta[String].xmap(Dataset.Label.unsafeFromString, _.format)
+    Dataset.Label.Optics.fromString.asMeta
```
swalker2m: Cool!

PR #194, opened 2018-01-17
Target Optics and Eq

swalker2m:
A small update to the target model to incorporate optics and `Eq` instances.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Target.scala
+++ b/modules/core/shared/src/main/scala/gem/Target.scala
@@ -3,5 +3,18 @@
 
 package gem
 
+import gem.enum.Site
+import gem.math.Ephemeris
+
+import monocle.Optional
+import monocle.macros.Lenses
+
 /** A target of observation. */
-final case class Target(name: String, track: Track)
+@Lenses final case class Target(name: String, track: Track)
+
+@SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+object Target {
+
+  val ephemerides: Optional[Target, Map[Site, Ephemeris]] =
+    track composeOptional Track.ephemerides
+}
--- a/modules/core/shared/src/main/scala/gem/Target.scala
+++ b/modules/core/shared/src/main/scala/gem/Target.scala
@@ -3,6 +3,8 @@
 
 package gem
 
+import cats.Eq
+
 import gem.enum.Site
 import gem.math.Ephemeris
 
@@ -17,4 +19,7 @@ object Target {
 
   val ephemerides: Optional[Target, Map[Site, Ephemeris]] =
     track composeOptional Track.ephemerides
+
+  implicit val EqTarget: Eq[Target] =
+    Eq.fromUniversalEquals
 }
--- a/modules/core/shared/src/main/scala/gem/Track.scala
+++ b/modules/core/shared/src/main/scala/gem/Track.scala
@@ -8,6 +8,9 @@ import gem.math._
 import gem.util.Timestamp
 import java.time.Instant
 
+import monocle.{ Optional, Prism }
+import monocle.macros.{ Lenses, GenPrism }
+
 /**
  * Time/site-parameterized coordinates over a span of time. This generalizes proper motion and
  * ephemerides.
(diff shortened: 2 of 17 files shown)


modules/core/shared/src/main/scala/gem/Target.scala
```diff
@@ -3,5 +3,23 @@
...
+
 /** A target of observation. */
-final case class Target(name: String, track: Track)
+@Lenses final case class Target(name: String, track: Track)
+
+@SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+object Target {
+
+  val ephemerides: Optional[Target, Map[Site, Ephemeris]] =
+    track composeOptional Track.ephemerides
```
swalker2m: They are kind of long but I think I prefer the `composeX` methods over the corresponding weathervane operators like `^|-?` in the same way that we prefer `foldLeft` over `/:`.

sraaphorst: Yes, please use `composeX` over `^|-?`. As someone who is trying to keep an eye on OCS3 / gems so as to not fall behind, obscure operators make it much more difficult.

modules/core/shared/src/main/scala/gem/Track.scala
```diff
@@ -51,11 +65,27 @@ object Track {
...
+    GenPrism[Track, Sidereal]
+
+  val nonsidereal: Prism[Track, Nonsidereal] =
+    GenPrism[Track, Nonsidereal]
+
+  val ephemerides: Optional[Track, Map[Site, Ephemeris]] =
+    nonsidereal composeLens Nonsidereal.ephemerides
+
+  implicit val EqTrack: Eq[Track] =
+    Eq.fromUniversalEquals
```
jdnavarro: What does this do?

modules/core/shared/src/test/scala/gem/arb/ProperMotion.scala
```diff
@@ -30,5 +32,10 @@ trait ArbProperMotion {
       } yield ProperMotion(cs, ap, pv, rv, px)
     }
 
+  implicit val cogProperMotion: Cogen[ProperMotion] =
+    Cogen[(Coordinates, Epoch, Option[Offset], Option[RadialVelocity], Option[Angle])].contramap { p =>
+      (p.baseCoordinates, p.epoch, p.properVelocity, p.radialVelocity, p.parallax)
+    }
+
```
jdnavarro: Is it a scalacheck requirement (or recommendation to find the simplest failing test case) to define a `Cogen` for `ProperMotion` in order to run the Arbitrary tests? If that's not the case where will the Cogen instance be used?

modules/core/shared/src/main/scala/gem/Track.scala
tpolecat: This constructs an `Eq` instance that delegates to universal equality (i.e, Scala `==`). This is ethical for case classes whose members also have sane universal equality.

modules/core/shared/src/main/scala/gem/Target.scala
tpolecat: If we end up doing a lot of composition of optics we may find that the short versions work out better, but this is fine.

modules/core/shared/src/test/scala/gem/arb/ProperMotion.scala
tpolecat: `Shrink` is used to minimize test cases but you don't normally need to specify it.

`Cogen` is used to generate interesting arbitrary functions. Given `Cogen[A]` and `Gen[B]` you can get `Gen[A => B]`. The idea is that `Cogen` takes an `A` and returns a hash that can be absorbed into the random seed that's passed on to the `Gen[B]`, so the incoming `A` helps to determine the returned `B`. Basically `type Cogen[A] = A => Long`. This [talk](https://www.youtube.com/watch?v=UlZGZh7hfbs) by Erik Osheim explains it very well starting around 36:00.

modules/core/shared/src/main/scala/gem/Target.scala
```diff
@@ -3,5 +3,23 @@
...
+
+  implicit val EqTarget: Eq[Target] =
```
tpolecat: If we can define `Order[Target]` we can use `TreeSet` in the `TargetEnvironment` which is a bit more sane. It would also give us a deterministic ordering which might be good. We also may be able to define that traversal you were looking for. Checking with Julien.

swalker2m: I'll circle back to the `Order[Target]` and `TreeSet` idea.

Issue #176, opened 2017-12-21
Ephemeris in Nonsidereal Target


swalker2m:
At the moment the target model includes a `Nonsidereal` `Track` instance that associates a unique ephemeris key with a map of ephemeris data.

```scala
final case class Nonsidereal(ephemerisKey: EphemerisKey, ephemerides: Map[Site, Ephemeris]) extends Track {

  override def at(time: Instant, s: Site): Option[Coordinates] =
    ephemeris(s).flatMap(_.get(InstantMicros.truncate(time)).map(_.coord))

  ...
}
```

When such a target is read from or written to the database the ephemeris data is not included.   We'd like to avoid necessarily having to read ephemeris data every time we read targets, and yet we need the information to compute coordinates.

I'm opening this issue to discuss options for how to proceed.

tpolecat:
The issue I keep thinking about is that when we load an ephemeris we're going to want to provide a timespan and perhaps a resolution, to minimize the amount of data we're moving. And in general we can't know this statically … a high-res ephemeris for today and a long-range one for next semester will have the same type. So if we frame the question as "how much do we load?" rather than "do we load it at all?" the current setup works. When you load a target you have to specify a date range and if the range is empty then ¯\\_(ツ)_/¯ ok your ephemeris is empty.

Re: storing I think we probably don't ever want to store Horizons ephemerides since they're owned by the robot, but I guess it makes sense to save custom ones when we store the target.

What do you think? I'm not really at a point of clarity yet.

swalker2m:
If a `TargetEnvironment` is part of an `Observation` this implies that we need to specify a time range (or `None` if we don't care about coordinates) when we load an observation as well.  Does that sound right?

tpolecat:
I think this is right, although maybe we just want to pass an empty range instead … offhand I can't think of a good reason to distinguish `Some(emptyRange)` from `None`.

swalker2m:
One issue with this approach is that you have to understand what you're getting with equality comparisons.  You can have two nonsidereal targets with the same id and yet they will not be considered equal unless you happen to have loaded the same set of ephemeris data for each.

I'm not sure whether this should be considered surprising or incorrect, but it took me a while to figure out what was going wrong in test cases.

swalker2m:
[...]

PR #195, opened 2018-01-17
Target Update Proposal

swalker2m:
This PR isn't ready to be merged (👎) so I'm not looking for approval in that sense.  It fleshes out the latest idea in #176 a bit though and I thought it would be good to get some feedback before continuing.

Highlights (?):

* Changes `Target` to replace `Track` with an `Either[EphemerisKey, ProperMotion]`.  I'm not sure about this idea but at this point having an additional layer wrapping the key or proper motion didn't seem right either.

* In doing so, it mangles the `TargetDao` a bit.  I feel bad about that.  It can probably be done correctly with the `TaggedCoproduct` idea that was there?

* Adds `TrackDao` to have a single API for getting a `Track` for coordinate lookups.  The sidereal case is trivial there of course and the nonsidereal case just delegates to `EphemerisDao`.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Target.scala
+++ b/modules/core/shared/src/main/scala/gem/Target.scala
@@ -5,20 +5,15 @@ package gem
 
 import cats.Eq
 
-import gem.enum.Site
-import gem.math.Ephemeris
+import gem.math.ProperMotion
 
-import monocle.Optional
 import monocle.macros.Lenses
 
 /** A target of observation. */
-@Lenses final case class Target(name: String, track: Track)
+@Lenses final case class Target(name: String, track: Either[EphemerisKey, ProperMotion])
 
 @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
 object Target {
-
-  val ephemerides: Optional[Target, Map[Site, Ephemeris]] =
-    track composeOptional Track.ephemerides
 
   implicit val EqTarget: Eq[Target] =
     Eq.fromUniversalEquals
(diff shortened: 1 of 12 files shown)


modules/core/shared/src/main/scala/gem/Target.scala
```diff
@@ -5,21 +5,16 @@ package gem
...
-import gem.enum.Site
-import gem.math.Ephemeris
+import gem.math.ProperMotion
 
-import monocle.Optional
 import monocle.macros.Lenses
 
 /** A target of observation. */
-@Lenses final case class Target(name: String, track: Track)
+@Lenses final case class Target(name: String, track: Either[EphemerisKey, ProperMotion])
```
tpolecat: This seems fine to me.

modules/db/src/main/scala/gem/dao/TargetDao.scala
```diff
@@ -36,59 +35,70 @@ object TargetDao extends EnumeratedMeta /* extend EnumeratedMeta to lower the pr
...
       sql"""
         SELECT name, track_type,
-               ra, dec, epoch, pv_ra, pv_dec, rv, px, -- proper motion
-               e_key_type, e_key                      -- ephemeris key
+               e_key_type, e_key,                     -- ephemeris key
+               ra, dec, epoch, pv_ra, pv_dec, rv, px  -- proper motion
           FROM target
          WHERE id = $id
-      """.query[Target]
+      """.query[TargetKernel].map(_.toTarget)
```
tpolecat: This could be done directly with `TaggedCoproduct` but it's not important. We can go back later if we want to.

PR #196, opened 2018-01-17
unsafeGet for Prism and Format

tpolecat:
This adds `unsafeGet` methods for `Format` and (via syntax) for `Prism`. A sleazy trick gives us better error messages when types are known.

```
// unsafe get from a Prism
scala> Track.sidereal.unsafeGet(Track.Nonsidereal.empty(EphemerisKey.Comet("foo")))
java.lang.IllegalArgumentException: gem.Track.Sidereal: unsafeGet failed: Nonsidereal(Comet(foo),Map())
  at gem.util.Format.$anonfun$unsafeGet$1(Format.scala:26)
  at scala.Option.getOrElse(Option.scala:121)
  at gem.util.Format.unsafeGet(Format.scala:23)
  at gem.syntax.PrismOps$.unsafeGet$extension(Prism.scala:20)
  ... 39 elided

// Unsafe get from a Format
scala> Coordinates.Optics.fromHmsDms.unsafeGet("x")
Coordinates.Optics.fromHmsDms.unsafeGet("x")
java.lang.IllegalArgumentException: gem.math.Coordinates: unsafeGet failed: x
  at gem.util.Format.$anonfun$unsafeGet$1(Format.scala:26)
  at scala.Option.getOrElse(Option.scala:121)
  at gem.util.Format.unsafeGet(Format.scala:23)
  ... 39 elided

// With generic types it still works but the error message is less good
scala> def foo[A,B](f: Format[A,B], a: A) = f.unsafeGet(a)
def foo[A,B](f: Format[A,B], a: A) = f.unsafeGet(a)
foo: [A, B](f: gem.util.Format[A,B], a: A)B

scala> foo(Coordinates.Optics.fromHmsDms, "x")
foo(Coordinates.Optics.fromHmsDms, "x")
java.lang.IllegalArgumentException: unsafeGet failed: x
  at gem.util.Format.$anonfun$unsafeGet$1(Format.scala:26)
  at scala.Option.getOrElse(Option.scala:121)
  at gem.util.Format.unsafeGet(Format.scala:23)
  at .foo(<console>:21)
  ... 39 elided
```


modules/core/shared/src/main/scala/gem/util/Format.scala
```diff
@@ -4,16 +4,28 @@
...
  */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
 final case class Format[A, B](getOption: A => Option[B], reverseGet: B => A) {
 
+  /** Like getOption, but throws IllegalArgumentException on failure. */
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments", "org.wartremover.warts.Throw"))
+  def unsafeGet(a: A)(implicit ev: TypeTag[B] = null): B =
+    getOption(a).getOrElse {
+      // If a TypeTag is available we can construct a more useful message.
+      val prefix = Option(ev).foldMap(tag => s"${tag.tpe}: ")
```
cquiroz: nice

modules/core/shared/src/main/scala/gem/util/Format.scala
```diff
@@ -4,16 +4,28 @@
...
  * need not be injective; i.e., distinct inputs may have the same getOption result, which combined
  * with a subsequent `reverseGet` yield a normalized form for A. Composition with stronger optics
  * (`Prism` and `Iso`) yields another `Format`.
  */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
 final case class Format[A, B](getOption: A => Option[B], reverseGet: B => A) {
 
+  /** Like getOption, but throws IllegalArgumentException on failure. */
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments", "org.wartremover.warts.Throw"))
+  def unsafeGet(a: A)(implicit ev: TypeTag[B] = null): B =
```
cquiroz: Can you tell a bit more about the default `null` argument? when would it be set?

tpolecat: If there's a `TypeTag[B]` in implicit scope (which will be the case if `B` is a concrete type) then it will be passed, otherwise `null` will be passed. So this lets us say ok if `B` is instantiated to a concrete type then we can include that information in the error message, otherwise we just skip that part. This probably isn't a great pattern to use all over the place but for diagnostic information it can be helpful.

modules/core/shared/src/main/scala/gem/util/Format.scala
tpolecat: `foldMap` on `Option` ends up being really useful.

modules/core/shared/src/main/scala/gem/util/Format.scala
cquiroz: When would `B` be not a concrete type?

tpolecat: In a generic context like in `foo` in the example above.

```scala
def foo[A,B](f: Format[A,B], a: A) = f.unsafeGet(a) // no tag available here
```

cquiroz: I see, but wouldn't it need to be a concrete type when called? or is the `TypeTag` decided at compilation time?

tpolecat: Remember that implicit arguments are resolved statically at the call site. `foo` isn't asking for `(implicit ev: TypeTag[B])` so the only way to compile it is to pass `null`.

cquiroz: So without the default `null` param it wouldn't compile?

tpolecat: Correct. It would say it can't find an implicit `TypeTag[B]`.

tpolecat:
```
@ import reflect.runtime.universe.TypeTag 
import reflect.runtime.universe.TypeTag

@ implicitly[TypeTag[Int]] // ok 
res1: TypeTag[Int] = TypeTag[Int]

@ def foo[A] = implicitly[TypeTag[A]] // nope 
cmd2.sc:1: No TypeTag available for A
def foo[A] = implicitly[TypeTag[A]] // nope
                       ^
Compilation Failed

@ def foo[A: TypeTag] = implicitly[TypeTag[A]] // ok 
defined function foo
```

tpolecat:
Yeah good idea. I'll add tests for these.

modules/core/shared/src/main/scala/gem/util/Format.scala
swalker2m: That's pretty cool, thanks for the explanation.

modules/core/shared/src/main/scala/gem/util/Format.scala
swalker2m: Nothing is being appended so it's just a shorter version of

```
Option(ev).map(tag => s"${tag.tpe}: ").orEmpty
```
 ?

tpolecat: For `Option` yes. More generally it's `foldRight` with the monoid identity as zero.

tpolecat:
Added a test for `unsafeGet` behavior.

tpolecat:
Ah hell, the tag thing doesn't work with scala-js.

cquiroz:
Right, no reflection on scala.js

tpolecat:
Ok I removed the tag stuff. Oh well.

PR #197, opened 2018-01-17
generate import statements for enum types

tpolecat:
This improves code generation for enumerated data types. Somewhat. Instead of using fully-qualified names we now generate `import` statements.

This looks like a big PR but it's mostly generated code. The only interesting bit is `EnumDef.scala`.


cquiroz:
I kind of liked the types to be explicit on the generation, mostly because we had repeated the same class names across several packages. What's the reason for this change?

tpolecat:
The code was getting hard to read because the lines were getting so long. I want the generated code to look the same as if we had written it by hand.

PR #198, opened 2018-01-18
Rename arbitrary classes files

cquiroz:
This is a small change to match the file names of arbitraries to the class names thus avoiding more filenames like `Angle.scala`, `Coordinates.scala`, etc


tpolecat:
Thanks. We need to do the same with `XxxMeta` and `XxxParsers` I guess. It's interesting that namespacing isn't all that useful these days because we do everything by search rather than by navigation.

cquiroz:
Right, the usual operation with the lack of reliable scala tools is to search by file rather than by class

PR #199, opened 2018-01-19
Add core-fastopt.js size check

cquiroz:
This PR is a bit experimental but should add the size checks after travis run

It will measure the size of the js and how much it has changed since the last code in master. In this particular case the start is 0

Note that to make this run on PRs (which is where we want it) I need to expose a Github token with minimal permissions. This a trade off i'm ok but I'm open to discuss it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ script:
 
 after_script:
   - ls modules/core/js/target/scala-2.12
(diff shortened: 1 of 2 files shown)


tpolecat:
Can the weigh-in script be in the gem repo so we don't have to curl it over from yours?

cquiroz:
Interned the script in the repo

PR #200, opened 2018-01-19
wavelength prism

tpolecat:
After all the bikeshedding about how to do this stuff I finally have a PR with a single prism. Behold!


modules/core/shared/src/main/scala/gem/enum/package.scala
```diff
@@ -14,7 +15,7 @@ import cats.implicits._
 /**
  * Enumerated types (normally generated from database tables) and related syntactic enrichments.
  */
-package object enum {
+package object enum extends ToPrismOps {
```
tpolecat: This makes `Prism.unsafeGet` available to the generated enums. Alternatively we could have all the enum companions extend this trait. I don't have strong feelings.

modules/json/src/main/scala/gem/json.scala
```diff
@@ -43,6 +45,13 @@ package object json {
       moveLeft(b, 3)
   }
 
+  private implicit class PrismOps[A: Encoder: Decoder, B](p: Prism[A, B]) {
+    def toCodec: (Encoder[B], Decoder[B]) = (
+      Encoder[A].contramap(p.reverseGet),
+      Decoder[A].map(p.unsafeGet(_))
+    )
+  }
+
```
tpolecat: When we finally commit to a strategy re: Json we can organize this stuff properly.

tpolecat:
Ugh incremental compilation is not doing very well.

PR #202, opened 2018-01-19
wavelength prism

tpolecat:
Ok one more time. This adds a prism from integral angstroms into `Wavelength`.


tpolecat:
I have no idea what's going on. This compiles clean for me. My local repo may be messed up.

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -5,46 +5,23 @@ package gem.math
...
  * which means the largest representable wavelength is 214.7483647 mm.
  * @param toAngstroms This wavelength in integral angstroms (10^-10 of a meter).
  */
-final class Wavelength private (val toAngstroms: Int) {
-
-  // Sanity checks … should be correct via the companion constructor.
+sealed abstract case class Wavelength private (toAngstroms: Int) {
+  // Sanity check … should be correct via the companion constructor.
   assert(toAngstroms >= 0, s"Invariant violated. $toAngstroms is negative.")
-  assert(toAngstroms <= Int.MaxValue, s"Invariant violated. $toAngstroms is larger than Int.MaxValue.")
```
swalker2m: I thought that was a pretty odd check.

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -54,4 +31,11 @@ object Wavelength {
   implicit val WavelengthOrd: Order[Wavelength] =
     Order.by(_.toAngstroms)
 
+  /**
+   * Prism from Int into Wavelength and back.
+   * @group Optics
+   */
+  def fromAngstroms: Prism[Int, Wavelength] =
```
jdnavarro: Is the `from...` prefix in the name conventional for monocle prisms? It sounds odd to me since they are bidirectional.

tpolecat: My reasoning was that the prism really should be on `Int` because you're "narrowing" it to `Wavelength`, but since it's on `Wavelength` I added `from-` to indicate that the sense is reversed. I'm not super happy with it. Open to suggestions.

tpolecat: I mean, we *could* use syntax to add it to `Int`, so you would say `Int.angstroms.getOption(42)`, it just seems like it might be hard to discover. I'm new to this so I don't really have a feel for it.

jdnavarro: With Haskell `lens` it would be `_Ansgtroms`, but I've never seen that in monocle. There must be some stock prisms coming with monocle, how are they named them?

tpolecat: I don't know if there's a convention. We could do `_optic` if we wanted to.

tpolecat: In any case I'm going to merge this and we can go back and rename things if we come up with a scheme we like better.

jdnavarro: From the monocle examples I've seem it'd be named either just `angstroms` or `angstromsToInt` which isn't great. In any case I guess the `from-` is better than any other naming scheme I could think of.

PR #205, opened 2018-01-23
Order UserTarget

swalker2m:
This PR addresses a comment from @tpolecat in #194:

> If we can define `Order[Target]` we can use `TreeSet` in the `TargetEnvironment` which is a bit more sane. It would also give us a deterministic ordering which might be good. 

To define an `Order[Target]` requires making a few choices.  I'm not sure whether I came down on the right side in every case.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
+++ b/modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
@@ -6,19 +6,21 @@ package gem
 import cats.Eq
 import monocle.macros.Lenses
 
+import scala.collection.immutable.TreeSet
+
 /** Collection of targets associated with an observation.
   */
 @Lenses final case class TargetEnvironment(
   /* asterism, */
   /* guide stars, */
-  userTargets: Set[UserTarget]
+  userTargets: TreeSet[UserTarget]
 )
 
 @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
 object TargetEnvironment {
 
   val empty: TargetEnvironment =
-    TargetEnvironment(Set.empty)
+    TargetEnvironment(TreeSet.empty)
 
   implicit val EqTargetEnvironment: Eq[TargetEnvironment] =
     Eq.fromUniversalEquals
--- a/modules/core/shared/src/main/scala/gem/math/Offset.scala
+++ b/modules/core/shared/src/main/scala/gem/math/Offset.scala
@@ -4,7 +4,7 @@
 package gem
 package math
 
-import cats.{ Eq, Show }
+import cats.{ Order, Show }
 import cats.kernel.CommutativeGroup
 import cats.implicits._
 
@@ -42,9 +42,9 @@ object Offset {
   implicit val ShowOffset: Show[Offset] =
     Show.fromToString
 
-  /** Offsets are equal if their components are pairwise equal. */
-  implicit val EqualOffset: Eq[Offset] =
-    Eq.by(o => (o.p, o.q))
+  /** Offsets are ordered by p, then q. */
+  implicit val OrderOffset: Order[Offset] =
+    Order.by(o => (o.p, o.q))
 
   /** P component of an angular offset.. */
   final case class P(toAngle: Angle) {
@@ -79,9 +79,9 @@ object Offset {
     implicit val ShowP: Show[P] =
       Show.fromToString
 
-    /** P components are equal if their angles are equal. */
-    implicit val EqualP: Eq[P] =
-      Eq.by(_.toAngle)
+    /** P components are by signed microarcseconds. */
+    implicit val OrderP: Order[P] =
+      Order.by(_.toAngle.toSignedMicroarcseconds)
 
   }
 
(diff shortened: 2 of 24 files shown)


modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
```diff
@@ -170,9 +170,21 @@ object ProperMotion {
...
+      Order.whenEqualMonoid[ProperMotion]
+
+    def order[A: Order](f: ProperMotion => A): Order[ProperMotion] =
+      Order.by(f)
+
+    order(_.baseCoordinates)  |+|
+      order(_.epoch)          |+|
+      order(_.properVelocity) |+|
+      order(_.radialVelocity) |+|
+      order(_.parallax)(catsKernelStdOrderForOption(Angle.SignedAngleOrder))
```
swalker2m: @tpolecat This is the pattern I was looking for yesterday on the chatter.   It seems okay to me for a case class with many fields.

modules/core/shared/src/main/scala/gem/syntax/TreeSet.scala
```diff
@@ -0,0 +1,21 @@
...
+
+package gem.syntax
+
+import scala.collection.immutable.TreeSet
+
+final class TreeSetCompanionOps(val self: TreeSet.type) extends AnyVal {
+
+  /** Creates a `TreeSet` from a `List`, provided an `Ordering` is available. */
+  def fromList[A: Ordering](lst: List[A]): TreeSet[A] =
+    TreeSet(lst: _*)
```
swalker2m: This may not be worth the small gain but I find the `: _*` syntax awkward and didn't want to replicate it everywhere.

modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
```diff
@@ -6,19 +6,21 @@ package gem
 import cats.Eq
 import monocle.macros.Lenses
 
+import scala.collection.immutable.TreeSet
```
cquiroz: I think `TreeSet` isn't present in scala.js. let's see if the test fails

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
cquiroz: Why's parallax different?

cquiroz:
Looks good in general. One question, is there a performance impact on making a tuple inside `Order.by`?

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
swalker2m: Parallax is just an `Angle` and I didn't want to define an implicit `Order` for `Angle` in its companion, though I guess I could define one in this local scope and make the implementation a bit nicer.

The reason to not define a standard `Order[Angle]` was that it isn't always clear whether you want that ordered from [180, 180) or [0, 360).  Or that was my thinking at any rate.

swalker2m:
> Looks good in general. One question, is there a performance impact on making a tuple inside `Order.by`?

That was my concern.  If you look at, for example, `catsKernelStdOrderForTuple5[A0, A1, A2, A3, A4]`, it implements the compare by doing each comparison pairwise for the two tuple instances and placing the results in an `Array`:

```scala
def compare(x: (A0, A1, A2, A3, A4), y: (A0, A1, A2, A3, A4)): Int =
Array(A0.compare(x._1, y._1), A1.compare(x._2, y._2), A2.compare(x._3, y._3), A3.compare(x._4, y._4), A4.compare(x._5, y._5)).find(_ != 0).getOrElse(0)
```

The monoid approach, if I understand it correctly, will compare them pairwise only until it finds one that isn't the same.  Since `ProperMotion` is a 5-tuple and some of the things it contains are themselves ordered by converting to tuples, etc., I decided to do it that way instead.  Maybe it doesn't really matter though.

modules/core/shared/src/main/scala/gem/math/Offset.scala
```diff
@@ -79,9 +79,9 @@ object Offset {
     implicit val ShowP: Show[P] =
       Show.fromToString
 
-    /** P components are equal if their angles are equal. */
-    implicit val EqualP: Eq[P] =
-      Eq.by(_.toAngle)
+    /** P components are by signed angle. */
+    implicit val OrderP: Order[P] =
+      Order.by(_.toAngle.toSignedMicroarcseconds)
```
tpolecat: Any reason why these aren't `Angle.SignedAngleOrder.contramap(_.toAngle)`?

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
tpolecat: Can you add a comment explaining why you're not using `Order.by` here? To avoid boxing I assume?

modules/core/shared/src/main/scala/gem/syntax/TreeSet.scala
tpolecat: 👍

modules/core/shared/src/main/scala/gem/math/ProperMotion.scala
swalker2m: I put the reason in a response to Carlos about how `Order` is defined for tuples, but I can add it to the code if you think it is valuable.  Maybe it's not a good reason though?  I'm happy to switch to `Order.by`.

tpolecat: I think it might be useful to have in the code since we're not doing it this way elsewhere.

modules/core/shared/src/main/scala/gem/math/Offset.scala
swalker2m: Didn't occur to me, but now that you mention it I like that better!

tpolecat: I just didn't want to define the same ordering twice.

PR #206, opened 2018-01-24
UI module

cquiroz:
This is a minimal (truly just a println) "UI" module. It basically sets the infrastructure so we can track the size as we add features. The PR contains:

* A new module compiled with scala.js that just does a println
* A update to the weigh script to track this new module's UI js
* Copied from ocs3 a task to upload the js to amazon s3 once merged. This will let us have a history of changes. Maybe useful in the future

This is just infrastructure. The idea is to start adding references to core classes from the "UI" and see the effects on the size as we go along

Feel free to discuss if this approach makes sense to you


.travis.yml
```diff
@@ -21,13 +21,13 @@ install:
   - nvm install stable
   - nvm use stable
   - npm install jsdom
-
+  - pip install --user awscli
 script:
-  - sbt headerCheck test:headerCheck scalastyle sql/flywayMigrate compile test coreJS/test coreJS/fastOptJS
+  - sbt headerCheck test:headerCheck scalastyle sql/flywayMigrate compile test ui/fastOptJS ui/fullOptJS
```
swalker2m: We won't run the test cases in JS now?

build.sbt
```diff
@@ -186,7 +196,7 @@ lazy val flywaySettings = Seq(
 lazy val gem = project
   .in(file("."))
   .settings(commonSettings)
-  .aggregate(coreJVM, db, json, ocs2, ephemeris, service, telnetd, ctl, web, sql, main)
+  .aggregate(coreJVM, coreJS, db, json, ocs2, ephemeris, service, telnetd, ctl, web, sql, main, ui)
```
swalker2m: Size aside, how much does this add to compile time?

build/upload.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# upload merged
+if [ -z "${AWS_ACCESS_KEY_ID}" ]; then
```
swalker2m: Is this an environment variable?

cquiroz: yes, it's only available when you merge though to keep it secret

.travis.yml
cquiroz: they should run implicitly now with the `aggregate` change

build.sbt
cquiroz: Travis shows maybe 1m more though it varies, in some travis runs is even faster but that's likely due to the way travis work. This number will increase though as we evolve the ui especially `fullOptJS`

tpolecat:
Can you explain what we're doing with s3 here?

tpolecat:
This seems fine to me but I need convincing re: uploading to s3 if there's no reason to do so.

cquiroz:
it seems we could get the same result without uploading the files but using a script to track the sizes instead, so I'm removing the S3 upload but keep the ui module

PR #208, opened 2018-01-24
optics for angle, ra, dec

### LINKED ISSUE #201: Spurious failure on CoordinatesSpec
i hit this error on travis

```
info] - Formats.HmsDms.format.normalize *** FAILED ***
[info]   NumberFormatException was thrown during property evaluation.
[info]     Message: For input string: "໒"
[info]     Occurred when passed generated values (
[info]       arg0 = "໒     00:00:00.000000"
[info]     )
[info] - Formats.HmsDms
```

tpolecat:
This continues the work introducing optics, in particular fusing `parse`/`format` pairs together.

Instead of having an `Optics` module, which is kind of a pain, I am instead slicing off a `FooOptics` companion mix-ins, which make it easier to find the optics associated with a given type. I tried it with everything directly inside the companions and it was getting cluttered.

Resolves #201


modules/core/shared/src/main/scala/gem/math/Declination.scala
```diff
@@ -16,7 +18,7 @@ import gem.syntax.parser._
  * [0 - 90] in terms of the underlying `Angle`. Note that the range is *inclusive* of both poles.
  * @see The helpful [[https://en.wikipedia.org/wiki/Declination Wikipedia]] article.
  */
-sealed abstract case class Declination private (toAngle: Angle) {
+sealed abstract case class Declination protected (toAngle: Angle) {
```
swalker2m: Is this privacy downgrade required to make the constructor available in `DeclinationOptics`?  I would have thought that perhaps the `this: Declination.type =>` would give it access even if private.

tpolecat: Yes, that's why. I'm kind of surprised it worked actually. In any case it's redundant since the trait is sealed.

tpolecat:
```
[info] - Formats.HmsDms.format.normalize *** FAILED ***
[info]   NumberFormatException was thrown during property evaluation.
[info]     Message: For input string: "８"
[info]     Occurred when passed generated values (
[info]       arg0 = "00:00:00.000000 ８"
[info]     )
```

doh

cquiroz:
Isn't that error the same as #201

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -245,6 +227,21 @@ object Angle {
 
 }
 
+trait AngleOptics { this: Angle.type =>
```
cquiroz: Why do you require it to be mixed with `Angle`? Couldn't we mix the optics anywhere?

tpolecat: It's intended as a mix-in but I guess in cases where you don't need access to the companion it could be unconstrained.

tpolecat:
Ok this incorporates a fix in atto that causes parsing to fail correctly (without an exception) when international chars are encountered. It also works around JS floating point slop in one of the Coordinates tests.

build.sbt
```diff
@@ -1,12 +1,12 @@
 import com.typesafe.sbt.packager.docker._
 
-lazy val circeVersion        = "0.9.0"
-lazy val attoVersion         = "0.6.1"
+lazy val circeVersion        = "0.9.1"
+lazy val attoVersion         = "0.6.2-M1"
 lazy val catsEffectVersion   = "0.8"
 lazy val catsVersion         = "1.0.1"
-lazy val declineVersion      = "0.4.0"
+lazy val declineVersion      = "0.4.1"
```
cquiroz: What's decline?

cquiroz:
The fix looks fine to me

build.sbt
tpolecat: It's an applicative command line parsing library that I'm using for the deployment stuff. It handles parsing commands options and printing help messages. The telnet server also uses it. A benefit beyond parsing is that every option is self-documenting so the help messages never get out of date.

PR #209, opened 2018-01-25
Print a test program on the ui

cquiroz:
This PR adds a test program (thanks @swalker2m) and just prints it on the "UI"
The purpose of this PR is to have a baseline of how big the js will be


cquiroz:
It's good to take a look at the weigh in of the files. Just printing a program to console makes a 4.1M unoptimized js file and a 1.3M optimized js file

I did a check on the file and counted the function definitions per "package" to get an idea where size is distributed. Note that this is just counting the functions, it doesn't tell anything about how big the functions are

```
atto: 239
cats: 646
gem: 1240
java: 964
zonedb: 843
locales_cldr: 214
scalalib: 4020

total: 8166
```

PR #211, opened 2018-01-25
remove Dataset.Label.Optics module

tpolecat:
This removes the nested module in `Dataset.Label` to make it consistent with other code.


tpolecat:
Ah hell.

tpolecat:
Aggregating the JS stuff in the top level project makes the build so slow that I go down into core and miss things that break elsewhere.

cquiroz:
We used to not have `coreJS` on the aggregate but now we have `ui` which will bring `coreJS` anyway

PR #212, opened 2018-01-26
EphemerisKey Formats

tpolecat:
This adds `Format` instances for `EphemerisKey`.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/arb/ArbEphemerisKey.scala
+++ b/modules/core/shared/src/test/scala/gem/arb/ArbEphemerisKey.scala
@@ -5,11 +5,14 @@ package gem
 package arb
 
 import EphemerisKey._
+import gem.enum.EphemerisKeyType
 
 import org.scalacheck._
 import org.scalacheck.Arbitrary._
 
 trait ArbEphemerisKey {
+  import ArbEnumerated._
+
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/test/scala/gem/arb/ArbEphemerisKey.scala
```diff
@@ -28,7 +31,26 @@ trait ArbEphemerisKey {
     }
 
   implicit val CogenEphemerisKey: Cogen[EphemerisKey] =
-    Cogen[String].contramap(_.format)
+    Cogen[String].contramap(EphemerisKey.fromString.reverseGet)
+
+  private val perturbations: List[String => Gen[String]] =
+    List(
+      s => arbitrary[String],             // swap for a random string
+      s => Gen.const(s.replace("2", "0")) // create a leading zero, perhaps
```
cquiroz: Why this?

tpolecat: For testing we want to be sure to generate ephemeris keys that parse but *don't* format the same. So `"MajorBody_0123"` parses to `MajorBody(123)` which formats as `"MajorBody_123"`. By randomly replacing `2` with `0` we generate enough leading zeroes to test this. Otherwise the test fails because one of the laws checks that we're generating good enough test data.

PR #214, opened 2018-01-26
GNIRS static configuration

jdnavarro:
I could only identify *well depth* as static. The rest of the parameters seem to be dynamic.

I already started working with the dynamic configuration but since it seems it's going to take me a while I'm sending the static changes first.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/config/Arbitraries.scala
+++ b/modules/core/shared/src/test/scala/gem/config/Arbitraries.scala
@@ -54,7 +54,6 @@ trait Arbitraries {
 
   implicit val arbAcqCamStatic    = const(StaticConfig.AcqCam()         )
   implicit val arbBhrosStatic     = const(StaticConfig.Bhros()          )
-  implicit val arbGnirsStatic     = const(StaticConfig.Gnirs()          )
   implicit val arbGpiStatic       = const(StaticConfig.Gpi()            )
   implicit val arbGsaoiStatic     = const(StaticConfig.Gsaoi()          )
   implicit val arbMichelleStatic  = const(StaticConfig.Michelle()       )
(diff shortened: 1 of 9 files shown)


modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -116,21 +116,19 @@ object StaticConfigDao {
...
     object F2 {
 
-      import Instrument.Flamingos2
-
       def select(sid: Int): Query0[StaticConfig.F2] =
         sql"""
           SELECT mos_preimaging
             FROM static_f2
-           WHERE static_id = $sid AND instrument = ${Flamingos2: Instrument}
+           WHERE static_id = $sid
```
cquiroz: Why this change?

jdnavarro: While writing the same code for GNIRS we realized there is already postgres constraint for the instrument (`is_f2`), so the extra `AND` is not needed.

modules/core/shared/src/test/scala/gem/config/Arbitraries.scala
```diff
@@ -122,6 +121,15 @@ trait Arbitraries {
       } yield StaticConfig.GmosSouth(c, s)
     )
 
+  implicit val arbGnirsWellDepth: Arbitrary[GnirsWellDepth] =
+    Arbitrary(
+      Gen.oneOf(GnirsWellDepth.Shallow,
+                GnirsWellDepth.Deep)
+    )
```
swalker2m: There is an `Enumerated` instance for `GnirsWellDepth` so I don't think this needs to be explicitly defined if `ArbEnumerated.arbEnumerated` is in scope.

PR #215, opened 2018-01-26
Section

tpolecat:
A **section** arises when you have a pair of arrows `f: A => B` and `s: B => A` such that `f compose s = id`. This means you can go `B => A => B` without losing information, but not necessarily the other way. We can see this as a normalizing isomorphism, where `f` might not be injective; i.e., it might squeeze a bunch of `A`s into the same `B`, but once you're "in" you can go back and forth. Consider `f: Int => Byte = _.toByte` and `s: Byte => Int = _.toInt`. Anyway in this situation we say that `s` is a **section** for `f`.

The encoding here looks like `Iso` where `f = get` and `s = reverseGet`. This is useful for things like `Angle.fromMicroarcseconds` ↔︎ `a.toMicroarcseconds` which normalizes its values to `[0°, 360°)`.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/util/Section.scala
+++ b/modules/core/shared/src/main/scala/gem/util/Section.scala
@@ -1,0 +1,83 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.util
+
+import cats.arrow.Category
+import monocle.Iso
+
+/**
+ * A normalizing optic, isomorphic to Iso but with different laws, specifically `get`
+ * need not be injective; i.e., distinct inputs may have the same `get` result, which combined
+ * with a subsequent `reverseGet` yield a normalized form for A.
+ */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
+final case class Section[A, B](get: A => B, reverseGet: B => A) {
+
+  /** Compose with another Section. */
+  def composeSection[C](f: Section[B, C]): Section[A, C] =
+    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
+
+  /** Compose with an Iso. */
+  def composeIso[C](f: Iso[B, C]): Section[A, C] =
+    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
+
+  /** Alias to composeSection. */
+  def ^<-![C](f: Section[B, C]): Section[A, C] =
+    composeSection(f)
+
+  /** Alias to composeIso. */
+  def ^<->[C](f: Iso[B, C]): Section[A, C] =
+    composeIso(f)
+
+  /** Section is an invariant functor over A. */
+  def imapA[C](f: C => B, g: B => C): Section[A, C] =
+    Section(get andThen g, f andThen reverseGet)
+
+  /** Section is an invariant functor over B. */
+  def imapB[C](f: A => C, g: C => A): Section[C, B] =
+    Section(g andThen get, reverseGet andThen f)
+
+  /**
+   * get and reverseGet, yielding a normalized formatted value. Subsequent get/reverseGet cycles are
+   * idempotent.
+   */
+  def normalize(b: A): A =
+    reverseGet(get(b))
+
+  /** If we can reverseGet a Product as a String we can implement a tagged toString like "Foo(stuff)". */
+  def productToString(b: B)(
+    implicit as: A =:= String,
+             bp: B <:< Product
+  ): String =
+    taggedToString(b.productPrefix, b)
+
+  /**
+   * If we provide a tag like "Foo" and reverseGet as a String we can implement a nice toString like
+   * "Foo(stuff)".
+   */
+  def taggedToString(tag: String, b: B)(
+    implicit as: A =:= String
+  ): String =
+    new StringBuilder(tag)
+      .append('(')
+      .append(as(reverseGet(b)))
+      .append(')')
+      .toString
+
+}
+
+object Section {
+
+  /** An Iso is trivially a Section. */
+  def fromIso[A, B](p: Iso[A, B]): Section[A, B] =
+    Section(p.get, p.reverseGet)
+
+  /** Section forms a category. */
+  implicit def SectionCategory: Category[Section] =
+    new Category[Section] {
+      def id[A] = Section(identity, identity)
+      def compose[A, B, C](f: Section[B, C], g: Section[A, B]): Section[A, C] = g ^<-! f
+    }
+
+}
--- a/modules/core/shared/src/main/scala/gem/util/Section.scala
+++ b/modules/core/shared/src/main/scala/gem/util/Section.scala
@@ -16,11 +16,11 @@ final case class Section[A, B](get: A => B, reverseGet: B => A) {
 
   /** Compose with another Section. */
   def composeSection[C](f: Section[B, C]): Section[A, C] =
-    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
+    Section(get andThen f.get, reverseGet compose f.reverseGet)
 
   /** Compose with an Iso. */
   def composeIso[C](f: Iso[B, C]): Section[A, C] =
-    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/gem/util/Section.scala
```diff
@@ -0,0 +1,83 @@
...
+ * A normalizing optic, isomorphic to Iso but with different laws, specifically `get`
+ * need not be injective; i.e., distinct inputs may have the same `get` result, which combined
+ * with a subsequent `reverseGet` yield a normalized form for A.
+ */
+@SuppressWarnings(Array("org.wartremover.warts.Null"))
+final case class Section[A, B](get: A => B, reverseGet: B => A) {
+
+  /** Compose with another Section. */
+  def composeSection[C](f: Section[B, C]): Section[A, C] =
+    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
```
sraaphorst: Am I missing something here? Couldn't the new `get` parameter be written as:
`f.get compose get`?

sraaphorst: (Or maybe this breaks a law due to the fact that the function may not be injective?)

sraaphorst:
I hadn't heard of a `Section` before, so this was new to me and informative.
I'll let someone actively working on gem approve it :-).

jdnavarro:
As a consumer of a `Section`s the first thing I'd look would be where in the optics hierarchy it would fit or what can I compose it with and what is the resulting type when doing so?

I'm guessing *iso* + *section* -> *section* and *prism* + *section* -> *section*, but it seems there'll be more.

modules/core/shared/src/main/scala/gem/util/Section.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  /** Compose with an Iso. */
+  def composeIso[C](f: Iso[B, C]): Section[A, C] =
+    Section(a => f.get(get(a)), reverseGet compose f.reverseGet)
+
+  /** Alias to composeSection. */
+  def ^<-![C](f: Section[B, C]): Section[A, C] =
```
cquiroz: are you using the same operators as in `Monocle`?

modules/core/shared/src/main/scala/gem/util/Section.scala
```diff
@@ -0,0 +1,83 @@
...
+    taggedToString(b.productPrefix, b)
+
+  /**
+   * If we provide a tag like "Foo" and reverseGet as a String we can implement a nice toString like
+   * "Foo(stuff)".
+   */
+  def taggedToString(tag: String, b: B)(
+    implicit as: A =:= String
+  ): String =
+    new StringBuilder(tag)
```
cquiroz: Why a `StringBuilder` instead of just string interpolation?

modules/core/shared/src/main/scala/gem/util/Section.scala
```diff
@@ -0,0 +1,83 @@
...
+}
+
+object Section {
+
+  /** An Iso is trivially a Section. */
+  def fromIso[A, B](p: Iso[A, B]): Section[A, B] =
+    Section(p.get, p.reverseGet)
+
+  /** Section forms a category. */
+  implicit def SectionCategory: Category[Section] =
```
cquiroz: Is `Category` from `cats`? What do we get with the `Category` typeclass, compose?

modules/core/shared/src/test/scala/gem/laws/SectionLaws.scala
```diff
@@ -0,0 +1,28 @@
...
+package gem.laws
+
+import cats.Eq
+import cats.implicits._
+import gem.util.Section
+
+final case class SectionLaws[A, B](fab: Section[A, B]) {
+
+  def normalize(a: A): IsEq[B] =
+    fab.get(fab.normalize(a)) <-> fab.get(a)
```
cquiroz: What's `<->`? From cats?

modules/core/shared/src/test/scala/gem/laws/discipline/SectionTests.scala
```diff
@@ -0,0 +1,43 @@
...
+package gem.laws
+package discipline
+
+import cats.Eq
+import gem.util.Section
+import org.scalacheck.{ Arbitrary, Gen }
+import org.scalacheck.Prop._
+import org.typelevel.discipline.Laws
+
+trait SectionTests[A, B] extends Laws {
```
cquiroz: I really like that we can define our own discipline checks

modules/core/shared/src/test/scala/gem/util/SectionSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+import gem.laws.discipline._
+
+final class SectionSpec extends CatsSuite {
+
+  // Our example Section injects Int into Byte
+  val example: Section[Int, Byte] =
+    Section(_.toByte, _.toInt)
+
+  // Ensure it's lawful
+  checkAll("Sections.HmsDms", SectionTests(example).section)
```
cquiroz: Why `HmsDms` is a tested with `Section(_.toByte, _.toInt)` Will `HmsDms` be converted to an int?

modules/core/shared/src/test/scala/gem/laws/SectionLaws.scala
swalker2m: Rob wrote it, but I wonder if it is better than using the existing `<==>` from `IsEqOps`?  Or maybe that doesn't apply for some reason?

```
package gem

package object laws {

  type IsEq[A] = cats.kernel.laws.IsEq[A]
  val IsEq: cats.kernel.laws.IsEq.type = cats.kernel.laws.IsEq

  implicit final class IsEqArrow[A](val lhs: A) extends AnyVal {
    def <->(rhs: A): IsEq[A] = IsEq(lhs, rhs)
  }
  
}
```

modules/core/shared/src/test/scala/gem/util/SectionSpec.scala
swalker2m: I'm guessing that's just a mistaken label.

modules/core/shared/src/main/scala/gem/util/Section.scala
swalker2m: It needs a new operator since `Section` doesn't exist in Monocle, but it definitely is consistent with those in Monocle.

modules/core/shared/src/main/scala/gem/util/Section.scala
tpolecat: Yes, thanks.

tpolecat:
> As a consumer of a Sections the first thing I'd look would be where in the optics hierarchy it would fit or what can I compose it with and what is the resulting type when doing so?

Yeah. I have not tried to build it out but I suspect there are parallel hierarchies for both section-like `reverseGet` and retraction-like `get`, and of course the composition matrix is N^2 so there are a lot of cases to consider. I think `Format` and `Section` make sense but there's definitely more here. Trying to stay out of the weeds.

modules/core/shared/src/main/scala/gem/util/Section.scala
tpolecat: I could be talked out of these, I just kind of wanted to make it consistent with monocle.

modules/core/shared/src/main/scala/gem/util/Section.scala
tpolecat: At the time it seemed important for it to be efficient since we're providing it as a "service" for other classes to use. But yeah it kind of sticks out. I don't feel strongly.

modules/core/shared/src/main/scala/gem/util/Section.scala
tpolecat: Yeah you get `<<<` and `>>>` for `compose` and `andThen`, respectively, and you get a `Monoid` and `MonoidK` for cases where `A` and `B` are the same. It's not that useful but no harm in defining it.

modules/core/shared/src/test/scala/gem/laws/SectionLaws.scala
tpolecat: I defined this so our law-checking stuff would have exactly the same structure used by cats, which defines this alias and syntax the same way. There are so many arbitrary choices in testing setups that I thought we should just copy what cats is doing and trust that the decisions taken there were sensible.

PR #216, opened 2018-01-31
TreeMap.fromList

swalker2m:
Simply adds a `fromList` method to `TreeMap` to avoid the `: _*` syntax.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Program.scala
+++ b/modules/core/shared/src/main/scala/gem/Program.scala
@@ -6,6 +6,8 @@ package gem
 import cats.{ Applicative, Eval, Traverse }
 import cats.implicits._
 
+import gem.syntax.treemapcompanion._
+
 import scala.collection.immutable.TreeMap
 
 /**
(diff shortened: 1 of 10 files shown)


modules/core/shared/src/main/scala/gem/Program.scala
```diff
@@ -6,6 +6,8 @@ package gem
 import cats.{ Applicative, Eval, Traverse }
 import cats.implicits._
 
+import gem.syntax.treemapcompanion._
```
cquiroz: Could it be folded into `gem.syntax._`?

swalker2m: Yes, I could have also just imported `gem.syntax.all._`.  I just did the one since it was all I needed.

modules/core/shared/src/main/scala/gem/syntax/TreeMap.scala
```diff
@@ -0,0 +1,30 @@
...
+  def fromFoldable[F[_], A, B](fab: F[(A, B)])(implicit F: Foldable[F], A: Ordering[A]): TreeMap[A, B] =
+    fromList(F.toList(fab))
+}
+
+trait ToTreeMapCompanionOps {
+  implicit def ToTreeMapCompanionOps(c: TreeMap.type): TreeMapCompanionOps =
+    new TreeMapCompanionOps(c)
+}
+
+object treemapcompanion extends ToTreeMapCompanionOps
```
tpolecat: In cats and scalaz this would be called `treemap` and if we had a bedazzlement trait `ToTreeMapOps` we would add it here. Not a big deal but I think it would be ok to fold X and companion X into the same hunk of syntax. Same with `TreeSet` now that I look.

swalker2m: Thanks, I am in fact adding a `ToTreeMapOps` and I renamed the object `treemap` there.

PR #217, opened 2018-02-01
Add targets type parameter to Observation

swalker2m:
While working on adding asterisms and seeing the `TargetEnvironment` grow more complex, I realized we probably should allow loading `Observation`s without their targets as @tpolecat suggested.  This PR adds a `T` type parameter for targets along the lines of the parameter for the static and dynamic configuration.

One question I have is whether there is a better way to deal with the three `Observation` functors that result from this.  See `Observation.targetsFunctor` etc.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Observation.scala
+++ b/modules/core/shared/src/main/scala/gem/Observation.scala
@@ -3,27 +3,32 @@
 
 package gem
 
+import gem.config.{ StaticConfig, DynamicConfig }
 import gem.syntax.string._
 
 import mouse.boolean._
 
-import cats.{ Applicative, Eval, Order, Show, Bitraverse }
+import cats.{ Functor, Order, Show }
 import cats.implicits._
 
 /**
- * An observation, parameterized over the types of its static config and steps (typically
- * [[gem.config.StaticConfig StaticConfig]] and [[gem.Step Step[α]]], respectively, for a
- * fully-specified Observation; or `Unit` and `Nothing` for a
- * minimally-specified Observation.
+ * An observation, parameterized over the types of its targets, static config
+ * and steps (typically [[gem.TargetEnvironment]], [[gem.config.StaticConfig StaticConfig]]
+ * and [[gem.Step Step[α]]], respectively, for a fully-specified Observation;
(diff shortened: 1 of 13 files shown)


modules/core/shared/src/main/scala/gem/Observation.scala
```diff
@@ -89,15 +94,30 @@ object Observation {
...
-      def bifoldLeft[A, B, C](fab: Observation[A,B], c: C)(f: (C, A) => C, g: (C, B) => C): C =
-        fab.steps.foldLeft(f(c, fab.staticConfig))(g)
-      def bifoldRight[A, B, C](fab: Observation[A,B], c: Eval[C])(f: (A, Eval[C]) => Eval[C], g: (B, Eval[C]) => Eval[C]): Eval[C] =
-        fab.steps.foldRight(f(fab.staticConfig, c))(g)
-      def bitraverse[G[_]: Applicative, A, B, C, D](fab: Observation[A, B])(f: A => G[C], g: B => G[D]): G[Observation[C,D]] =
-        (f(fab.staticConfig), fab.steps.traverse(g)).mapN((c, d) => fab.copy(staticConfig = c, steps = d))
+  /** A functor over `Observation` on the `T`, or targets, type parameter.
+    * Not implicit.
+    */
+  def targetsFunctor[S, D]: Functor[Observation[?, S, D]] =
```
tpolecat: I think this approach is sensible, but can we strengthen these to `Traverse`? I can do a followup PR if you like.

modules/core/shared/src/main/scala/gem/Observation.scala
```diff
@@ -89,15 +94,30 @@ object Observation {
 
   }
 
-  /** Observation is a bitraversable functor. */
```
cquiroz: I guess there is no such thing as a `Tritraverse`?

tpolecat: It's possible to abstract over `NTraverse` but I don't think it's worth the effort. As long as none of the instances is implicit I think we're fine.

swalker2m:
D'oh, this got put into the commit message: Thanks, I'll go ahead and merge this now so I can continue but we can come back and strengthen `Functor` to `Traverse` in a followup.

PR #219, opened 2018-02-02
Asterism

swalker2m:
This is a request for comments PR.  It adds the "asterism" concept from OCS2 but I think it probably needs more work.  In particular, asterisms are at least sometimes associated with particular instruments (e.g., `GhostDualTarget`) and so the observation's static config, steps, and asterism should all agree on the instrument.  This version will guarantee that at the database-consistency level but not in the Scala model.

Currently all OCS2 programs use the `SingleTarget` asterism so to make it more realistic I added `Ghost` to the `Instrument` enum and added a trivial `GhostDualTarget` asterism as well.  That gives us a couple of different asterisms to work with and requires sorting out instrument-dependencies.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Asterism.scala
+++ b/modules/core/shared/src/main/scala/gem/Asterism.scala
@@ -1,0 +1,73 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+import cats.Eq
+import cats.data.NonEmptyList
+
+import gem.enum.{ AsterismType, Instrument }
+
+import monocle.Prism
+import monocle.macros.{ Lenses, GenPrism }
+
+
+sealed trait Asterism extends Product with Serializable {
+
+  def asterismType: AsterismType
+
+  def instrument: Instrument
+
+  def targets: NonEmptyList[Target]
+
+}
+
+object Asterism extends AsterismOptics {
+
+  @Lenses final case class SingleTarget(target: Target, instrument: Instrument) extends Asterism {
+
+    override def asterismType: AsterismType =
+      AsterismType.SingleTarget
+
+    override def targets: NonEmptyList[Target] =
+      NonEmptyList.one(target)
+
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  object SingleTarget {
+    implicit val EqSingleTarget: Eq[SingleTarget] =
+      Eq.fromUniversalEquals
+  }
+
+  @Lenses final case class GhostDualTarget(ifu1: Target, ifu2: Target) extends Asterism {
+
+    override def asterismType: AsterismType =
+      AsterismType.GhostDualTarget
+
+    override def instrument: Instrument =
+      Instrument.Ghost
+
+    override def targets: NonEmptyList[Target] =
+      NonEmptyList.of(ifu1, ifu2)
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  object GhostDualTarget {
+    implicit val EqGhostDualTarget: Eq[GhostDualTarget] =
+      Eq.fromUniversalEquals
+  }
+
+  implicit val EqAsterism: Eq[Asterism] =
+    Eq.fromUniversalEquals
+}
+
+trait AsterismOptics { self: Asterism.type =>
+
+  val singleTarget: Prism[Asterism, Asterism.SingleTarget] =
+    GenPrism[Asterism, Asterism.SingleTarget]
+
+  val ghostDualTarget: Prism[Asterism, Asterism.GhostDualTarget] =
+    GenPrism[Asterism, Asterism.GhostDualTarget]
+
+}
--- a/modules/core/shared/src/main/scala/gem/Asterism.scala
+++ b/modules/core/shared/src/main/scala/gem/Asterism.scala
@@ -5,17 +5,13 @@ package gem
 
 import cats.Eq
 import cats.data.NonEmptyList
-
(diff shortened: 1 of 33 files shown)


modules/core/shared/src/main/scala/gem/Asterism.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import gem.enum.{ AsterismType, Instrument }
+
+import monocle.Prism
+import monocle.macros.{ Lenses, GenPrism }
+
+
+sealed trait Asterism extends Product with Serializable {
+
+  def asterismType: AsterismType
```
tpolecat: Is the `AsterismType` used outside of serialization code? In general it would be nice to avoid propagating the discriminant into the model. I wonder if it would make sense to indicate these in the Schema somehow, to differentiate them from other enums.

modules/core/shared/src/main/scala/gem/Asterism.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  def instrument: Instrument
```
tpolecat: As you hinted, we may want this to be indexed on the instrument; i.e., `Asterism[I <: Instrument with Singleton]` and `val instrument: I`.

modules/core/shared/src/main/scala/gem/syntax/TreeMap.scala
```diff
@@ -20,11 +21,52 @@ final class TreeMapCompanionOps(val self: TreeMap.type) extends AnyVal {
     */
   def fromFoldable[F[_], A, B](fab: F[(A, B)])(implicit F: Foldable[F], A: Ordering[A]): TreeMap[A, B] =
     fromList(F.toList(fab))
+
+  /** Combines all the given maps into a single map, where keys common to two or
+    * more maps are the value of the last occurrence in the list.
+    */
+  def join[A: Ordering, B](ms: List[TreeMap[A, B]]): TreeMap[A, B] =
+    ms.foldLeft(TreeMap.empty[A, B]) { case (m0, m1) => m0 ++ m1 }
```
tpolecat: We need to be a little careful here because this is a cromulent monoid but it's inconsistent with the default one. I think this is fine for now though.

tpolecat:
Ah crap, I clicked the update branch button by mistake,

tpolecat:
This seems good so far. I do think we'll probably end up indexing the asterism on the instrument to ensure that everything matches up.

modules/core/shared/src/main/scala/gem/Asterism.scala
tpolecat: Or `trait Asterism { type I <: Instrument with Singleton ...` and then `Asterism.Aux`.

swalker2m: I've been struggling a bit with this.  Will I need a single-target asterism instance for each instrument that it applies to?

tpolecat: `case class Single[I0 <: Instrument with Singleton](...) extends Asterism { type I = I0 }` .. something like that?

swalker2m:
Rut roh, scala-js strikes again:

```
[info] Fast optimizing /Users/swalker/dev/gem/modules/core/js/target/scala-2.12/gem-core-test-fastopt.js
[error] Referring to non-existent method scala.Predef$.valueOf(java.lang.Object)java.lang.Object
[error]   called from gem.arb.ArbAsterism.genAsterism(gem.enum.Instrument)org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism$.genAsterism(gem.enum.Instrument)org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism.$$anonfun$arbAsterism$2(gem.enum.Instrument)org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism$.$$anonfun$arbAsterism$2(gem.enum.Instrument)org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism.$$anonfun$arbAsterism$1()org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism$.$$anonfun$arbAsterism$1()org.scalacheck.Gen
[error]   called from gem.arb.ArbAsterism.$$init$()scala.Unit
[error]   called from gem.arb.ArbAsterism$.<init>()
[error]   called from gem.arb.ArbTargetEnvironment.$$init$()scala.Unit
[error]   called from gem.arb.ArbTargetEnvironment$.<init>()
[error]   called from gem.TargetEnvironmentSpec.<init>()
[error]   called from gem.TargetEnvironmentSpec.__exportedInits
[error]   exported to JavaScript with @JSExport
[error] involving instantiated classes:
[error]   gem.arb.ArbAsterism$
[error]   gem.TargetEnvironmentSpec
```

tpolecat:
Well that's weird. I will investigate.

cquiroz:
That's well known, TLS lib doesn't work properly on scala.js

cquiroz:
https://github.com/typelevel/scala/issues/164

tpolecat:
Bah

tpolecat:
I'm starting to hate scala-js.

I can change this to use `shapeless.Witness` instead of `valueOf`

swalker2m:
Other than being admittedly much less cool, what are the disadvantages of

```
def genTargetEnvironment[I <: Instrument with Singleton](i: I): Gen[TargetEnvironment] =
```

as opposed to

```
def genTargetEnvironment[I <: Instrument with Singleton: ValueOf]: Gen[TargetEnvironment] =		 
```

tpolecat:
Yeah that might be the way to go.

swalker2m:
I'm not sure where to go from here.  This works but leaves it possible for the asterism, static configuration, and steps to each have a different idea of what the instrument should be.  Of course, it will be prevented at the database level.  Should we keep going in this PR or fix these issues in a separate one?

tpolecat:
Yes if you'e good merging this PR I can attempt the next step.

PR #222, opened 2018-02-10
WIP on Observation.Full

tpolecat:
This adds a type member `I` to `TargetEnvironment` and then stitches all the pieces together in `Observation.Full`, requiring that everything in the observation that cares about the instrument is indexed the same way; i.e., each component's `I` type member has the singleton type of some `Instrument` value.

The way we maintain this is a little roundabout. Basically the rule is that when you want to construct something that's indexed on a value, instead of doing

```scala
def foo(i: Instrument): Foo[i.type]
```

we need to say

```scala
def foo[I <: Instrument with Singleton](i: Instrument.Aux[I]): Foo[I]
```

Where `trait Instrument { type Self = this.type }` and `Aux[I] = Instrument { type Self = I }` and in fact all enum types have this pattern now.  So it's just turning the [existential] singleton type into a universally quantified type. It's just a way to get Scala to understand what's going on.

Code that needs to know that the `Observation.Full` parts line up should be written as

```scala
def foo[I <: Instrument with Singleton](i: Observation.Full.Aux[I]): ...
```

to which you can pass an unadorned `Observation.Full`. See `ObservationSpec` for the kinds of things we can prove.

Anyway via this trick all the non-essential casts from the initial PR are gone. There are still some in `db` but this is necessary because we're loading different hunks and then asserting that they line up. We might want to add some sanity checks in the future.

It remains to be seen whether this approach is good or not. We might end up removing the `I` type members and turning them into type parameters, which will simplify some things but complicate others. I think we should go with this and see how well it works.


modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
```diff
@@ -4,25 +4,66 @@
...
-  userTargets: TreeSet[UserTarget]
-)
+/** Collection of targets associated with an observation. */
+sealed trait TargetEnvironment {
+
+  type I <: Instrument with Singleton
+
+  val asterism: Option[Asterism.Aux[I]]
+
+  /** Guide stars. */
```
swalker2m: Actually, guide stars are still pending.  The user targets are for blind offsetting, etc.

When guide stars are added, they too will have an instrument dependency.

modules/core/shared/src/main/scala/gem/Observation.scala
```diff
@@ -27,8 +26,19 @@ final case class Observation[+T, +S, +D](
 
 object Observation {
 
-  /** A fully specified observation. */
-  type Full = Observation[TargetEnvironment, StaticConfig, Step[DynamicConfig]]
+  /** A fully specified observation, with unknown instrument. */
+  type Full = Full.Aux[i.type] forSome { val i: Instrument }
```
swalker2m: If you have an `Observation.Full` will the compiler be able to prevent you from, say, making the static config and steps disagree about the instrument?  If so, is all of the additional complexity buying us much?

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -70,11 +70,12 @@ object ObservationDao {
   /** Construct a program to select a fully specified observation, with targets,
     * static config and steps.
     */
+  @SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf"))
   def select(id: Observation.Id): ConnectionIO[Observation.Full] =
     for {
       o <- selectConfig(id)
       t <- TargetEnvironmentDao.selectObs(id, o.targets)
-    } yield Observation.targetsFunctor.as(o, t)
+    } yield Observation.targetsFunctor.as(o, t).asInstanceOf[Observation.Full] // TODO: push this assertion lower
```
swalker2m: Seems like we'll usually be working with an `Observation.Full` with unknown instrument.  I guess I don't *know* that will be the case but how could we know a priori what the instrument will be.

modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
tpolecat: This comment was left over from the original case class. I'll remove it.

modules/core/shared/src/main/scala/gem/Observation.scala
tpolecat: Yes. That's what this type means.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
tpolecat: In the comprehension above `o` knows the instrument so we could push this down into `selectObs` as an expectation that we check, and if it lines up we can perform the cast down there, such that `o` and `t` line up. Maybe.

Anyway in the same way we assert the types of database rows and blow up (programmer error!) if we're wrong, we need to check that the parts line up and then perform the cast so *Scala* will know. And we want to do this as quickly as possible so the largest amount of code knows it.

cquiroz:
I'm not sure I follow all the details here. At any rate, before merging could you check how much bigger the client becomes? This may bring `shapeless` along and want to make sure if the effect on js size isn't too large

tpolecat:
How do I check the client size?

cquiroz:
Once travis completes successfully it will show the client size on the checks on github. Let's see how it goes, `shapeless` isn't that big after all, but the advanced use of scala features may introduce some problems

modules/core/shared/src/main/scala/gem/config/StaticConfig.scala
```diff
@@ -15,12 +15,14 @@ import monocle.macros.Lenses
  */
 sealed abstract class StaticConfig {
   type I <: Instrument with Singleton
-  def instrument: I
+  val instrument: Instrument.Aux[I]
```
swalker2m: Seems like that should remain a `def` if it isn't defined.

tpolecat: Yeah I'll switch that back. An earlier attempt needed a stable reference.

tpolecat:
Ok I think this is probably as good as I can get it for now. I think we need to see how it works out. I updated the description with info on the trick I had to use.

modules/core/shared/src/main/scala/gem/Observation.scala
```diff
@@ -27,8 +26,13 @@ final case class Observation[+T, +S, +D](
 
 object Observation {
 
-  /** A fully specified observation. */
-  type Full = Observation[TargetEnvironment, StaticConfig, Step[DynamicConfig]]
+  /** A fully specified observation, with unknown instrument. */
+  type Full = Full.Aux[I] forSome { type I <: Instrument with Singleton }
```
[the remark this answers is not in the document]
tpolecat: Yes, I think it achieves what we want.

If we try to change an unadorned `Observation.Full` we find that we can't. The resulting type is no longer an `Observation.Full`.

```scala
def edit(o: Observation.Full): Observation.Full =
  o.copy(staticConfig = StaticConfig.Gpi())

error: type mismatch;
 found   : gem.Observation[gem.TargetEnvironment.Aux[I(in type Full)],gem.config.StaticConfig.Gpi,gem.Step[gem.config.DynamicConfig.Aux[(some other)I(in type Full)]]]
    (which expands to)  gem.Observation[gem.TargetEnvironment{type +I = I(in type Full)},gem.config.StaticConfig.Gpi,gem.Step[gem.config.DynamicConfig{type +I = (some other)I(in type Full)}]]
 required: gem.Observation.Full
    (which expands to)  gem.Observation[gem.TargetEnvironment{type I = I},gem.config.StaticConfig{type I = I},gem.Step[gem.config.DynamicConfig{type I = I}]]( forSome { type I <: gem.enum.Instrument with Singleton })
```

This *does* work if we want to muck around with a narrow-typed observation.
```scala
type GpiObservation = Observation.Full.Aux[Instrument.Gpi.type]

def edit(o: GpiObservation): GpiObservation =
  o.copy(staticConfig = StaticConfig.Gpi()) // static config is the right type
```

This also works if we want to edit a generic observation.

```scala
def edit2[I <: Instrument with Singleton](
  o: Observation.Full.Aux[I],
  s: StaticConfig.Aux[I]
): Observation.Full.Aux[I] = 
  o.copy(staticConfig = s)
```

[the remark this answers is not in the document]
swalker2m: Is a plain `Observation.Full` useful for anything?  Should it even exist?  Sorry to be so dense, it's just unfamiliar. 

`ObservationDao.select` returns an `Observation.Full`.  You can't know a priori what the instrument is so that makes sense.  But then how to you narrow it down to a specific instrument in order to work with it?  I suppose a match on the instrument type, somehow, must be involved but I doubt that's correct.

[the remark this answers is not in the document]
tpolecat: Yeah that's a good point. I'll add a narrowing operation.

PR #223, opened 2018-02-12
Add GNIRS dynamic configuration

jdnavarro:
This also includes the Step DAO.

I was already working on the GNIRS ocs2 decoder with the intention of including it in this same PR but I'm realizing it's going to take me a while. I placed a provisional stub to make this PR compile.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
@@ -82,6 +82,7 @@ object DynamicConfig {
 
     type GmosNorthDefinition = GmosDefinition[GmosNorthDisperser, GmosNorthFilter, GmosNorthFpu]
     type GmosSouthDefinition = GmosDefinition[GmosSouthDisperser, GmosSouthFilter, GmosSouthFpu]
+
   }
 
   sealed abstract class Impl[I0 <: Instrument with Singleton](val instrument: Instrument.Aux[I0]) extends DynamicConfig {
@@ -91,7 +92,6 @@ object DynamicConfig {
   /** @group Constructors */ final case class AcqCam()   extends DynamicConfig.Impl(Instrument.AcqCam)
   /** @group Constructors */ final case class Bhros()    extends DynamicConfig.Impl(Instrument.Bhros)
   /** @group Constructors */ final case class Ghost()    extends DynamicConfig.Impl(Instrument.Ghost)
-  /** @group Constructors */ final case class Gnirs()    extends DynamicConfig.Impl(Instrument.Gnirs)
   /** @group Constructors */ final case class Gpi()      extends DynamicConfig.Impl(Instrument.Gpi)
   /** @group Constructors */ final case class Gsaoi()    extends DynamicConfig.Impl(Instrument.Gsaoi)
   /** @group Constructors */ final case class Michelle() extends DynamicConfig.Impl(Instrument.Michelle)
@@ -184,6 +184,32 @@ object DynamicConfig {
   object GmosSouth {
     val Default: GmosSouth =
       GmosSouth(GmosCommonDynamicConfig.Default, None, None, None)
+  }
+
+  final case class Gnirs(
+    camera:               GnirsCamera,
+    decker:               GnirsDecker,
+    disperser:            GnirsDisperser,
+    disperserOrder:       GnirsDisperserOrder,
+    exposureTime:         Duration,
+    filter:               GnirsFilter,
+    fpu:                  Either[GnirsFpuOther, GnirsFpuSlit],
+    prism:                GnirsPrism,
+    readMode:             GnirsReadMode
+  ) extends DynamicConfig.Impl(Instrument.Gnirs)
+
+  object Gnirs {
+    val Default: Gnirs = Gnirs(
+      GnirsCamera.ShortBlue,
+      GnirsDecker.Acquisition,
+      GnirsDisperser.D32,
+      GnirsDisperserOrder.Three,
+      java.time.Duration.ofSeconds(17),
+      GnirsFilter.Order5,
+      Right(GnirsFpuSlit.LongSlit_0_30),
+      GnirsPrism.Mirror,
+      GnirsReadMode.Bright
+    )
   }
 
 }
--- a/modules/db/src/main/scala/gem/dao/StepDao.scala
+++ b/modules/db/src/main/scala/gem/dao/StepDao.scala
@@ -83,7 +83,7 @@ object StepDao {
         case Instrument.Ghost      => pure(DynamicConfig.Ghost())
         case Instrument.GmosN      => Gmos.selectOneNorth(oid, loc).maybe.widen[DynamicConfig]
         case Instrument.GmosS      => Gmos.selectOneSouth(oid, loc).maybe.widen[DynamicConfig]
-        case Instrument.Gnirs      => pure(DynamicConfig.Gnirs())
+        case Instrument.Gnirs      => Gnirs.selectOne(oid, loc)    .maybe.widen[DynamicConfig]
         case Instrument.Gpi        => pure(DynamicConfig.Gpi())
         case Instrument.Gsaoi      => pure(DynamicConfig.Gsaoi())
         case Instrument.Michelle   => pure(DynamicConfig.Michelle())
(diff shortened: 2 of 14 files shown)


modules/ocs2/src/main/scala/gem/ocs2/SequenceDecoder.scala
```diff
@@ -71,7 +71,9 @@ object SequenceDecoder extends PioDecoder[List[Step[DynamicConfig]]] {
       case Instrument.GmosN      => Gmos.parseNorth(cm)
       case Instrument.GmosS      => Gmos.parseSouth(cm)
 
-      case Instrument.Gnirs      => DynamicConfig.Gnirs()   .asRight
+      /// TODO: Implement GNIRS decoder
+      case Instrument.Gnirs      => DynamicConfig.Gnirs.Default.asRight
```
jdnavarro: I'm already working on implementing this properly for the next PR.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -82,6 +82,12 @@ object DynamicConfig {
 
     type GmosNorthDefinition = GmosDefinition[GmosNorthDisperser, GmosNorthFilter, GmosNorthFpu]
     type GmosSouthDefinition = GmosDefinition[GmosSouthDisperser, GmosSouthFilter, GmosSouthFpu]
+
+    final case class Gnirs(
+      disperser: Option[GnirsDisperser],
+      filter:    GnirsDisperser,
```
swalker2m: Oops: `filter:    GnirsDisperser`

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -82,6 +82,12 @@ object DynamicConfig {
...
+      fpu:       Option[GnirsFpuSlit]
+    )
```
swalker2m: This is for Smart Gcal correct?  Does the same case class serve both for searching and for defining entries (as for F2)?  If so you'll need to extend `SmartGcalSearchKey with SmartGcalDefinitionKey`.  Or if you're not ready to do Smart Gcal maybe just leave this undefined for now.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -186,4 +191,34 @@ object DynamicConfig {
...
+
+  object Gnirs {
+    val Default: Gnirs = Gnirs(
+      GnirsCamera.ShortBlue,
+      GnirsDecker.Acquisition,
+      GnirsDisperser.D32,
+      GnirsDisperserOrder.Three,
+      java.time.Duration.ZERO,
+      GnirsFilter.Order5,
+      GnirsFpu.Slit(GnirsFpuSlit.LongSlit4),
```
swalker2m: Perhaps we should pick a name like LongSlit_0_30 that corresponds, somewhat, to the display name?  Or do they naturally think of the FPU as LongSlit4 instead of 0.30 arcsec?

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -186,4 +191,34 @@ object DynamicConfig {
...
+    wavelengthSuggestion: GnirsWavelengthSuggestion
+  ) extends DynamicConfig.Impl(Instrument.Gnirs)
+
+  object Gnirs {
+    val Default: Gnirs = Gnirs(
+      GnirsCamera.ShortBlue,
+      GnirsDecker.Acquisition,
+      GnirsDisperser.D32,
+      GnirsDisperserOrder.Three,
+      java.time.Duration.ZERO,
```
swalker2m: The default exposure time for this configuration according to the OT is 17 seconds.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -186,4 +191,34 @@ object DynamicConfig {
...
+      GnirsFilter.Order5,
+      GnirsFpu.Slit(GnirsFpuSlit.LongSlit4),
+      GnirsPrism.Mirror,
+      GnirsReadMode.Bright,
+      GnirsWavelengthSuggestion.One
```
swalker2m: Is this a field of GNIRS?  I would guess that the actual observing wavelength is what we want to store and that the "suggestions" are for the UI?

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -542,5 +542,118 @@ object StepDao {
...
+            disperser,
+            disperserOrder,
+            exposureTime,
+            filter,
+            fpuBuilder.toFpu,
+            prism,
+            readMode,
+            wavelengthSuggestion
+          )
+      }
```
swalker2m: I'm guessing (maybe wrong) that we could define a composite for the FPU and then this `GnirsBuilder` wouldn't be necessary.  Did you run into problems with it?

jdnavarro: No, it was just straightforward enough with the builder instead of the composite.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
jdnavarro: I wasn't thinking about Smart Gcal yet. I think in GNIRS, as opposed to F2, the set of possible FPU parameters is fixed. There are 2 groups of FPU parameters, Slits and Other (for acquisition, or pupil), which if I understood they are mutually exclusive, thus the sum type.

From what you say I think I might need to rework it for Smart Gcal. I'll have a look at it now.

jdnavarro: Oh, I see now what you mean. I had the other Gnirs case class defined below in mind when making my previous comment. I'll comment this case class out.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
jdnavarro: In SQL for `short_name` and `long_name` the 0.30 arsec name is used. LongSlit4 is only used for the Id, and therefore in code. GMOS and F2 follow the same naming convention. 

If we want to change this convention I could open a ticket take care of renaming the slits for all instruments later.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
jdnavarro: `WavelengthSuggestion` only shows up because of renaming it from `WaveLengthSuggestion`.

I can remember when doing the enumerations that I had the exact same inclination of not considering this parameter since it sounds like UI code only, but the fact is that this suggestion is used in different instances (through getters and setters) in code that is not related to UI.

It's still not clear to me how is it being used in the current OCS, but let me investigate it further. I should, at least, write a comment of why the suggestion is an enumeration.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
swalker2m: When the science staff writes a query to find observations that use that FPU, it would probably be better to match on the id LongSlit_0_30 instead of LongSlit4 I'm guessing.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
jdnavarro: I can't figure out now how this parameter is is being used in OCS, not even in the UI. I'll remove it in gem. We can always add it back if the need arises.

jdnavarro:
This is ready for another review.

modules/core/shared/src/main/scala/gem/config/GnirsConfig.scala
```diff
@@ -0,0 +1,42 @@
...
+        case _         => None
+      }
+
+    /** Extracts the FPU Other, if applicable. */
+    def toFpuOther: Option[GnirsFpuOther] =
+      this match {
+        case Other(fpu) => Some(fpu)
+        case _          => None
+      }
+
```
swalker2m: A `fold` might come in handy?

```scala
def fold[A](slit: GnirsFpuSlit => A, other: GnirsFpuOther => A): A =
  this match {
    case Slit(fpu)  => slit(fpu)
    case Other(fpu) => other(fpu)
  }

def toFpuSlit: Option[GnirsFpuSlit] =
  fold(Option(_), None)

  ...
```

Or thinking about it a bit more, is there any reason this isn't just a type alias?

```scala
type GnirsFpu = Either[GnirsFpuOther, GnirsFpuSlit]
```

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -542,5 +542,112 @@ object StepDao {
...
+            decker,
+            disperser,
+            disperserOrder,
+            exposureTime,
+            filter,
+            fpuBuilder.toFpu,
+            prism,
+            readMode
+          )
+      }
```
swalker2m: I think you can get rid of the builder altogether, though I'm not sure about the FPU part.  Maybe if it is just a type alias for `Either`.

modules/core/shared/src/main/scala/gem/config/GnirsConfig.scala
jdnavarro: In a review a while back I was told that, in Gemini code, the `Left` case of an `Either` was usually associated with an error or fallback connotation, so since then, I was always creating distinct sum types when there is no associated connotation for the `Left` case.

Speaking for myself, I never assume any connotation for `Left` anyway, only that, for whatever reason, there is slight bias towards `Right`, so I'm fine with making this an `Either`.

swalker2m: Hmm, if anybody holds the opinion that `Either` is only for errors we should discuss it.  There may be good reasons for that, but if so I'm ignorant of them.  At any rate, when I had to make a similar call for GMOS FPU I went with `Either[GmosCustomMask, GmosNorthFpu]`.

jdnavarro:
I replaced the with `GnirsFPU` with th `Either`, but I'm still not sure how to get rid of `GnirsFPUBuilder`.

PR #224, opened 2018-02-13
Generate license headers through the JVM only

### LINKED ISSUE #220: Sometimes genEnums generates bogus code
Sometimes when running `genEnums` extra bogus code snippets get inserted. See the compile error below as an example but the generated snippets are not always the same.

Running `genEnums` multiple times fixes it.

It sounds like an upstream error (sbt?) but I'm reporting it here to keep track of it.

```
[error] /Users/jnavarro/Develop/gem/modules/core/shared/src/main/scala/gem/enum/GcalLampType.scala:45:12: expected start of definition
[error]   implicit val GcalLampTypeEnumerated: Enumerated[GcalLampType] =
[error]            ^
[error] /Users/jnavarro/Develop/gem/modules/core/shared/src/main/scala/gem/enum/GcalLampType.scala:45:12: expected start of definition
[error]   implicit val GcalLampTypeEnumerated: Enumerated[GcalLampType] =
[error]            ^
[error] /Users/jnavarro/Develop/gem/modules/core/shared/src/main/scala/gem/enum/GcalLampType.scala:51:1: Unmatched closing brace '}' ignored here
[error] }
[error] ^
[error] /Users/jnavarro/Develop/gem/modules/core/shared/src/main/scala/gem/enum/GcalLampType.scala:51:1: Unmatched closing brace '}' ignored here
[error] }
[error] ^
[error] two errors found
[error] two errors found
```

jdnavarro:
Fixes #220.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ onLoad in Global := { s =>
 cancelable in Global := true
 
 // some extra commands for us
-addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum; headerCreate")
+addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum")
 addCommandAlias("schemaSpy", "sql/runMain org.schemaspy.Main -t pgsql -port 5432 -db gem -o modules/sql/target/schemaspy -u postgres -host localhost -s public")
 addCommandAlias("gemctl", "ctl/runMain gem.ctl.main")//
 
@@ -201,7 +201,6 @@ lazy val gem = project
 lazy val core = crossProject
   .crossType(CrossType.Full)
   .in(file("modules/core"))
-  .enablePlugins(AutomateHeaderPlugin)
   .settings(commonSettings)
   .settings(
     libraryDependencies ++= Seq(
@@ -226,7 +225,7 @@ lazy val core = crossProject
   )
 
 
-lazy val coreJVM = core.jvm
+lazy val coreJVM = core.jvm.enablePlugins(AutomateHeaderPlugin)
 lazy val coreJS = core.js
 
 lazy val db = project


cquiroz:
could it be better to fix it by removing the plugin from `coreJS`. That should work on all cases rather than only on the custom command

swalker2m:
> could it be better to fix it by removing the plugin from coreJS. That should work on all cases rather than only on the custom command

Agree, that would be ideal.

swalker2m:
> Oddly I have never had a problem with this but if it fixes your issue 

Don't you usually skip the scala.js build?  If so, that may be why.

tpolecat:
Yeah I spend most of my time in `coreJVM` and `db` but I didn't think that would affect a command since they're defined globally. I don't understand sbt.

jdnavarro:
I still have this issue with the latest commit I pushed. After trying several ways of disabling `automateHeaderPlugin` for JS, I've only seen this problem go away consistently in my original first commit, when I place `coreJVM/automateHead` in the `genEnums` alias.

I can keep trying different combinations of disabling the plugin if you still don't like that solution.

swalker2m:
In addition to the current version of this PR, you might also remove the header generation after the enum generation. Headers will be added by the compile step.  Could you give that a try?

So, also update the line:

```
addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum; headerCreate")
```

to

```
addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum")
```

swalker2m:
In my experience, the problem occurs when the set of enums are divided into two groups at the time headers are generated.  One or more of the enums appears in both groups, and those are the ones that end up with corrupt source files.  Is that the case for you when you experience the problem too?  If so, I assume the command is somehow still being run simultaneously by the normal build and the JS build.

The biggest issue will be with the enums because it is a large set of source files that need the headers all at once.  If the problem does happen with a normal source file though it will be annoying because you won't be able to just regenerate it.

cquiroz:
This could be tested by removing the headers and run `coreJS/compile` and verify the headers are created

swalker2m:
> This could be tested by removing the headers and run `coreJS/compile` and verify the headers are created

In the current version of this PR, my theory is that it is happening as a result of the `genEnums` alias.

cquiroz:
If adding headers is implicit with compile we shouldn't include it in the alias. I think

jdnavarro:
>  One or more of the enums appears in both groups, and those are the ones that end up with corrupt source files. Is that the case for you when you experience the problem too?

I've seen all kinds of corrupt enumerations, but I think this is the case for me too.

jdnavarro:
If I remove the header generation command from the alias, no headers get generated in the enumerations. I also tried different ways of enabling the plugin below too, but no luck.

cquiroz:
i have never seen this happening in ocs3 with crossprojects. I suppose because we don't generate code

jdnavarro:
> This could be tested by removing the headers and run coreJS/compile and verify the headers are created.

`coreJS/headersCreate` also fixes it but I keep seeing the 2 groups of enumerations being generated.

What it's fixing it with just 1 group is just disabling the plugin for the core project, and using `headersCreate` in the `genEmum` alias, as it was before.

build.sbt
```diff
@@ -36,7 +36,7 @@ onLoad in Global := { s =>
 cancelable in Global := true
 
 // some extra commands for us
-addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum; headerCreate")
```
swalker2m: I think we can leave the compile off and just let that be a separate step.  The headers will be added then just as for everything else.

tpolecat: That's fine with me.

PR #226, opened 2018-02-13
Target Id

swalker2m:
There are a number of `Int` database ids that pop up in the codebase.  Working with guide stars, I came across the need for a guide target id, a guide group id, and of course the target id.  I think the time has come to have types for the various ids beyond the primitive `Int`s.  I'm not sure if this is the way we want to do it though, so I offer `Target.Id` for discussion first.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Target.scala
+++ b/modules/core/shared/src/main/scala/gem/Target.scala
@@ -16,6 +16,15 @@ import monocle.macros.Lenses
 @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
 object Target {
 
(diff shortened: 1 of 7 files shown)


cquiroz:
Looks good to me. It seems fairly straightforward

cquiroz:
Your PR increased the js size by 1 byte 😉

swalker2m:
> I kind of want to see what happens with SIP-35 before committing to a strategy for these little wrappers.

I'm going to add wrappers for the others, but we can go back and fix them if we want to take another approach.

tpolecat:
Sounds good.

PR #228, opened 2018-02-14
Turn on -Ywarn-unused:patvars

tpolecat:
This should have been on … I don't remember why I turned it off. Anyway @som-snytt advises that it will prevent the `Position.point on NoPosition` issue https://github.com/scala/bug/issues/10448 (!)


som-snytt:
I think I noticed that side-effect last weekend. It will also warn on unused patvars, for which you can use `case C(x @ _, y @ _) =>` syntax to say the name is only for documentation.

PR #230, opened 2018-02-16
observation narrowing

tpolecat:
This provides a way to narrow the type of an `Observation.Full` by case analysis and an example of its use. Attempts to address @swalker2m comments in #222


tpolecat:
I can't find an obvious improvement here. Things that enable exhaustiveness checking for the case analysis in `foo` also break the dependency so `o`'s type becomes unknown again.

PR #231, opened 2018-02-21
non-injective optics

tpolecat:
`Iso[A,B]` has the property that round-trips from either side are identities; i.e., that the elements of `A` and `B` are in a 1:1 relationship. It we weaken this and allow the round-trip to be merely *idempotent* in one direction or the other we get a "normalizing" 1:N or N:1 relationship between types of different sizes. An`Angle ⟺ HourAngle` conversion that rounds angles to the nearest 15 microarcseconds is an example of such a relationship.

So this PR adds a `gem.optics` package, moves `Format` and `Section` (renamed to `SplitEpi`), and adds `SplitMono` and `Wedge`. See the `README.md` below for a slightly longer explanation.


cquiroz:
It is unclear to me what's the problem this is solving, can you give me an example?

modules/core/shared/src/main/scala/gem/optics/README.md
```diff
@@ -0,0 +1,32 @@
...
+                                           │
+    SplitMono[A,B]     SplitEpi[A,B]  ─────┤
+        A < B             A > B            │
+
+          │                 │         Prism[A,B]
+          └────────┬────────┘       ∃ a ⊂ A | a = B
+                   │                       │
+                                           │
+                Iso[A,B]  ─────────────────┘
+                 A = B
```
sraaphorst: This was all very clearly written and understandable (thank you for making this easily digestible). My only confusion is with the use of the symbols `<` and `>`.

For a monomorphism, isn't it required that `A` map surjectively onto `B` (hence I'm not sure what `A < B` means, as `|A| ≥ |B|` is necessary to be able to be surjective), and for an epimorphism `φ`, isn't it required that for all `x, y` in `A`, `x ≠ y` implies `φ(x) ≠ φ(y)`, i.e. the mapping is injective? (And in that case, `|A| ≤ |B|` is required.)

Furthermore, just a pedantic point of mathematical contention, but an isomorphism (or other mapping) between `A` and `B` doesn't really mean that `A = B`: it means that there is an invertible bijective map `φ` such that `φ(A) = B`.

tpolecat: I may misunderstand what you're saying … monomorphisms generalize injective functions, epimorphisms generalize surjective functions. I'm following the definitions [here](https://ncatlab.org/nlab/show/split+monomorphism) where `get` is the mono/epi and `reverseGet` is its retraction/section. 

And yeah, the '<', '>', '=' thing is just an intuition.

sraaphorst: Ahhh, you're right. I haven't studied this in years and now I get to suffer the embarrassment of having mixed up epimorphisms and monomorphisms! Apologies for that.

Now that my definitions are not reversed, this makes much more sense.

tpolecat: No worries. It's really easy to get everything backwards because they're duals. I confused myself several time writing this.

tpolecat:
> It is unclear to me what's the problem this is solving, can you give me an example?

@cquiroz this gives us a way to define the relationships between pairs of functions like `fromDegrees: Int => Angle` and `toDegrees: Angle => Int` for example, which are inverses in a sense but not exactly, because the values coming from one direction are normalized (`Int` into `[0,360)` in this case). There is no existing optic that handles this situation. `Iso[Int, Angle]` is too strong and `Prism[Int, Angle]` is too weak. `SplitEpi[Int, Angle]` fits in between. The dual, `SplitMono[Angle, Int]` is equivalent but oriented in the other direction.

tpolecat:
It's probably worth noting that structurally `SpliMono`, `SplitEpi`, and `Wedge` are identical to `Iso`; the only difference is in the laws. Same with `Format` and `Prism`.

PR #232, opened 2018-02-22
Update to scala-java-time 2.0.0-M13

cquiroz:
This PR updates the client side to use `scala-java-time` version 2.0.0-M13 with a custom tzdb. This should produce a significant reduction in the generated `js` size and reduced time to produce those files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ ensime.sbt
 archive
 smartgcal
 .jvmopts
+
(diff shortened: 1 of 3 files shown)


cquiroz:
this leads to 887K less on fastOpt and 552K less on fullOpt 🎉

tpolecat:
I'm updating a bunch of deps and will pick this up too.

PR #233, opened 2018-02-26
Ephemeris Commands

swalker2m:
Provides a few simple commands for working with ephemerides.  It isn't clear which commands, if any, will be needed but this gives us a starting point.  An example session:

```
Welcome, local time is Mon Feb 26 10:45:59 CLST 2018.
root@gem> eph-report -s GS MajorBody_505
Looking up MajorBody(505) @ GS ...

MajorBody(505) @ GS
  Last Update.......: 
  Last Update Check.: 
  Solution Reference: 
  Available Range...: 
       
root@gem> eph-update -s GS MajorBody_505
Updating MajorBody(505) @ GS ...

MajorBody(505) @ GS
  Last Update.......: Mon, 26 Feb 2018 13:46:26 GMT
  Last Update Check.: Mon, 26 Feb 2018 13:46:26 GMT
  Solution Reference: Jun 17, 2016
  Available Range...: Sun, 31 Dec 2017 17:00:00 GMT - Fri, 31 Aug 2018 17:00:00 GMT
       
root@gem> eph-report -s GS MajorBody_505
Looking up MajorBody(505) @ GS ...

MajorBody(505) @ GS
  Last Update.......: Mon, 26 Feb 2018 13:46:26 GMT
  Last Update Check.: Mon, 26 Feb 2018 13:46:26 GMT
  Solution Reference: Jun 17, 2016
  Available Range...: Sun, 31 Dec 2017 17:00:00 GMT - Fri, 31 Aug 2018 17:00:00 GMT
       
root@gem> eph-export -s GS -d /Users/swalker MajorBody_505
Exporting MajorBody(505) @ GS ... /Users/swalker/MajorBody_505.eph
```


modules/ephemeris/src/main/scala/gem/horizons/EphemerisContext.scala
```diff
@@ -0,0 +1,52 @@
...
+/** Collection of information related to an ephemeris.  Used to determine
+  * whether an update is needed.
+  *
+  * @param key  unique horizons id of the object
+  * @param site information valid for the given site
+  * @param meta associated ephemeris metadata, if any
+  * @param rnge earliest and latest ephemeris element times, if any
+  * @param soln current horizons solution reference from JPL, if any
+  */
+final case class EphemerisContext(
```
swalker2m: This was just pulled out of another class to the top level so make it easily available.

modules/ephemeris/src/main/scala/gem/horizons/HorizonsClient.scala
```diff
@@ -32,18 +32,19 @@ import scala.concurrent.duration._
 object HorizonsClient {
 
   /** A global mutex for limiting concurrent connections to Horizons. */
-  private val mutex: Semaphore[IO] =
-    Semaphore[IO](1L)(implicitly, scala.concurrent.ExecutionContext.global).unsafeRunSync // note
+//  private val mutex: Semaphore[IO] =
+//    Semaphore[IO](1L)(implicitly, scala.concurrent.ExecutionContext.global).unsafeRunSync // note
```
swalker2m: It turns out that this causes problems when you try to fold over a stream of multiple queries as we do when downloading a full semester of data.  We'll need another solution here.

modules/ephemeris/src/test/scala/gem/horizons/HorizonsUpdaterApp.scala
```diff
@@ -1,97 +0,0 @@
...
-  *   > ephemeris/test:run gem.horizons.HorizonsUpdater site key ...
-  * }}}
-  *
-  * e.g.,
-  *
-  * {{{
-  *   ephemeris/test:run gem.horizons.HorizonsUpdater GS MajorBody_605 MajorBody_606
-  * }}}
-  */
-object HorizonsUpdaterApp {
```
swalker2m: This was a way to drive horizons updating from the sbt command line, but now we have a telnet command so it isn't needed.

modules/telnetd/src/main/scala/command/eph.scala
```diff
@@ -0,0 +1,159 @@
...
+  /** Argument for one or more horizons keys. */
+  val horizonsKeys: Opts[NonEmptyList[EphemerisKey.Horizons]] =
+    keys("horizons-key", keyStr => s"'$keyStr' is not a valid horizons key") {
+      case k: EphemerisKey.Horizons => k
+    }
+
+  /** Command that exports ephemeris files in TCS format to a directory on the
+    * server.
+    */
+  val exportCommand: GemCommand =
```
swalker2m: Export is kind of lame because it exports to a directory on the server.  Since there is no configuration, you have to supply said directory.  This is how ephemerides are communicated to the TCS though.  In production of course we'll want some kind of cron job that does this on a daily basis.

modules/service/src/main/scala/gem/Service.scala
```diff
@@ -30,20 +34,47 @@ final class Service[M[_]: Monad] private (private val xa: Transactor[M], val log
       UserDao.changePassword(user.id, oldPassword, newPassword).transact(xa)
     }
 
+  object ephemeris {
```
tpolecat: 👍 on the namespacing. Good idea.

modules/telnetd/src/main/scala/command/eph.scala
tpolecat: I remember the old NCSA Telnet program had a feature to insert the ftp command to go back to your local machine, so this problem has been around forever. Maybe it could `scp` the file for you, so you have to give user@machine:path and it would prompt for the password. But this is fine for now.

Issue #234, opened 2018-02-26
Guide Star Question


swalker2m:
I'm looking for a few suggestions for implementing guide stars.  The difficulty is in attempting to guarantee that the guider being used is legitimate.  I've created an enumeration of guiders and their associated instrument (if any):

```sql
CREATE TABLE e_guider (
  id         text       PRIMARY KEY,
  instrument identifier REFERENCES  e_instrument,
  short_name text       NOT NULL,
  long_name  text       NOT NULL,
  UNIQUE (id, instrument) -- FK referent
);

ALTER TABLE e_guider OWNER TO postgres;

COPY e_guider (id, instrument, short_name, long_name) FROM stdin;
F2OI	Flamingos2	F2 OI	Flamingos2 OIWFS
GmosNOI	GmosN	GMOS-N OI	GMOS North OIWFS
GmosSOI	GmosS	GMOS-S OI	GMOS South OIWFS
P1GN	\N	P1 GN	PWFS1 North
P1GS	\N	P1 GS	PWFS1 South
P2GN	\N	P2 GN	PWFS2 North
P2GS	\N	P2 GS	PWFS2 South
\.
```

Some guiders, like PWFS2, can be used with multiple instruments so the instrument field is left `NULL`.  A corresponding ADT is generated from this:

```scala
  /** @group Constructors */ case object F2OI extends Guider("F2OI", Some(Instrument.Flamingos2), "F2 OI", "Flamingos2 OIWFS")
  /** @group Constructors */ case object GmosNOI extends Guider("GmosNOI", Some(Instrument.GmosN), "GMOS-N OI", "GMOS North OIWFS")
  /** @group Constructors */ case object GmosSOI extends Guider("GmosSOI", Some(Instrument.GmosS), "GMOS-S OI", "GMOS South OIWFS")
  /** @group Constructors */ case object P1GN extends Guider("P1GN", None, "P1 GN", "PWFS1 North")
  /** @group Constructors */ case object P1GS extends Guider("P1GS", None, "P1 GS", "PWFS1 South")
  /** @group Constructors */ case object P2GN extends Guider("P2GN", None, "P2 GN", "PWFS2 North")
  /** @group Constructors */ case object P2GS extends Guider("P2GS", None, "P2 GS", "PWFS2 South")
```

[...]

PR #235, opened 2018-02-26
Guiding

swalker2m:
This is a work in progress and doesn't even compile (👎).  Also, I created a `GuideStar` class and DAO but was planning on getting rid of that and folding it into `GuideGroup`.  At any rate, I'm opening this PR to accompany #234


swalker2m:
This was a dead-end unfortunately.

PR #236, opened 2018-02-26
more optics tweaks

tpolecat:
This addresses some things I ran into implementing optics for `Angle`.

- The names of `imapA` and `imapB` got transposed in two places, so that's fixed.
- Every `SplitEpi` induces a `Format` so it makes sense for `SplitEpiTests` to extend `FormatTests` to add a bit more checking. Refactored a bit to support this.
- It turns out the round-trip morphisms for `Wedge` are not idempotent, but rather they reach a fixed point after two round trips. This is because both directions squeeze a big type into a smaller one so you have two rounds of normalization to deal with. Anyway weakened the laws to accommodate this.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/optics/README.md
+++ b/modules/core/shared/src/main/scala/gem/optics/README.md
@@ -6,7 +6,7 @@ If we loosen the requirement that types be the same size we get a different kind
 
 This [phenomenon](https://ncatlab.org/nlab/show/split+epimorphism) is a thing, called a **split monomorphism** or a **split epimorphism** depending on which side is bigger. Note that every `Iso` is trivially a split where the idempotent round-trip happens to be an identity.
 
-When we compose a split mono and a split epi end-to-end in either direction we end up with a situation where neither round-trip is necessarily an identity or idempotent, but will reach a fixed point in two round-trips. I'm calling this a `Wedge` for lack of a better idea. Splits are trivially wedges where one of the round-trips happens to be an identity.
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/gem/optics/README.md
```diff
@@ -6,7 +6,7 @@ If we loosen the requirement that types be the same size we get a different kind
 
 This [phenomenon](https://ncatlab.org/nlab/show/split+epimorphism) is a thing, called a **split monomorphism** or a **split epimorphism** depending on which side is bigger. Note that every `Iso` is trivially a split where the idempotent round-trip happens to be an identity.
 
-When we compose a split mono and a split epi end-to-end in either direction we end up with a situation where neither round-trip is necessarily an identity but both are idempotent. I'm calling this a `Wedge` for lack of a better idea. Splits are trivially wedges where one of the idempotent round-trips happens to be an identity.
+When we compose a split mono and a split epi end-to-end in either direction we end up with a situation where neither round-trip is necessarily an identity or idempotent, but will reach a fixed point in two round-trips. I'm calling this a `Wedge` for lack of a better idea. Splits are trivially wedges where one of the round-trips happens to be an identity.
```
swalker2m: > in two round-trips

Isn't it just one round trip? `A -> B ->  A`.  You might not get the original `A` but from that point on you should? The same for `B -> A -> B`.

tpolecat: Huh, I ran into cases where tests were failing because idempotence was too strong but I can't reproduce it now. Must have imagined it. I'll revert that change.

PR #238, opened 2018-02-26
angle optics

tpolecat:
This unifies the to/from data conversion pairs for `Angle` and `HourAngle` by defining optics that have defined laws we can check. This should help ensure that 2-way conversions remain consistent as the code evolves.

There is a question of whether to include accessors at all, or just force the user to go through optics all the time. Apparently this is a common design issue with no good answer. The compromise I reached here is that `Angle` wraps `Long` microarcseconds and `HourAngle` [logically] wraps `Long` microseconds so these "direct" members have traditional accessors. Unrelated/derived types and units have no accessors so you must use optics for these.

The huge diff for `Angle.scala` is mostly Scaladoc comments to group members by function.


tpolecat:
Doh I forgot to change a bunch of stuff.

PR #239, opened 2018-02-26
angle optics

tpolecat:
Ok, reopening #238 here because I force-pushed.

This unifies the to/from data conversion pairs for `Angle` and `HourAngle` by defining optics that have defined laws we can check. This should help ensure that 2-way conversions remain consistent as the code evolves.

There is a question of whether to include accessors at all, or just force the user to go through optics all the time. Apparently this is a common design issue with no good answer. The compromise I reached here is that `Angle` wraps `Long` microarcseconds and `HourAngle` [logically] wraps `Long` microseconds so these "direct" members have traditional accessors. Unrelated/derived types and units have no accessors so you must use optics for these.

The huge diff for `Angle.scala` is mostly Scaladoc comments to group members by function.


modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -117,68 +115,77 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
...
 object Angle extends AngleOptics {
 
-  val Angle0:   Angle = fromDegrees(0)
-  val Angle90:  Angle = fromDegrees(90)
-  val Angle180: Angle = fromDegrees(180)
-  val Angle270: Angle = fromDegrees(270)
+  /** @group Constants */ lazy val Angle0:   Angle = degrees.reverseGet(0)
+  /** @group Constants */ lazy val Angle90:  Angle = degrees.reverseGet(90)
+  /** @group Constants */ lazy val Angle180: Angle = degrees.reverseGet(180)
+  /** @group Constants */ lazy val Angle270: Angle = degrees.reverseGet(270)
```
swalker2m: What does `lazy` buy us (beyond not paying the cost for creating these constants upfront)?  Every subsequent access requires a lock as I understand it.  `degrees` should be initialized by the time this runs?

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -227,17 +239,91 @@ object Angle extends AngleOptics {
...
 }
 
-trait AngleOptics { this: Angle.type =>
+trait AngleOptics extends OpticsHelpers { this: Angle.type =>
+
+  /**
+   * Microarcseconds, in [0, 360°).
+   * @group Optics
+   */
+  lazy val microarcseconds: SplitMono[Angle, Long] =
```
swalker2m: Same question on `lazy` for all of these.

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -227,17 +239,91 @@ object Angle extends AngleOptics {
...
+    SplitMono(_.toMicroarcseconds, Angle.fromMicroarcseconds)
+
+  /**
+   * Signed microarcseconds, in [-180°, 180°).
+   * @group Optics
+   */
+  lazy val signedMicroarcseconds: SplitMono[Angle, Long] = {
+    lazy val µas360 = Angle.Angle180.toMicroarcseconds * 2L
```
swalker2m: Here I guess something has to be `lazy` since `Angle` extends this trait and hasn't yet defined `Angle180`.

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -227,17 +239,91 @@ object Angle extends AngleOptics {
...
+  lazy val arcminutes: Wedge[Angle, Int] =
+      microarcseconds.scaled(1000L * 1000L * 60L)
 
-  val hourAngle: Prism[Angle, HourAngle] =
-    Prism((_: Angle).toHourAngleExact)(_.toAngle)
+  /**
+   * Degrees, in [0, 360).
+   * @group Optics
+   */
+  lazy val degrees: Wedge[Angle, Int] = microarcseconds.scaled(1000L * 1000L * 60L * 60L)
```
swalker2m: I don't yet have a feel for when you use `Wedge` instead of, I would guess, `SplitMono`.

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -227,17 +239,91 @@ object Angle extends AngleOptics {
...
+   */
+  lazy val hourAngle: SplitEpi[Angle, HourAngle] =
+    SplitEpi(a => HourAngle.microseconds.reverseGet(a.toMicroarcseconds.toLong / 15L), identity)
 
-  val fromStringSignedDMS: Format[String, Angle] =
+  /**
+   * This angle as an HourAngle, where defined.
+   * @group Optics
+   */
+  lazy val hourAngleExact: Prism[Angle, HourAngle] =
```
swalker2m: Here too, why is it a `Prism` instead of `SplitEpi`?

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -376,12 +464,83 @@ object HourAngle extends HourAngleOptics {
...
-  def fromStringHMS: Format[String, HourAngle] =
-    Format(AngleParsers.hms.parseExact, _.toHMS.format)
+  /**
+   * This `HourAngle` as an `Angle`.
+   * @group Optics
+   */
+  lazy val angle: SplitMono[HourAngle, Angle] = Angle.hourAngle.reverse
+
+  /**
+   * This `HourAngle` as an `Angle`.
```
swalker2m: "This `HourAngle` in microseconds."

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -376,12 +464,83 @@ object HourAngle extends HourAngleOptics {
...
+trait OpticsHelpers {
+
+  // Syntax to scale down and squeeze into Int
+  protected implicit class SplitMonoOps[A](self: SplitMono[A, Long]) {
+
+    private val longToInt: SplitEpi[Long, Int] =
+      SplitEpi(_.toInt, _.toLong)
+
+    def scaled(n: Long): Wedge[A, Int] =
+      self.imapB[Long](_ * n, _ / n) composeSplitEpi longToInt
```
swalker2m: Very nice. 👍

swalker2m:
It seems like all the new optics would make a good conference talk.

modules/core/shared/src/main/scala/gem/math/Angle.scala
tpolecat: I was getting some initialization order errors, and I worry that as we accumulate static initializers elsewhere it could get worse, so I decided to be defensive. We can review lazy vals at some point but for now I'd like to leave it.

modules/core/shared/src/main/scala/gem/math/Angle.scala
tpolecat: Going `Angle => Int` we squeeze `µas: Long` into `Int`, but on the way back we squeeze `Int` into `[0, 360)` so there is normalization happening in both directions.

The laws for all these new optics have a "coverage" test that checks to be sure you couldn't be using a stronger optic. For instance I accidentally had `Angle => DMS` as a `SplitEpi` and the test failed because it's actually an `Iso`. So when in doubt you can guess and the tests should tell you whether it's the right optic or not.

modules/core/shared/src/main/scala/gem/math/Angle.scala
tpolecat: There are two here. `hourAngle` is a `SplitEpi` that rounds to the nearest 15 µas, so it's losing information: it provides `Angle => HourAngle`.

`hourAngleExact` only succeeds if the angle is divisible by 15 µas, so it's not defined in cases where information would be lost. It provides `Angle => Option[HourAngle]`.

tpolecat:
> It seems like all the new optics would make a good conference talk.

I'm going to try to do an unconference session at NEScala to see what people think.

tpolecat:
Wheeee "update branch" roulette. Wish me luck! Big money big money!

PR #242, opened 2018-02-27
Add ocs2 GNIRS importer

jdnavarro:
I could successfully import some dummy GNIRS programs I exported from the odb.

Are there any tests for the importers where I could check GNIRS XML files under different circumstances? I'm pretty sure that if we push it we'll find issues while importing.

### DIFF
--- a/modules/ocs2/src/main/scala/gem/ocs2/Legacy.scala
+++ b/modules/ocs2/src/main/scala/gem/ocs2/Legacy.scala
@@ -139,7 +139,16 @@ object Legacy {
     }
 
     object Gnirs {
-      val WellDepth = Key("wellDepth")(Parsers.Gnirs.wellDepth)
+      import Parsers.Gnirs._
+      val Camera         = Key("camera"           )(camera)
+      val Decker         = Key("decker"           )(decker)
+      val Disperser      = Key("disperser"        )(disperser)
+      val Filter         = Key("filter"           )(filter)
+      val Fpu            = Key("slitWidth"        )(fpu)
+      val Prism          = Key("crossDispersed"   )(prism)
+      val ReadMode       = Key("readMode"         )(readMode)
+      val Wavelength     = Key("centralWavelength")(centralWavelength)
+      val WellDepth      = Key("wellDepth"        )(wellDepth)
     }
   }
 
--- a/modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
+++ b/modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
@@ -539,9 +539,137 @@ object Parsers {
   }
 
   object Gnirs {
-    val wellDepth: PioParse[GnirsWellDepth] = enum(
-      "SHALLOW" -> GnirsWellDepth.Shallow,
-      "DEEP"    -> GnirsWellDepth.Deep
-    )
+
+    val camera: PioParse[GnirsCamera] = {
+
+      import GnirsCamera._
+
+      enum(
+        "short blue" -> ShortBlue,
+        "long blue"  -> LongBlue,
+        "short red"  -> ShortRed,
+        "long red"   -> LongRed
+      )
+    }
+
+    val decker: PioParse[GnirsDecker] = {
+
+      import GnirsDecker._
+
+      enum(
+        "acquisition"            -> Acquisition,
+        "pupil viewer"           -> PupilViewer,
+        "short camera long slit" -> ShortCamLongSlit,
+        "short camera x-disp"    -> ShortCamCrossDispersed,
+        "IFU"                    -> Ifu,
+        "long camera long slit"  -> LongCamLongSlit,
+        "wollaston"              -> Wollaston
+      )
+
+    }
+
+    val disperser: PioParse[GnirsDisperser] = {
+
+      import GnirsDisperser._
+
+      enum(
+        "10 l/mm grating"  -> D10,
+        "32 l/mm grating"  -> D32,
+        "111 l/mm grating" -> D111
+      )
+
+    }
+
+    val filter: PioParse[GnirsFilter] = {
+
+      import GnirsFilter._
+
+      enum(
+        "x-dispersed"            -> CrossDispersed,
+        "order 6 (X)"            -> Order6,
+        "order 5 (J)"            -> Order5,
+        "order 4 (H-MK: 1.65um)" -> Order4,
+        "order 3 (K)"            -> Order3,
+        "order 2 (L)"            -> Order2,
+        "order 1 (M)"            -> Order1,
+        "H2: 2.12um"             -> H2,
+        "H + ND100X"             -> HNd100x,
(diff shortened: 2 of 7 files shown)


modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -539,9 +539,144 @@ object Parsers {
...
+    }
+
+    val disperser: PioParse[GnirsDisperser] = {
+
+      import GnirsDisperser._
+
+      enum(
+        "10 l/mm grating"  -> D10,
+        "32 l/mm grating"  -> D32,
+        "112 l/mm grating" -> D111
```
swalker2m: Oops, that looks like it should be `"111 l/mm grating" -> D111`.

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -539,9 +539,144 @@ object Parsers {
...
+      import GnirsDisperserOrder._
+
+      enum(
+        "4.85" -> One,
+        "3.4"  -> Two,
+        "2.2"  -> Three,
+        "1.65" -> FourXD,
+        "1.63" -> Four,
+        "1.25" -> Five,
+        "1.1"  -> Six
```
swalker2m: What happened to `Seven` and `Eight`?

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -539,9 +539,144 @@ object Parsers {
...
+        "0.30 arcsec"  -> Right(LongSlit_0_30),
+        "0.45 arcsec"  -> Right(LongSlit_0_45),
+        "0.675 arcsec" -> Right(LongSlit_0_675),
+        "1.0 arcsec"   -> Right(LongSlit_1_00),
+        "3.0 arcsec"   -> Right(LongSlit_3_00),
+        "IFU"          -> Left(Ifu),
+        "acquisition"  -> Left(Acquisition),
+        "pupil viewer" -> Left(PupilViewer),
+        "pinhole 0.1"  -> Left(Pinhole1),
+        "pinhole 0.3"  -> Left(Pinhole2)
```
swalker2m: `Pinhole3`?

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
jdnavarro: I forgot them because `Seven` and `Eight` are not selectable through in the OT UI but are present in the `OCS`. I'll check if they are ever used anywhere in the `OCS` codebase. Without the UI label I'm not sure under which exact values they will be exported.

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
jdnavarro: The right value is actually `Pinhole3` but that needs to be fixed in the enumerations. I already have a branch with this and other enumerations fixes I identified while writing the import code. I guess I'll identify more enumeration errors while trying to import real GNIRS programs.

I'll send this enumerations fixes PR before trying to get this merged.

modules/ocs2/src/main/scala/gem/ocs2/Legacy.scala
```diff
@@ -139,7 +139,16 @@ object Legacy {
     }
 
     object Gnirs {
-      val WellDepth = Key("wellDepth")(Parsers.Gnirs.wellDepth)
+      import Parsers.Gnirs._
+      val Camera         = Key("camera"           )(camera)
+      val Decker         = Key("decker"           )(decker)
+      val Disperser      = Key("disperser"        )(disperser)
+      val DisperserOrder = Key("centralWavelength")(disperserOrder)
```
swalker2m: There's an issue here, I think.  The central wavelength can be anything that the user types in.  It seems that you can go back from central wavelength to the corresponding disperser order using a function that tries to match them (see `GNIRSParams.Order.getOrder`) and failing that, picks a default value.

 Looking at the sequence XML, it seems that we don't even write the disperser order into the sequence.  We do write the wavelength though.  I think we need to update the GNIRS config to explicitly store the wavelength.  Since you cannot explicitly set the disperser order, we should probably remove that from the config and provide a function to compute it.

jdnavarro: Now it makes sense to me. In the OT UI the central wavelength the UI is a dropbox where you can either select a value or you can type directly the central wavelength, I didn't notice you could override the selectable values until now.

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -539,9 +539,133 @@ object Parsers {
...
+        "Very Bright/Acq./High Bckgrd." -> VeryBright,
+        "Bright Objects"                -> Bright,
+        "Faint Objects"                 -> Faint,
+        "Very Faint Objects"            -> VeryFaint
+      )
+
+    }
+
+    val centralWavelength: PioParse[Wavelength] = {
+      bigDecimal.map(w => Wavelength.fromAngstroms.unsafeGet((w * 1000.0).rounded.toInt))
```
swalker2m: 1 angstrom = 10 nm = 10,000 um so this should be multiplied by 10,000?

modules/ocs2/src/main/scala/gem/ocs2/pio/PioParse.scala
```diff
@@ -69,5 +69,5 @@ object PioParse {
     long.map(Instant.ofEpochMilli)
 
   val seconds: PioParse[Duration] =
-    double.map { d => Duration.ofMillis((d * 1000).round) }
+    double.map { d => Duration.ofMillis((d * 10000).round) }
```
swalker2m: Why the change to 10,000 here?  There are 1000 ms in a second.

modules/ocs2/src/main/scala/gem/ocs2/Parsers.scala
```diff
@@ -539,9 +539,133 @@ object Parsers {
...
+        "Very Bright/Acq./High Bckgrd." -> VeryBright,
+        "Bright Objects"                -> Bright,
+        "Faint Objects"                 -> Faint,
+        "Very Faint Objects"            -> VeryFaint
+      )
+
+    }
+
+    val centralWavelength: PioParse[Wavelength] = {
+      bigDecimal.map(w => Wavelength.fromAngstroms.unsafeGet((w * 10000.0).rounded.toInt))
```
swalker2m: I think using `BigDecimal` from `String` is better than the Gmos version which uses `double`. 👍 Since you have a `BigDecimal` though, you might as well move the decimal right?

`bigDecimal.map(w => Wavelength.fromAngstroms.unsafeGet(w.underlying.movePointRight(4).intValue))`

PR #243, opened 2018-02-28
ODB Import Commands

swalker2m:
This PR would add a couple of telnet commands (`fetch-obs` and `fetch-prog`) that simplify OCS2 import testing.  Namely, they allow the user to import directly from an ODB without first having to explicitly export to an XML file.  For example:

```
Welcome, local time is Wed Feb 28 16:39:51 CLST 2018.
root@gem> fetch-obs --help
Usage: fetch-obs [--host <host>] <obs-id>

Fetch an observation from an OCS2 ODB and store it.

Options and flags:
    --help
        Display this help text.
    --host <host>, -h <host>
        ODB hostname (default localhost)
root@gem> 
root@gem> fetch-obs -h gsodb GS-2018A-Q-101-21
Imported GS-2018A-Q-101-21
root@gem> 
root@gem> fetch-prog -h gsodb GS-2018A-Q-208
Imported GS-2018A-Q-208
```

As you can see, the host defaults to `localhost` to simplify development but the production ODBs are fair game as well.


swalker2m:
> Is there any risk in importing from a production odb?

It should be fine (hehe). It just goes to a servlet that exports to XML using the normal export code and then manipulates the XML to unroll the sequence and make it easier to read.

Issue #241, opened 2018-02-27
Sporadic(?) EphemerisDao Test Error


swalker2m:
```
[info] - EphemerisDao should selectAll *** FAILED ***
[info]   BatchUpdateException was thrown during property evaluation.
[info]     Message: Batch entry 0 
[info]             INSERT INTO ephemeris (key_type,
[info]                                    key,
[info]                                    site,
[info]                                    timestamp,
[info]                                    ra,
[info]                                    dec,
[info]                                    ra_str,
[info]                                    dec_str,
[info]                                    delta_ra,
[info]                                    delta_dec)
[info]                  VALUES ('Comet', 'jxd6mdgafn', 'GS', '1301-02-29 22:13:15.805159+00 BC'::timestamp with time zone, 14454775807, 0, '04:00:54.775807', '+00:00:00.000000', '-4.923000', '-8.245000')
[info]            was aborted: ERROR: date/time field value out of range: "1301-02-29 22:13:15.805159+00 BC"  Call getNextException to see other errors in the batch.
[info]     Occurred when passed generated values (
[info]       arg0 = KS(AsteroidNew(ocyh5sijty),GN),
[...]

PR #246, opened 2018-03-01
Issue #241: don't generate ancient timestamps

swalker2m:
A workaround for what appears to be a `java.sql.Timestamp` formatting issue for ancient timestamps.  See #241.


tpolecat:
I wonder if we should go ahead and change `Timestamp.Min` to year zero since the whole point was to have a class with a total mapping to and from Postgres.

swalker2m:
`gem.util.Timestamp` does map to Postgres, but something about `java.sql.Timestamp` seems to generate an issue when actually writing ancient values to the database with JDBC.  So maybe you're right.

tpolecat:
What you have here is fine for now.

cquiroz:
I haven't read this at all but perhaps this tool is useful?

https://47deg.github.io/scalacheck-toolbox/docs/datetime/index.html

PR #250, opened 2018-03-06
ObservingNight

swalker2m:
This brings `ObservingNight` from the OCS codebase, updated to Scala and the new Java time API.  `ObservingNight` will appear in a number of places but there was one immediate need -- computing the time range for exporting ephemeris data.

#### Night

The `Night` concept applies to various definitions of start and end time, so it is pulled out in a separate trait like in the OCS codebase.  We will want to add `TwilightBoundedNight` at some point because knowing twilight bounds is important in many places.

#### Formatting / Parsing

An `ObservingNight` can be formatted into yyyyMMdd and, given a `Site`, parsed back into an `ObservingNight`.  I didn't try to use any optics here because you need a `Site`. I suppose I could have tried to create a `Format` given a particular `Site`, but I wasn't sure that really made sense.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
+++ b/modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
@@ -1,0 +1,133 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.math
+
+import gem.enum.Site
+
+import cats._
+import cats.data.Validated
+import cats.effect.IO
+import cats.implicits._
+
+import java.time._
+import java.time.format.DateTimeFormatter
+
+import scala.math.Ordering.Implicits._
+
+
+/** An observing night is defined as the period of time from 14:00 on one day
+  * until 14:00 on the following day.  In a timezone that honors daylight saving,
+  * it is sometimes onger and sometimes shorter than a period of 24 hours.  This
+  * is also true of days which contain leap seconds.
+  */
+final case class ObservingNight private (start: Instant, end: Instant, site: Site) extends Night {
+
+  // Sanity check ... should be correct via the companion constructor
+  assert(start < end)
+
+  def previous: ObservingNight =
+    ObservingNight.forInstant(start.minusNanos(1L), site)
+
+  def next: ObservingNight =
+    ObservingNight.forInstant(end, site)
+
+  def format: String =
+    ObservingNight.Formatter.withZone(site.timezone).format(end)
+
+}
+
+object ObservingNight {
+
+  /** The hour, in the local time zone, at which the night is considered to
+    * officially start.
+    *
+    * @group Constants
+    */
+  val LocalNightStartHour: Int =
+    14
+
+  /** The local time at which the night is considered to officially start.
+    *
+    * @group Constants
+    */
+  val LocalNightStart: LocalTime =
+    LocalTime.of(LocalNightStartHour, 0)
+
+  /** Formatter for nights.  The night string representation corresponds to the
+    * date YYYYMMDD for which the night ends in UTC.
+    *
+    * @group Constants
+    */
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import scala.math.Ordering.Implicits._
+
+
+/** An observing night is defined as the period of time from 14:00 on one day
+  * until 14:00 on the following day.  In a timezone that honors daylight saving,
+  * it is sometimes onger and sometimes shorter than a period of 24 hours.  This
+  * is also true of days which contain leap seconds.
+  */
+final case class ObservingNight private (start: Instant, end: Instant, site: Site) extends Night {
```
tpolecat: Did you intend `sealed abstract` here? The `private` constructor doesn't hide the factory methods and `copy`.

modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  /** Constructs the observing night that includes the given time at the
+    * specified site.
+    *
+    * @group Constructors
+    */
+  def forInstant(i: Instant, s: Site): ObservingNight = {
+    val zdt   = ZonedDateTime.ofInstant(i, s.timezone)
+
+    val twoPM = zdt.withHour(LocalNightStartHour)
```
tpolecat: Although 2:00p is probably set in stone the identifier `twoPM` makes me a little uneasy. Could we call this `zonedStart` or something?

modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
swalker2m: D'oh! I knew that, and yet screwed it up.  Thanks.

modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
```diff
@@ -0,0 +1,132 @@
...
+                   .withMinute(0)
+                   .withSecond(0)
+                   .withNano(0)
+
+    val (start, end) =
+      if (zdt.toLocalTime >= LocalNightStart) (twoPM, twoPM.plusDays(1L))
```
cquiroz: How do you get `LocalTime` to take `>=`? are there some implicit `Ordering`?

modules/core/shared/src/main/scala/gem/math/ObservingNight.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  /** Constructs the observing night for the given year, month, and day at the
+    * given site.
+    *
+    * @group Constructors
+    */
+  def forYMD(year: Year, month: Month, day: Int, site: Site): Option[ObservingNight] =
+    Validated.catchNonFatal {
+      ZonedDateTime.of(year.getValue, month.getValue, day, 0, 0, 0, 0, site.timezone)
+                   .minusNanos(1L)
```
cquiroz: This makes me slightly uneasy, why the -1 nano?

swalker2m: Actually, I should have made that `LocalNightStartHour` instead of 0.  The `ObservingNight` corresponding to that year, month, and day ends at the computed `ZoneDateTime`.  Since the end time is not included in the `ObservingNight` we need to get an instant that does fall in range for that night to pass to the `forInstant` constructor.  One nano is the minimum time before so I used that, but it could just as easily have been anything up to `.minusDays(1L)`.

tpolecat:
👍

tpolecat:
Note that `Semester` [adds 10h to push the time into the next day if it falls after 2pm](https://github.com/gemini-hlsw/gem/blob/master/modules/core/shared/src/main/scala/gem/Semester.scala#L106), so we should probably make it delegate to `ObservingNight` for that information so the 2pm thing is only defined once.

PR #254, opened 2018-03-08
Add GNIRS SmartGCAL

jdnavarro:
The tests and the importer are still missing, see  
functional-streams-for-scala/fs2#1099, but I think this may be good enough for merging.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
@@ -24,11 +24,11 @@ sealed abstract class DynamicConfig extends Product with Serializable {
     */
   def smartGcalKey(s: StaticConfig): Option[DynamicConfig.SmartGcalSearchKey] =
     (this, s) match {
-      case (d: DynamicConfig.F2,        _) => Some(d.key)
-      case (d: DynamicConfig.GmosNorth, _) => Some(d.key)
-      case (d: DynamicConfig.GmosSouth, _) => Some(d.key)
-//    case (d: DynamicConfig.Gnirs, s: StaticConfig.Gnirs) => Some(d.key(s))
-      case _                               => None
+      case (d: DynamicConfig.F2,        _)                 => Some(d.key)
+      case (d: DynamicConfig.GmosNorth, _)                 => Some(d.key)
+      case (d: DynamicConfig.GmosSouth, _)                 => Some(d.key)
+      case (d: DynamicConfig.Gnirs, s: StaticConfig.Gnirs) => Some(d.key(s))
+      case _                                               => None
     }
 }
 
@@ -83,6 +83,25 @@ object DynamicConfig {
 
     type GmosNorthDefinition = GmosDefinition[GmosNorthDisperser, GmosNorthFilter, GmosNorthFpu]
     type GmosSouthDefinition = GmosDefinition[GmosSouthDisperser, GmosSouthFilter, GmosSouthFpu]
+
+    final case class Gnirs(
+      acquisitionMirror: GnirsAcquisitionMirror,
+      pixelScale:        GnirsPixelScale,
+      disperser:         GnirsDisperser,
+      fpu:               Either[GnirsFpuOther, GnirsFpuSlit],
+      prism:             GnirsPrism,
+      wellDepth:         GnirsWellDepth
+    )
+
+    final case class GnirsSearch(
+      gnirs: Gnirs,
+      wavelength: Wavelength
+    ) extends SmartGcalSearchKey
+
+    final case class GnirsDefinition(
+      gnirs: Gnirs,
+      wavelengthRange: (Wavelength, Wavelength)
+    ) extends SmartGcalDefinitionKey
 
   }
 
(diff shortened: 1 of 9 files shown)


modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -83,6 +84,17 @@ object DynamicConfig {
...
 
+    final case class Gnirs(
+      camera:    GnirsCamera,
+      disperser: GnirsDisperser,
+      filter:    GnirsFilter,
+      fpu:       Either[GnirsFpuOther, GnirsFpuSlit],
+      prism:     GnirsPrism,
+      // TODO: static smart
+      wellDepth: GnirsWellDepth
+    ) extends SmartGcalSearchKey with SmartGcalDefinitionKey
```
swalker2m: I'm a bit confused because this doesn't seem to map very well to what I see in the [old codebase](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/smartgcal/keys/ConfigKeyGnirs.java#L36-L41).  There is no camera but there is an imaging vs. spectroscopy mode, filter shouldn't be there but cross dispersed mode should be, etc.  Am I looking at it wrong?

modules/db/src/main/scala/gem/dao/SmartGcalDao.scala
```diff
@@ -317,5 +323,69 @@ object SmartGcalDao {
...
       Update[GmosSouthRow](bulkInsertGmosSql("smart_gmos_south"))
+
+    // -----------------------------------------------------------------------
+    // GNIRS
+    // -----------------------------------------------------------------------
+
+    val createIndexGnirs: Update0 =
+      sql"""
+        CREATE INDEX IF NOT EXISTS smart_gnirs_index ON smart_gnirs
+          (camera, disperser, filter, fpu_slit, fpu_other, prism, well_depth)
```
swalker2m: From the [postgres doc](https://www.postgresql.org/docs/9.5/static/indexes-multicolumn.html), 

> Multicolumn indexes should be used sparingly. In most situations, an index on a single column is sufficient and saves space and time. Indexes with more than three columns are unlikely to be helpful unless the usage of the table is extremely stylized.

For GMOS, which is huge, having 3-column indices did help query performance.  GNIRS is comparatively tiny and I think we'd be fine with picking just a couple that fairly well evenly divide the table.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
jdnavarro: Those correspond to some of the GNIRS changes that were relatively easy to implement in the enumerations. These were discussed with @andrewwstephens, ideally there should be more changes, but the remaining changes were trickier to implement.

For example, Pixel Scale is now a column on the camera table, since the camera the choice of camera dictates the pixel scale.

I believe imaging vs. spectroscopy is distinguished now by the FPU.

swalker2m: The mapping from the old keys to the new one isn't clear to me.  Before it was imaging vs. spectroscopy.  How will you go from "SPECTROSCOPY" in the configuration file to a specific FPU?  Will the smart gcal config now differ depending upon which of the spectroscopy FPU options is selected?  How will you set the filter if it wasn't present before?  Maybe the best thing is to press on with the importer since these types of decisions will have to be made and it will become clearer.

jdnavarro: >  Maybe the best thing is to press on with the importer since these types of decisions will have to be made and it will become clearer.

Agreed, I'll include the importer in this PR.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -188,17 +207,33 @@ object DynamicConfig {
...
+      SmartGcalKey.GnirsSearch(
+        SmartGcalKey.Gnirs(
+          acquisitionMirror,
+          camera.pixelScale,
+          disperser,
+          fpu,
+          prism,
+          s.wellDepth
+        ),
+        Some(wavelength)
```
swalker2m: Perhaps the wavelength parameter doesn't need to be an `Option` if it is always `Some`?

modules/sql/src/main/resources/db/migration/V054__Add_Gnirs_SmartGcal.sql
```diff
@@ -0,0 +1,32 @@
...
+    fpu_slit           identifier                   REFERENCES e_gnirs_fpu_slit           ON DELETE CASCADE,
+    fpu_other          identifier                   REFERENCES e_gnirs_fpu_other          ON DELETE CASCADE,
+    prism              identifier         NOT NULL  REFERENCES e_gnirs_prism              ON DELETE CASCADE,
+    well_depth         identifier         NOT NULL  REFERENCES e_gnirs_well_depth         ON DELETE CASCADE,
+    min_wavelength     int                NOT NULL  CHECK (min_wavelength >= 0),
+    max_wavelength     int                NOT NULL  CHECK (max_wavelength >= 0),
+    gcal_id            integer            NOT NULL  REFERENCES gcal                       ON DELETE CASCADE
+    CONSTRAINT fpu_slit_or_other
+      CHECK ((fpu_slit IS     NULL AND fpu_other IS NOT NULL) OR
+             (fpu_slit IS NOT NULL AND fpu_other IS     NULL))
```
swalker2m: Also probably good to add `CHECK (min_wavelength <= max_wavelength)`.

PR #255, opened 2018-03-08
Workaround for FS2 issue 1099

swalker2m:
Temporary workaround for FS2 issue 1099:

https://github.com/functional-streams-for-scala/fs2/issues/1099


swalker2m:
By the way @jdnavarro, I suggest editing the various GMOS `.csv` files down to just a few lines of configuration when testing your importer.  GMOS takes a long time to import and it will slow you down otherwise.

PR #256, opened 2018-03-12
Add static config to smartgcal key creation

swalker2m:
@jdnavarro discovered that GNIRS smartgcal keys require a bit of static configuration (GNIRS well depth).  The existing smartgcal code assumes that the smartgcal key can be extracted from dynamic config alone.  This PR adds static configuration to key creation so that it can work for GNIRS as well.

You might want to look at the two commits individually or maybe focus on just the second commit since it contains the payload.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
@@ -22,12 +22,13 @@ sealed abstract class DynamicConfig extends Product with Serializable {
     * configuration.
     * @return corresponding smart gcal search key, if any
     */
-  def smartGcalKey: Option[DynamicConfig.SmartGcalSearchKey] =
-    this match {
-      case f2: DynamicConfig.F2        => Some(f2.key)
-      case gn: DynamicConfig.GmosNorth => Some(gn.key)
-      case gs: DynamicConfig.GmosSouth => Some(gs.key)
-      case _                           => None
+  def smartGcalKey(s: StaticConfig): Option[DynamicConfig.SmartGcalSearchKey] =
+    (this, s) match {
+      case (d: DynamicConfig.F2,        _) => Some(d.key)
+      case (d: DynamicConfig.GmosNorth, _) => Some(d.key)
+      case (d: DynamicConfig.GmosSouth, _) => Some(d.key)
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -22,12 +22,13 @@ sealed abstract class DynamicConfig extends Product with Serializable {
...
-      case gn: DynamicConfig.GmosNorth => Some(gn.key)
-      case gs: DynamicConfig.GmosSouth => Some(gs.key)
-      case _                           => None
+  def smartGcalKey(s: StaticConfig): Option[DynamicConfig.SmartGcalSearchKey] =
+    (this, s) match {
+      case (d: DynamicConfig.F2,        _) => Some(d.key)
+      case (d: DynamicConfig.GmosNorth, _) => Some(d.key)
+      case (d: DynamicConfig.GmosSouth, _) => Some(d.key)
+//    case (d: DynamicConfig.Gnirs, s: StaticConfig.Gnirs) => Some(d.key(s))
+      case _                               => None
```
tpolecat: Do we want to contrive a way to check this at compile time?

swalker2m: > Do we want to contrive a way to check this at compile time?

You mean check that the dynamic and static config are for the same instrument?  That would be ideal.

tpolecat: I mean know at compile time whether or not there is a key for a config of the given type, so we can get rid of the `Option` return type. Are there cases where we have to look at the data to know?

swalker2m: Some instruments don't have any smartgcal mappings, but I don't think we ever need to look at the configuration in the step to know.  The `Option` seems simple, but if there is a straightforward way to do this at compile time that would be even better.  How much would that complicate the model?

tpolecat: I don't know. Maybe we could add a `HasSmartGcal` typeclass or something. Not a big deal. In any case should wait until we get the instrument-indexed types figured out.

PR #258, opened 2018-03-14
Enumerations for GNIRS SmartGCAL key lookup

jdnavarro:
- Make Acquisition Mirror an enumeration with its own table instead of a
  postgres enum type.
- Make Pixel Scale an enumeration with its own table instead of a column of
  GNIRS camera.

See #254.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
@@ -188,6 +188,7 @@ object DynamicConfig {
   }
 
   final case class Gnirs(
+    acquisitionMirror:    GnirsAcquisitionMirror,
     camera:               GnirsCamera,
     decker:               GnirsDecker,
     disperser:            GnirsDisperser,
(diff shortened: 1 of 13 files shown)


modules/core/shared/src/main/scala/gem/enum/PixelScale.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import cats.instances.string._
+import cats.syntax.eq._
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for Pixel Scale.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class PixelScale(
```
swalker2m: You asked about this and I mentioned that other instruments have the same concept.  They won't match the GNIRS values though so this should probably be `GnirsPixelScale`.

jdnavarro: :) I actually had it first as `GnirsPixelScale` but changed it when I had to add it to `EnumDef`. There is no instrument specific code in `EnumDef`, is it fine to have `GnirsPixelScale` in `EnumDef`? Where should I place it otherwise?

swalker2m: Hmm, unfortunately I don't know of another alternative.

tpolecat: In `EnumDef` is fine with me. We can split it up at some point if it gets to be too much of a junk drawer.

PR #259, opened 2018-03-14
gem.math optics

tpolecat:
I needed a break from type experiments so I filled out `Lens`and `Iso` for stuff in `gem.math`. I'm using `GenLens` instead of `@Lenses` because I don't like having to suppress the bogus warning and I want a doc comment with `@group Optics` so the lenses are grouped in the Scaladoc.

This adds optics and tests but doesn't change any existing code.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/EphemerisCoordinates.scala
+++ b/modules/core/shared/src/main/scala/gem/math/EphemerisCoordinates.scala
@@ -4,12 +4,14 @@
 package gem.math
 
 import cats._
+import monocle.Lens
+import monocle.macros.GenLens
 
 /** A coordinate along with a rate of change in RA and Dec for some time unit,
   * expressed as an offset in p and q.  In reality the velocity information
   * comes from horizons and is always arcseconds per hour in horizons data.
   *
-  * @param coord coordinates
+  * @param coord delta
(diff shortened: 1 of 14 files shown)


modules/core/shared/src/main/scala/gem/math/EphemerisCoordinates.scala
```diff
@@ -4,12 +4,14 @@
...
 import cats._
+import monocle.Lens
+import monocle.macros.GenLens
 
 /** A coordinate along with a rate of change in RA and Dec for some time unit,
   * expressed as an offset in p and q.  In reality the velocity information
   * comes from horizons and is always arcseconds per hour in horizons data.
   *
-  * @param coord coordinates
+  * @param coord delta
```
sraaphorst: Just curious why you changed this to be `delta` twice?

sraaphorst:
I haven't had the pleasure of using Monocle yet, but I am learning a lot through these PRs.

modules/core/shared/src/main/scala/gem/math/EphemerisCoordinates.scala
tpolecat: Doh! I'll fix that. Thanks.

tpolecat:
Works for me!

PR #260, opened 2018-03-14
move index to its own compilation unit

tpolecat:
`Observation.Id` uses `Observation.Index` to guarantee a positive observation number, but this is useful in other places as well (like `Program.Id` and `Dataset.Label`) so I'm pulling it out into its own compilation unit. This makes no changes other than disentanglement from `Observation`.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Observation.scala
+++ b/modules/core/shared/src/main/scala/gem/Observation.scala
@@ -7,8 +7,7 @@ import cats.{ Functor, Order, Show }
 import cats.implicits._
 import gem.config.{ StaticConfig, DynamicConfig }
 import gem.enum.Instrument
-import gem.syntax.string._
-import mouse.boolean._
+import gem.math.Index
 
 /**
  * An observation, parameterized over the types of its targets, static config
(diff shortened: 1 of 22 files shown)


tpolecat:
> don't we lose a bit of type safety

Yes, potentially, but I'm less concerned (for now) about people getting them confused and more concerned about the model allowing non-positive indices.

There are really two different concerns around newtyping. We want to have a refined type (non-negative `Short`) and we also want a semantic tag "serial id for an observation within a given program". In our code these ideas are often conflated but we may want to think about a more principled strategy.

This is a case where a type like `(Short Refined Positive) @@ ObservationIndex` might end up making sense, using [fthomas/refined](https://github.com/fthomas/refined) for refinement along with shapeless tags for semantic marking. This would let us remove some of our boilerplatey wrappers.

tpolecat:
With refined the `Index` type would look like this, which isn't a huge savings I guess.

```scala
type Index = Short Refined Positive
object Index {

  val fromShort: Prism[Short, Index] =
    refinedPrism[Short, Positive]

  val fromString: Format[String, Index] =
    Format[String, Short](short.parseExact(_), _.toString) composePrism fromShort

  implicit val OrderIndex: Order[Index] =
    Order.by(_.value)

  implicit val OrderingIndex: scala.math.Ordering[Index] =
    OrderIndex.toOrdering

  implicit val showIndex: Show[Index] =
    Show.fromToString

}
```

tpolecat:
A third concern in the newtype game is units/dimensions.

PR #261, opened 2018-03-15
Fix HourAngle.fromDoubleHours

swalker2m:
This PR fixes a bug in `HourAngle.fromDoubleHours`.  It also adds a `toDoubleHours` method.  I could easily be convinced to place that in the test method itself though.  I don't have a good enough feel for optics to know whether there should be a `doubleHours` optic but I didn't try to add one since I didn't see any others for approximate conversions like that one.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/math/Angle.scala
+++ b/modules/core/shared/src/main/scala/gem/math/Angle.scala
@@ -369,6 +369,13 @@ final class HourAngle private (µas: Long) extends Angle(µas) {
     toMicroarcseconds / 15
 
   /**
+   * This `HourAngle` in decimal hours. Approximate.
+   * @group Conversions
+   */
+  def toDoubleHours: Double =
+    toMicroseconds.toDouble / HourAngle.µsPerHour
+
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -410,7 +419,7 @@ object HourAngle extends HourAngleOptics {
    * @group Constructors
    */
   def fromDoubleHours(hs: Double): HourAngle =
-    fromMicroseconds((hs * 60.0 * 60.0 * 1000.0).toLong)
+    fromMicroseconds((hs * µsPerHour).round)
```
swalker2m: The bug was just that we were off by 1000, but I also changed this to `round`.  Is there any reason why we should just truncate instead?

tpolecat: No, I didn't know about `.round`. That's fine.

We can't define round-trip optics for doubles because in general they never work. That's another area we could explore I guess.

PR #264, opened 2018-03-16
Sidereal time calculator

swalker2m:
This is work-in-progress on porting "skycalc" into Scala and the new codebase.  It adds a basic sidereal time calculation that improves upon the API in the [old skycalc `lst` method](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/ImprovedSkyCalcMethods.java#L1053-L1089).

While researching sidereal time calculation, I learned we are using a really old algorithm that has been subsequently updated (see [The International Celestial Reference System](http://www.iausofa.org/publications/sofa_pn.pdf)).

> Greenwich sidereal time (GST), which is a complicated function of both Earth rotation (UT1) and time (not quite the same thing), was replaced by Earth rotation angle (ERA), ...

LST is used in a number of calculations in the old Skycalc code (such as the position of sun, which determines twilight boundaries).  I'd like to revisit this issue and consider instead porting the relevant bits of the International Astronomical Union's "Standards of Fundamental Astronomy" library.


swalker2m:
You'll see I added a `gem-math-old.jar` library to the `coreJVM` `lib` directory.  This is just `ImprovedSkyCalc` and friends.  I wanted to use it for comparison and originally placed the source in `coreJVM` test.  Unfortunately that would inexplicably cause the compiler to throw an exception when used so I extracted it out into an unmanaged jar.

Ideally this would only be included on the test classpath but I think at the moment it is just available everywhere in `coreJVM`.

tpolecat:
I think you can put the jarfile in `modules/core/lib_test/` or something and then say `unmanagedJars in (Test, compile) += baseDirectory.value / "lib_test" / "gem-math-old.jar"`. I think. It almost certainly won't work and you'll have to track down Eugene in Japan to get a real answer but it's worth a try I guess.

swalker2m:
I couldn't figure out how to do the old SkyCalc jar correctly and hope to take a stab at doing this in a different way anyway.

PR #271, opened 2018-03-21
add subgroup test for HourAngle

tpolecat:
It occurred to me that we hadn't actually proved `HourAngle` is a subgroup of `Angle` so I added a test for that. Slightly roundabout but the laws might come in handy again if we identify other structure-preserving transformations in the future.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/laws/HomomorphismLaws.scala
+++ b/modules/core/shared/src/test/scala/gem/laws/HomomorphismLaws.scala
@@ -1,0 +1,33 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.laws
+
+import cats.{ Semigroup, Monoid, Group }
+
+abstract class SemigroupHomomorphismLaws[A, B](f: A => B) {
+  val A: Semigroup[A]
+  val B: Semigroup[B]
+
+  def combine(a: A, b: A): IsEq[B] =
+    f(A.combine(a, b)) <-> B.combine(f(a), f(b))
+
+}
+
+abstract class MonoidHomomorphismLaws[A, B](f: A => B) extends SemigroupHomomorphismLaws[A, B](f) {
+  override val A: Monoid[A]
+  override val B: Monoid[B]
+
+  def empty: IsEq[B] =
+    f(A.empty) <-> B.empty
+
+}
+
+abstract class GroupHomomorphismLaws[A, B](f: A => B) extends MonoidHomomorphismLaws[A, B](f) {
+  override val A: Group[A]
+  override val B: Group[B]
+
+  def inverse(a: A): IsEq[B] =
+    f(A.inverse(a)) <-> B.inverse(f(a))
+
+}
--- a/modules/core/shared/src/test/scala/gem/laws/SubgroupLaws.scala
+++ b/modules/core/shared/src/test/scala/gem/laws/SubgroupLaws.scala
@@ -1,0 +1,7 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.laws
+
+// A is a subgroup of B if covariant widening is a group homomorphism.
+abstract class SubgroupLaws[A, B](f: A <:< B) extends GroupHomomorphismLaws[A, B](f)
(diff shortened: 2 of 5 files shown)


modules/core/shared/src/test/scala/gem/laws/HomomorphismLaws.scala
```diff
@@ -0,0 +1,33 @@
...
+
+}
+
+abstract class GroupHomomorphismLaws[A, B](f: A => B) extends MonoidHomomorphismLaws[A, B](f) {
+  override val A: Group[A]
+  override val B: Group[B]
+
+  def inverse(a: A): IsEq[B] =
+    f(A.inverse(a)) <-> B.inverse(f(a))
+
```
sraaphorst: This all seems perfect to me. Good stuff.

modules/core/shared/src/test/scala/gem/laws/SubgroupLaws.scala
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.laws
+
+// A is a subgroup of B if covariant widening is a group homomorphism.
+abstract class SubgroupLaws[A, B](f: A <:< B) extends GroupHomomorphismLaws[A, B](f)
```
sraaphorst: I assume that you mean by "covariant widening" that you're just talking about the natural map that embeds `A` into `B`. I'm not sure exactly what `<:<` means.

sraaphorst: I'm assuming in this case that you are considering that `A` is a subgroup (not necessarily proper) of `B`?

You're not considering, say, the group `A` of all the multiples of, say, the 2x2 identity matrix under addition  a direct subgroup of the rationals even though you can map `A` isomorphically to the integers, which clearly have an embedding homomorphism into the rationals?

modules/core/shared/src/test/scala/gem/laws/discipline/SubgroupTests.scala
```diff
@@ -0,0 +1,42 @@
...
+             ib: Group[B]
+  ): SubgroupTests[A, B] =
+    new SubgroupTests[A, B] {
+      val laws = new SubgroupLaws[A, B](ev) {
+        val A = ia
+        val B = ib
+      }
+    }
+
+}
```
sraaphorst: 👍

modules/core/shared/src/test/scala/gem/laws/SubgroupLaws.scala
tpolecat: Yes. In Scala terms

- if `A` is a subtype of `B` (a value of type `A <:< B` is a proof of this fact), and
- we have `Group[A]` and `Group[B]`, and
- we claim that `A` is a subgroup of `B`, then

the widening `A => B` (i.e., just forgetting it's an `A`) must be a group homomorphism.

tpolecat:
There was a problem with the `Index` spec, which I fixed here. Will need re-approval when it's green.

modules/core/shared/src/test/scala/gem/math/HourAngleSpec.scala
```diff
@@ -17,6 +17,7 @@ final class HourAngleSpec extends CatsSuite {
   // Laws
   checkAll("HourAngle", CommutativeGroupTests[HourAngle].commutativeGroup)
   checkAll("HourAngle", EqTests[HourAngle].eqv)
+  checkAll("HourAngle <: Angle", SubgroupTests[HourAngle, Angle].subgroup)
```
sraaphorst: I needed to read your description about what exactly an `HourAngle` was to see why it wasn't isomorphic to `Angle`, but now I get it. Yes, it should definitely embed via the identity homomorphism as a subgroup of `Angle`.

modules/core/shared/src/test/scala/gem/laws/SubgroupLaws.scala
tpolecat: > You're not considering ...

Yes, if I understand what you're saying. Here every `A` *is* a `B`; it's a subset. So this isn't an arbitrary mapping.

sraaphorst: That sounds exactly right to me. Thanks for the explanation regarding `<:<`.

modules/core/shared/src/test/scala/gem/math/HourAngleSpec.scala
tpolecat: 👍

PR #272, opened 2018-03-22
Add coadds to GNIRS

jdnavarro:
The type CoAdds was already existing for GCAL but it was factored out so it
could be used in GNIRS.

In the beginning this sounded like a trivial change but in the end I had to tweak the code in many places.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/CoAdds.scala
+++ b/modules/core/shared/src/main/scala/gem/CoAdds.scala
@@ -1,0 +1,35 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+import cats.{ Order, Show }
+import cats.instances.short._
+import gem.syntax.prism._
+import monocle.Prism
+
+sealed abstract case class CoAdds private (toShort: Short) {
+  // Sanity check … should be correct via the companion constructor.
+  assert(toShort > 0, s"Invariant violated. $toShort is 1 or greater.")
+}
+
+object CoAdds {
+
+  final val One: CoAdds = fromShort.unsafeGet(1)
+
+  /** @group Typeclass Instances */
+  implicit val CoAddsShow: Show[CoAdds] =
+    Show.fromToString
+
+  /** @group Typeclass Instances */
+  implicit val CoAddsOrd: Order[CoAdds] =
+    Order.by(_.toShort)
+
+  /**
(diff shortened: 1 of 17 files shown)


modules/core/shared/src/main/scala/gem/CoAdds.scala
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+// CoAdds has a DISTINCT type due to its check constraint so we need a fine-grained mapping
+// here to satisfy the query checker.
+final case class CoAdds(toShort: Short)
```
swalker2m: A couple of things here:

* The comment there applies to the DAO package version, not the model concept per se.  It requires this mapping because there is a database constraint that its value be `> 0`.

* We should guarantee that the model class is, correspondingly, always positive non-zero.  So I think the way we're doing this at least for now is with `sealed abstract case class` instances.  See `gem.math.Wavelength` which has the same deal.

* In the `CoAdds` companion we should define a `One` instance since that is the default, and an `Order` instance I expect along with a `fromShort: Prism[Short, CoAdds]` similar to the one for `Wavelength`.

modules/core/shared/src/main/scala/gem/CoAdds.scala
```diff
@@ -0,0 +1,35 @@
...
+  /** @group Typeclass Instances */
+  implicit val CoAddsOrd: Order[CoAdds] =
+    Order.by(_.toShort)
+
+  /**
+    * Prism from Short into CoAdds and back.
+    * @group Optics
+    */
+  def fromShort: Prism[Short, CoAdds] =
+    Prism((n: Short) => Some(n).filter(_ >= 0).map(new CoAdds(_) {}))(_.toShort)
```
swalker2m: `> 0` here too

modules/core/shared/src/main/scala/gem/CoAdds.scala
```diff
@@ -0,0 +1,35 @@
...
+package gem
+
+import cats.{ Order, Show }
+import cats.instances.short._
+import gem.syntax.prism._
+import monocle.Prism
+
+sealed abstract case class CoAdds private (toShort: Short) {
+  // Sanity check … should be correct via the companion constructor.
+  assert(toShort > 0, s"Invariant violated. $toShort is negative.")
```
swalker2m: `Invariant violated. $toShort must be > 0.`

PR #273, opened 2018-03-22
minor version updates

tpolecat:
- cats 1.1.0
- cats-effect 0.10
- fs2 0.10.3
- http4s 0.18.3
- monocle 1.5.1-cats


build.sbt
```diff
@@ -2,16 +2,16 @@ import com.typesafe.sbt.packager.docker._
...
 lazy val doobieVersion        = "0.5.1"
 lazy val flywayVersion        = "5.0.7"
-lazy val fs2Version           = "0.10.2"
-lazy val http4sVersion        = "0.18.2"
+lazy val fs2Version           = "0.10.3"
+lazy val http4sVersion        = "0.18.3"
 lazy val jwtVersion           = "0.16.0"
 lazy val kpVersion            = "0.9.6"
-lazy val monocleVersion       = "1.5.0-cats"
+lazy val monocleVersion       = "1.5.1-cats"
```
cquiroz: A new monocle version? I have to update my patch for `scalajs-react`

swalker2m:
Need to find a way to automate testing the stuff that uses FS2 (smart gcal import, ephemeris import).  The last couple of version updates introduced class cast exceptions.

tpolecat:
Indeed, we need tests for that code. Maybe we should add a code coverage check so we know where our blind spots are.

tpolecat:
I'm fine checking in smart gcal and ephemeris samples into the test resource directories. We don't need all the data but we do need at least `$BATCHSIZE` rows because chunking handling ended up being significant in the last few fs2 issues.

swalker2m:
> I'm fine checking in smart gcal and ephemeris samples into the test resource directories.

I think that's the way to go.  I'll add a task.

PR #276, opened 2018-03-22
ProgramId.fromString

tpolecat:
This fuses the parse/format pairs for program ids.

- Strings containing whitespace are not valid program ids. All other strings are. This means we have at best a `Prism`.
- Leading zeroes on science program indices are allowed (and discarded) which weakens the relationship to `Format`. If we drop this behavior we go back up to `Prism`.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Observation.scala
+++ b/modules/core/shared/src/main/scala/gem/Observation.scala
@@ -55,7 +55,7 @@ object Observation {
   /** An observation is identified by its program and a serial index. */
   final case class Id(pid: Program.Id, index: Index) {
     def format: String =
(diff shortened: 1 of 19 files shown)


tpolecat:
> `Format` because multiple strings will map to the same ID (because of the leading zeros), correct?

Yes, correct. But we could close that loophole. It seems like more of a parser oversight than an intentional design.

swalker2m:
> But we could close that loophole. It seems like more of a parser oversight than an intentional design.

Agree, we should probably just disallow leading zeros and upgrade to `Prism`.

tpolecat:
Ok I'll make that change.

PR #277, opened 2018-04-19
Switch back to LBS

tpolecat:
I think we have hit enough tool limitations that it makes sense to switch back to Lightbend Scala. The changes are relatively minor: `shapeless.Witness` rather than `ValueOf` and then some hacking to get around "unused" warnings.

(I regenerated all the enums and they were identical.)

### DIFF
--- a/modules/db/src/main/scala/gem/dao/composite/ProperMotion.scala
+++ b/modules/db/src/main/scala/gem/dao/composite/ProperMotion.scala
@@ -28,7 +28,8 @@ private object ProperMotionCompositeLemmas {
   import RadialVelocityMeta._
 
   // The parallax member is an angle, which we will store as signed milliarcseconds
-  private implicit lazy val AngleMasMeta: Meta[Angle] =
+  // N.B. if this is marked private we get a spurious "unused" warning
+  implicit lazy val AngleMasMeta: Meta[Angle] =
     AngleMeta.AngleMetaAsSignedMilliarcseconds
 
   val ProperMotionComposite: Composite[ProperMotion] = implicitly
--- a/modules/sql/src/main/scala/EnumDef.scala
+++ b/modules/sql/src/main/scala/EnumDef.scala
@@ -80,9 +80,9 @@ object EnumDef {
     implicit def caseOptionMagnitudeValue[S <: Symbol] = at[(S, Option[MagnitudeValue])] { case (s, _) => s"  val ${s.name}: Option[MagnitudeValue]" }
     implicit def caseGnirsPixelScale[S <: Symbol] = at[(S, GnirsPixelScale)] { case (s, _) => s"  val ${s.name}: GnirsPixelScale" }
 
-    implicit def caseEnumRef[T <: Symbol: ValueOf, S <: Symbol]       = at[(S, EnumRef[T])        ] { case (s, _) => s"  val ${s.name}: ${valueOf[T].name}" }
-    implicit def caseLazyEnumRef[T <: Symbol: ValueOf, S <: Symbol]   = at[(S, LazyEnumRef[T])    ] { case (s, _) => s"  val ${s.name}: Eval[${valueOf[T].name}]" }
-    implicit def caseOptionEnumRef[T <: Symbol: ValueOf, S <: Symbol] = at[(S, Option[EnumRef[T]])] { case (s, _) => s"  val ${s.name}: Option[${valueOf[T].name}]" }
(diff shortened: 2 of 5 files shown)


cquiroz:
I think you can remove https://github.com/gemini-hlsw/gem/pull/277/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceL188

That was required to make scala.js play nice with TLS

tpolecat:
> Would you go back to literal types once they are in LBS?

Yes, they'll be in 2.13

modules/db/src/main/scala/gem/dao/composite/ProperMotion.scala
```diff
@@ -28,7 +28,8 @@ private object ProperMotionCompositeLemmas {
   import RadialVelocityMeta._
 
   // The parallax member is an angle, which we will store as signed milliarcseconds
-  private implicit lazy val AngleMasMeta: Meta[Angle] =
+  // N.B. if this is marked private we get a spurious "unused" warning
+  implicit lazy val AngleMasMeta: Meta[Angle] =
```
swalker2m: I've seen this naming style a couple of times lately for typeclass instances.  I kind of like it better than just `AngleMeta` or `MetaAngle`.  Maybe we should do it everywhere else?

modules/sql/src/main/scala/EnumDef.scala
```diff
@@ -80,9 +80,9 @@ object EnumDef {
     implicit def caseOptionMagnitudeValue[S <: Symbol] = at[(S, Option[MagnitudeValue])] { case (s, _) => s"  val ${s.name}: Option[MagnitudeValue]" }
     implicit def caseGnirsPixelScale[S <: Symbol] = at[(S, GnirsPixelScale)] { case (s, _) => s"  val ${s.name}: GnirsPixelScale" }
 
-    implicit def caseEnumRef[T <: Symbol: ValueOf, S <: Symbol]       = at[(S, EnumRef[T])        ] { case (s, _) => s"  val ${s.name}: ${valueOf[T].name}" }
-    implicit def caseLazyEnumRef[T <: Symbol: ValueOf, S <: Symbol]   = at[(S, LazyEnumRef[T])    ] { case (s, _) => s"  val ${s.name}: Eval[${valueOf[T].name}]" }
-    implicit def caseOptionEnumRef[T <: Symbol: ValueOf, S <: Symbol] = at[(S, Option[EnumRef[T]])] { case (s, _) => s"  val ${s.name}: Option[${valueOf[T].name}]" }
+    implicit def caseEnumRef[T <: Symbol, S <: Symbol](implicit w: Witness.Aux[T])       = at[(S, EnumRef[T])        ] { case (s, _) => s"  val ${s.name}: ${w.value.name}" }
```
swalker2m: How does `Witness` differ from `ValueOf`?

modules/db/src/main/scala/gem/dao/composite/ProperMotion.scala
tpolecat: I don't feel strongly about it, but we do run the risk of the names getting stale if we do this … I have only been doing it for non-canonical instances (which we only use locally).

swalker2m: I just realized I misread the name as `AngleHasMeta`.  I'd seen `Is` used in other places as in `WuzzleIsFunctor`.

modules/sql/src/main/scala/EnumDef.scala
tpolecat: `Witness` kind of serves double duty. It lets you express both singleton types and singleton values (i.e., values that have an ascribed singleton type).

So, with literal types (which we will get back in 2.13) you can say

```scala
val op: Option[42] = Some(42)
```

You can't do that in Lightbend 2.12 … `42` can't appear in type position, and the value `42` has type `Int`. But you can do it with `Witness`.

```scala
val op: Option[Witness.`42`.T] = Some(Witness(42).value) 
val op: Option[Witness.`42`.T] = Some(42.narrow) // syntax 
```

The `42` in backticks is a trick. It's a method call (!) on `object Witness` (which extends `Dynamic`). The `selectDynamic` method is implemented with a macro that parses and typechecks the dynamic method name (!) and constructs the singleton type, which you can't express in Scala but *can* express in terms of the underlying AST.

`Witness.Aux[A]` is always available for singleton types `T` so you can write `valueOf` in terms of it.

```scala
@ def valueOf[A](implicit ev: Witness.Aux[A]): A = ev.value 
defined function valueOf

@ valueOf[Witness.`42`.T] 
res11: Int = 42
```

